PR #103, opened 2024-05-03
enhancement #92: run vitest for ui-vue in browsers

### LINKED ISSUE #92: Update vitest config for ui-vue to run tests in browsers as well
Current all component test for ui-vue run in jsdom environment, even when `yarn workspace @odk-web-forms/ui-vue test-browser:firefox` is executed. We would like to run them in browser environment as well.

ref: https://github.com/getodk/web-forms/pull/89#discussion_r1579988172

sadiqkhoja:
Closes #92


eyelidlessness (approved):
Thanks! I did a quick local spot check with a dummy test to see the output of `globalThis.navigator.userAgent`. Each command produced a value consistent with the environment.

PR #109, opened 2024-05-09
fixes #108: resolve existing eslint warnings

### LINKED ISSUE #108: Resolve existing eslint warnings
There are 14 eslint warnings in `xpath` and `xforms-engine`, turn all those rules into errors and resolve them.

sadiqkhoja:
Closes #108

### DIFF
--- a/packages/xforms-engine/src/body/text/TextElementDefinition.ts
+++ b/packages/xforms-engine/src/body/text/TextElementDefinition.ts
@@ -78,8 +78,6 @@ export abstract class TextElementDefinition<
 				return [];
 			}
 
-			console.error('Unexpected text element child', node);
-
 			throw new Error(`Unexpected <${element.nodeName}> child element`);
 		});
 
--- a/packages/xforms-engine/src/body/text/TextElementDefinition.ts
+++ b/packages/xforms-engine/src/body/text/TextElementDefinition.ts
@@ -78,6 +78,9 @@ export abstract class TextElementDefinition<
 				return [];
 			}
 
+			// eslint-disable-next-line no-console
+			console.error('Unexpected text element child', node);
+
 			throw new Error(`Unexpected <${element.nodeName}> child element`);
 		});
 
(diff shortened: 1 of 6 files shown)


packages/xforms-engine/src/body/text/TextElementDefinition.ts
```diff
@@ -78,8 +78,6 @@ export abstract class TextElementDefinition<
 				return [];
 			}
 
-			console.error('Unexpected text element child', node);
-
```
eyelidlessness, line 82: This log is useful. I'd prefer to ignore it rather than remove.

packages/xpath/test/native/comparison-operator2.spec.ts
```diff
@@ -161,12 +161,12 @@ describe('Comparison operator (2)', () => {
 	});
 
 	describe('<, <=, >, >=', () => {
-		interface TestCaseInput {
+		interface InequalityTestCaseInput {
```
eyelidlessness, line 164: I'd prefer `RelationalExprTestCaseInput`, as that's the term for these expressions in the [spec](https://www.w3.org/TR/1999/REC-xpath-19991116/#NT-RelationalExpr). (For future reference, it has been really helpful throughout the `xpath` package to hew close to spec terminology wherever possible.)

packages/xpath/test/native/node-name.spec.ts
```diff
@@ -84,7 +84,7 @@ describe('node name for', () => {
 	it('attribute with a specific name', () => {
 		const contextNode = document.getElementById('StepNodeTestCaseNameTestAttribute')!;
 		const attribute = getNonNamespaceAttributes(contextNode).find(
-			(attribute) => attribute.localName === 'attrib3'
+			(att) => att.localName === 'attrib3'
```
eyelidlessness, line 87: If we're going from full word to shorthand, I'd prefer to use the more idiomatic `attr`.

eyelidlessness (changes requested):
Thanks! Just a few minor nits.

eyelidlessness (approved):
Thanks!

PR #111, opened 2024-05-09
fixes #85: convert all chai-assertions to jest

### LINKED ISSUE #85: Convert existing Chai-style assertions to Jest-style
For the web-forms mono repository, we have decided to go with jest-style assertions. Currently there are few tests in `common` and `xpath` packages that use chai-style assertions, those should be converted to jest-style assertions.

sadiqkhoja:
Closes #85 

### How have I done this?

Searched for `to.` in all files and replaced each type of assertions separately, like replace all `to.be.eql` with `toEqual`

_Note: I have replaced `to.deep.equal` with `toEqual` because jest's toEqual checks equality deeply - please double check if that's true._


packages/xpath/vite.config.ts
```diff
@@ -125,7 +125,7 @@ export default defineConfig(({ mode }) => {
 
 			environment: TEST_ENVIRONMENT,
 			globals: false,
-			include: ['src/**/*.test.ts', 'test/native/index.ts', 'test/xforms/index.ts'],
+			include: ['test/**/*.test.ts', 'test/native/index.ts', 'test/xforms/index.ts'],
```
sadiqkhoja, line 128: 😁 some tests were not even running

packages/xpath/test/helpers.ts
```diff
@@ -194,9 +194,10 @@ export class TestContext<XForms extends boolean = false> {
 		const { message } = options;
 
 		if (Number.isNaN(expected)) {
-			expect(numberValue, message).to.be.NaN;
+			expect(numberValue, message).toBeNaN();
 		} else {
-			expect(numberValue, message).to.equal(expected);
+			// numberValue can be -0; jest's toEqual fails when -0 is compared with 0
+			expect(expected === numberValue, message).toEqual(true);
```
sadiqkhoja, line 200: 💭 maybe we should change evaluator to never return -0

packages/xpath/vite.config.ts
eyelidlessness: Yikes! Good catch.

I think in the future, whenever I'm poking around here again, I'll eliminate the `index.ts` indirections and make this more uniform with other projects. Those were helpful for quick iteration when I was deeply focused on the `xpath` package, but I don't think they're worth the minor test run time saved now.

packages/xpath/test/helpers.ts
eyelidlessness: No. The [spec](https://www.w3.org/TR/1999/REC-xpath-19991116/#numbers) is clear that negative zero is a distinct value, per IEEE 754 (emphasis added):

> A number represents a floating-point number. A number can have any double-precision 64-bit format IEEE 754 value [[IEEE 754]](https://www.w3.org/TR/1999/REC-xpath-19991116/#IEEE754). These include a special "Not-a-Number" (NaN) value, positive and negative infinity, and positive **and negative zero**. See [Section 4.2.3](http://java.sun.com/docs/books/jls/html/4.doc.html#9208) of [[JLS]](https://www.w3.org/TR/1999/REC-xpath-19991116/#JLS) for a summary of the key rules of the IEEE 754 standard.

I see three failures caused by the stricter check. Of those:

- One failure in `round.spec.ts` exercises an ODK extension of the standard `round`, and the behavior is consistent with the underlying [single-arity spec](https://www.w3.org/TR/1999/REC-xpath-19991116/#function-round).

- Two failures in `number-operator.spec.ts` exercise the `mod` and `*` operators. The XPath spec itself doesn't go into detail of the behavior, but the same `-0` values are exactly what the native XPath evaluator produce.

In all of these cases, I believe the functionality is correct and expected, and that the tests were previously too lax.

I think the right thing to do is to correct the expected values for those three failures, and change this to the stricter/more correct:

```suggestion
			expect(numberValue, message).toBe(expected);
```

... and probably remove the comment above it.

eyelidlessness (changes requested):

eyelidlessness (approved):
Nice! I'll file a followup issue to do a pass verifying that we're producing `-0` correctly across a wider set of (existing, maybe additive) checks. I'm a bit surprised that seemingly equivalent assertions have different levels of strictness, but I'm really glad this surfaced that.

PR #102, opened 2024-05-02
fixes #83: translate label of select options

sadiqkhoja:
Closes #83

PR #114, opened 2024-05-13
Fix: client reactivity of translated select item labels

### LINKED ISSUE #83: Apparently selectNode.currentState.valueOptions are not “reactive” to language change
Tried with `ui-solid/fixtures/xforms/select/1-static-selects.xml`, when I am changing the language of the form, the label of `valuesOptions` are not changed.

However, `valueOptions` do change based on xpath filter expression as defined in `2-dynamic-selects-itemsets.xml` fixture

This is working correctly in ui-solid though.

eyelidlessness:
Fixes #83.

Supersedes and closes #102 as well. Differences from that PR:

- Breaks the fixes for these (related, but) distinct bugs into separate commits:

  - Translated _dynamic select itemset labels_ (e.g. `<select1><itemset nodeset="instance('choices')/root/item"><label ref="jr:itext(itextId)"/>…`) are not client-reactive.

  - Translated labels of _static select items_ (e.g. `<select><item><label ref="jr:itext('foo')"/>…`) are **also** not client-reactive.

  The distinction is discussed in #83. I expect having separate commits could be helpful in the future, whenever we need to reason about the slight differences in how each case flows through the engine's reactive internals.

- Adds tests for both bugs, exercising reactivity of these scenarios:

  - Trigger an effect on load, reflecting initial state. This behavior would vary depending on the client's reactive factory and scheduling particulars; we want to be sure it's possible to achieve (even if a client may opt not to exercise it), so it's a distinct test case.

  - Trigger an effect on each change to the form's active language, for each translated label, reflecting the appropriate translation after each update.

  - Do not trigger an effect on repeated reassignment of the currently active language. This is/should be a reactive no-op, both internally and for clients. There's nothing special happening in this PR to achieve that, it relies on the behavior implemented in `createSharedNodeState`. While there are more general ways we could (and maybe should) test that generalization, it felt appropriate to test here because unnecessary translation reactivity is an area that could easily slow down **client performance** across a wide variety of larger forms.

- While the fixes are conceptually quite similar, I think there's some improvement to clarity in where/how some internal-reactive accessors are called. In particular I think it's important to preserve the separation between setup of reactive accessors, and their subsequent/downstream calls.

---

Also in this PR, CI checks that were used to skip certain jobs based on a branch's changes are removed (at least temporarily). In the past we've tolerated CI running seemingly **extraneous jobs**. But in this case, it was observed that no jobs were run for any package downstream from the change. That's risky enough that I think it's appropriate to address in the same PR. If we think it's important to reintroduce such change-based filtering (which we may not!), I'd recommend we file a separate issue so we can discuss how to do it right.

### DIFF
--- a/packages/xforms-engine/src/lib/reactivity/createSelectItems.ts
+++ b/packages/xforms-engine/src/lib/reactivity/createSelectItems.ts
@@ -126,14 +126,16 @@ const createItemset = (
 	selectField: SelectField,
 	itemset: ItemsetDefinition
 ): Accessor<readonly SelectItem[]> => {
-	const itemsetItems = createItemsetItems(selectField, itemset);
+	return selectField.scope.runTask(() => {
+		const itemsetItems = createItemsetItems(selectField, itemset);
 
-	return createMemo(() => {
-		return itemsetItems().map((item) => {
-			return {
-				label: item.label(),
-				value: item.value(),
-			};
+		return createMemo(() => {
+			return itemsetItems().map((item) => {
+				return {
+					label: item.label(),
+					value: item.value(),
+				};
+			});
 		});
 	});
 };
@@ -151,13 +153,11 @@ const createItemset = (
  *   referencing a form's `itext` translations, etc).
  */
 export const createSelectItems = (selectField: SelectField): Accessor<readonly SelectItem[]> => {
-	return selectField.scope.runTask(() => {
-		const { items, itemset } = selectField.definition.bodyElement;
+	const { items, itemset } = selectField.definition.bodyElement;
 
-		if (itemset == null) {
-			return createTranslatedStaticSelectItems(selectField, items);
-		}
+	if (itemset == null) {
+		return createTranslatedStaticSelectItems(selectField, items);
+	}
 
-		return createItemset(selectField, itemset);
-	});
+	return createItemset(selectField, itemset);
 };
--- a/packages/xforms-engine/src/lib/reactivity/createSelectItems.ts
+++ b/packages/xforms-engine/src/lib/reactivity/createSelectItems.ts
@@ -1,3 +1,4 @@
+import { UpsertableMap } from '@odk-web-forms/common/lib/collections/UpsertableMap.ts';
 import type { XFormsXPathEvaluator } from '@odk-web-forms/xpath';
 import type { Accessor } from 'solid-js';
 import { createMemo } from 'solid-js';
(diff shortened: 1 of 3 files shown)


packages/xforms-engine/src/lib/reactivity/createSelectItems.ts
```diff
@@ -26,21 +27,23 @@ const createSelectItemLabel = (
 	});
 };
 
-const buildStaticSelectItems = (
+const createTranslatedStaticSelectItems = (
 	selectField: SelectField,
 	items: readonly ItemDefinition[]
-): readonly SelectItem[] => {
+): Accessor<readonly SelectItem[]> => {
 	return selectField.scope.runTask(() => {
```
sadiqkhoja, line 34: Can you please explain why do we need to run this in `runTask` when `createTranslatedStaticSelectItems` is already called inside of a `runTask` in `createSelectItems`

sadiqkhoja (approved):
`selectField.scope.runTask` at `createSelectItems.ts:34` looks redundant to me, otherwise this looks good.

Tests capture the bug correctly, I ran those tests locally to ensure they fail without the fix

packages/xforms-engine/src/lib/reactivity/createSelectItems.ts
eyelidlessness: In hindsight, I think it might be better:

1. To wrap the body of `createItemset` in a `runTask`.
2. **Maybe** omit the call in `createSelectItems`, but maybe not.

After some time getting more comfortable with these reactive APIs, my thinking on tracking scopes generally goes like this:

- All "downstream" reactivity _must be_ associated with the appropriate tracking scope, as a correctness concern. Where "downstream" is anything producing a subscription. Effectively that means anything performing a read, or anything that causes an effect or effect-like behavior to rerun, or anything that produces a function which does either of those.

- It's best to be able to reason about the effective tracking scope locally wherever possible. Even if we "know" a function is called in another function's tracking scope context, we'd need to reason about multiple functions to know that... and we only know it insofar as it doesn't change. Even though several of these functions are sort of implicitly coupled now, I think we should still try to view them as discrete units of work. While I don't _expect_ to have multiple different calls into `createTranslatedStaticSelectItems` any time soon, it's certainly a possibility.

- Reinforcing the locality point: we almost certainly have quite a few redundant/nested calls of this kind already. They're just not all so close together in the same module, and not so easily identified as a direct 1:1 coupling with a specific logical branch. If we were to go ferret out and remove those redundancies, I think I'd lose clarity and confidence in the correctness of the affected code.

- Until and unless we have evidence to the contrary, there is _nothing logically wrong_ with extraneous/redundant `scope.runTask` calls that make the tracking scope explicit... **as long as they're correct**. The explicitness doesn't make the behavior less clear (and it may make it more clear, it generally does to me).

- Insofar as there's a concern about performance, I really think we should measure things like this so we know what the impact might be. My _hunch_ is that such redundant calls wouldn't have much more overhead than the indirection of the function call itself. After all, the thing it's probably doing is reassigning the same scope object to the same temporary global-internal which it's already assigned to (or a mechanism very much like that).

- Not directly related to the question, but part of my thinking along the same lines: _also wrapping_ state creation and/or writes is harmless at worst, and has potential explanatory/self-documentation benefits (e.g. "these reads are associated with the same scope as these writes").

- - -

Apologies if I sound like a broken record... when it comes to performance hypotheticals, my instinct is almost always going to be "let's measure". I mean, if we **already know** something is likely to perform poorly? Sure, let's address it before it becomes an issue! But I've speculated about likely performance, and been wrong, more than enough times that I've become pretty conservative about clarity/optimization tradeoffs without numbers to justify otherwise.

Lastly, if I've jumped to the wrong conclusion that performance is part of the concern here, I apologize! If there's an actual issue with clarity of the behavior, I do want to make sure we address that.

sadiqkhoja: I am still having hard time understanding the utility of "nested" `runTask`. If it is for code readability then why it is not in `createItemset`?

And if I understand this correctly, you are saying following two code snippet are equal?
```
const createTranslatedStaticSelectItems = (
	selectField: SelectField,
	items: readonly ItemDefinition[]
): Accessor<readonly SelectItem[]> => {
	return selectField.scope.runTask(() => {
		const labeledItems = items.map((item) => {
			const { value } = item;
			const label = createSelectItemLabel(selectField, item);

			return () => ({
				value,
				label: label(),
			});
		});

		return createMemo(() => {
			return labeledItems.map((item) => item());
		});
	});
};
```

```
const createTranslatedStaticSelectItems = (
	selectField: SelectField,
	items: readonly ItemDefinition[]
): Accessor<readonly SelectItem[]> => {
	return selectField.scope.runTask(() => {
		const labeledItems = items.map((item) => {
			const { value } = item;
			const label = createSelectItemLabel(selectField, item);

			return () => ({
				value,
				label: label(),
			});
		});

		return selectField.scope.runTask(() => {
			return selectField.scope.runTask(() => {
				return selectField.scope.runTask(() => {
					return createMemo(() => {
						return labeledItems.map((item) => item());
					});
				})	
			});		
		});
			
	});
};
```

sadiqkhoja: I wrote the above example (nested `runTask` in a single function) to showcase that it's not so useful. Similarly, having nested `runTask` across multiple functions doesn't feel useful to me either. 

It would be better if you 1) remove it from `createTranslatedStaticSelectItems` or 2) add it in `createItemset` and remove it from `createSelectItems`. 

It is even okay if you don't make any change, it's up to you.

eyelidlessness: We had a brief chat about this earlier. To sum up:

- I believe we reached a better shared understanding of the details in the above discussion

- We agreed to move the `scope.runTask` call into `createItemset` and remove it from `createSelectItems`

That change has been made in 93217c0.

PR #116, opened 2024-05-16
Rename `@odk-web-forms` NPM scope to `@getodk`

eyelidlessness:
In so doing:

- Also fix a few copypasta mistakes spotted along the way (references to the wrong package in e.g. README prose/instructions/examples, for the most part)

- Apply any applicable auto-formatting (i.e. changes were applied per file and saved so VSCode would apply the appropriate formatting). Note that this has a greater effect in .vue modules, e.g. stripping a bunch of trailing whitespace that hadn’t already been, adding EOF newlines, etc.

- Rewrap some text (.md files) and code comments. This was not done automatically on save, it’s a manual command I currently run while editing. As such, I may have missed some, but I tried my best not to leave any new formatting weirdness.

Also note that there is one remaining reference to the text “odk-web-forms”: this is an older directory name (before we split `xforms-engine` and `ui-solid` packages), in a GitHub permalink, so it’s correct.


sadiqkhoja:
> Apply any applicable auto-formatting

I ran `yarn format` on main branch, nothing was changed except `LICENSE.md` and `README.md`. Even without `--cache` flag, nothing changes in ui-vue package. Maybe there is some discrepancy between vscode and prettier formatting.

This is just an observation, no action is required for now.

sadiqkhoja (approved):
Looks good to me

PR #113, opened 2024-05-13
Move styles from theme.scss to SFCs

sadiqkhoja:
Moved component specific styles from theme.scss to individual component file.

Issue #73, opened 2024-04-11
Engine/Client interface: the next iteration


eyelidlessness:
I'm opening this issue in anticipation that, following #67 and #69, we'll further iterate on the interface between `xforms-engine` and its various clients (`scenario`, `ui-vue`, `ui-solid`, hypothetical future others). Some of this iteration will likely support refinement and resolution of current features in progress (selects in particular). We'll also surely expand the interface to support the full scope of data types, field appearances, and so on. While it's highly unlikely we'll cover all of these in a single pass, it's good to have a general sense of the paths we might take in their direction.

To kick it off, below captures some thoughts on these topics (lightly edited after brain dumping the same general thoughts on Slack earlier). This isn't intended to fully frame the topic, but as a potential starting point for discussion.

- - -

In discussion around #67, the possibility of splitting up `SelectField` (and potentially `SelectNode` with it) came up. Exploratory prototyping around this, as well as adjacent exploration of `StringNode` subtypes (i.e. whether the node is model-only or has an `input`), led me to rethinki some of my instincts about upcoming changes to the client interface.

At a client level I found it much simpler, and much more obvious to consume a _greater number of distinct node types_ (even if some of them shared a lot of the same structure) rather than fewer types with sub-structural variation. This is type/subtype distinction is currently[^1] represented in #67 as `nodeType` (which narrows between top level `*Node` interfaces) + `nodeVariant` (which I had deferred for a subsequent commit, but envisioned using to narrow variations within the same node type).

For the current feature set, that would look like splitting `SelectNode` into single-value and multi-value interfaces (corresponding to `select1` and `select` XForms elements respectively). It would also look like splitting `StringNode` into `model`-only and `input` interfaces, with that distinction (which clients are already concerned with), plus the additional distinction that it doesn't really make sense for a `model`-only node to have value setters. I expect the same distinctions will play out for other data types (especially those also associated with `input`)—and I expect that we'll probably see one or more corresponding naming conventions emerge naturally as a result.

In all this, I also realized that I'm increasingly doubtful that each node type's `definition` has much value to clients _at all_. More specifically, if we were to surface the meaningful sub-type distinctions within a given `definition` as distinct node types, I doubt that a node's `definition` would have much additional value. Where there _is remaining value_ beyond type/subtype distinction, it's generally places where there are other specific gaps in the client interface. (For instance, we somehow neglected to include a form's title in this first pass on the interface!)

[...]

Issue #80, opened 2024-04-15
Handle errors and warnings in a consistent way


lognaturel:
This issue is to track the broader approach to error handling.

## User stories
As an end user filling out a web form
I want to see clear error messages that are styled consistently with the host app
So that I can understand what happened when something goes wrong

As a host application developer
I want the option to let web forms show their own error messages in a consistent way
So that I don't have to think about error states at all

As a host application developer
I want the option to display error messages in a way that is appropriate for my host app
So that I can create a custom user experience

## Types of errors
* Initialization errors (e.g. invalid XML form def)
  * Could always result in a rejected promise
* Bugs we haven't caught yet
* Form design issues
  * Missing secondary instance or missing subpath into it
  * Typo in expression, especially things like second part of `if` which isn't run at initialization time (but there is/could be static analysis)
  * Unknown functions (maybe we warn?)
  * Divide by 0, other inevitable runtime issues 

## Types of warnings
* More than 1 node when at most one is expected (label, hint, etc) https://github.com/getodk/web-forms/pull/127#discussion_r1624971735

## Notes
* Use error condition return values rather than throwing
  * So we can specify what errors are possible
  * For saner control flow (we'll have to make sure to use catch-alls to handle unknown unknowns) 
* Create a clear interface for the UI to consume error messages, first for our own internal use, and eventually to let clients consume them in a custom way
* Aim for a dependency-injection type of pattern for the client

lognaturel:
Feel free to do what feels most helpful! This issue currently feels like an epic or even just a high-level exploration more than one that can be closed with code. #276 and #202 feel a little closer to being closable with code but you're right that they're also pretty exploratory. We could roll the general theme of prompting for feedback into the user stories. It also relates closely to analytics which is another theme we've touched on now and then.

#202 outlines some specific considerations around the kinds of errors that might happen when external secondary instances are attached (and some apply to form attachments in general). We could capture some of that in this issue and link to the original issue even if it's closed.

lognaturel:
Closing in favor of more targeted issues as specific error states feel more pressing to improve.

Issue #16, opened 2024-01-30
Grid styling


lognaturel:
In Enketo, `grid` can be applied at the form level to modify the layout of an entire form: https://blog.enketo.org/gorgeous-grid/ It was designed primarily to make printable forms and forms for computer data entry and works really well for those cases.

Users choose it because the find the layout and styling visually appealing.

It also makes repeats more compact and somewhat easier to use:
<img width="924" alt="Screenshot 2024-01-29 at 4 09 57 PM" src="https://github.com/getodk/web-forms/assets/967540/6131a475-bec3-496d-afc1-39cceeb4b441">

Sometimes it would be helpful to be able to insert a grid into a form that has a top-to-bottom layout. We have discussed also having a grid appearance to satisfy that need.

The main challenge with Enketo's current grid is that because it's a "theme", a lot of customizations that would make sense to apply to both (e.g. colors) must be specified separately.

PR #127, opened 2024-05-28
Initial support for appearances, body classes

### LINKED ISSUE #56: Represent appearances (parse, client interface)

eyelidlessness:
Closes #56.

This is primarily intended to introduce feature support for `appearance`, from parse through to the engine's client interface, represented by an `appearances` property on each node. Nodes which do not support appearances are assigned `appearances: null` for object shape consistency (as is typically the case across node types, with some exceptions for `RootNode`; more on that below).

Nodes that do support appearances currently include:

- `GroupNode`
- `RepeatInstanceNode`\*
- `RepeatRangeNode`\*
- `SelectNode`
- `StringNode`\*\*

\*: Presently both repeat node types share the same appearances definition. There are some JSDoc comments and commit notes discussing the details of this.

\*\*: A `StringNode`'s appearance will be present if the node's `definition` has a non-null `bodyElement` (presently an `InputDefinition`). This may be less confusing if we pursue additional node types, as discussed in #73 (i.e. we could better distinguish between a "string node" that's associated with an `<input>`, versus a "string node" that's only associated with a model node and/or `<bind>`).

The value for each node's `appearances` are produced by a node-specific `TokenListParser` into a `TokenList`[^1] with _roughly_ the same behavior:

- For each of that node's appearances (as defined as a whitespace separated list in an `appearance` attribute, on that node's body element), `node.appearances.$appearanceName` will produce `true` (and will produce `false` for any\*\*\* appearance _not present_ on the element in the form definition). For clients which are concerned with **specific appearances** (i.e. branching on those specific appearances to produce differences in rendering, presentation, interactive behavior), this is likely to be the most convenient way to access a node's appearances.

  \*\*\* Excepting aliases, see the last bullet point in this list.

- The `appearances` object is also an `Iterable`, yielding each such appearance from its definition in the form (e.g. `for (const appearance of node.appearances) { ... }`). It's likely this will have more utility for testing purposes.

- There is some variance in the `appearances` **types** for each respective node type. These are primarily intended to provide additional developer convenience in clients. For each underlying body element type, there are a set of _"canonical"_ (known, first class, explicitly documented, and so on). These are specified as part of body parsing logic, and those appearances are promoted in the node `appearances` type to provide additional editor support (e.g. autosuggest on dot access and/or when referencing an iterated appearance).

- **All form-defined appearances** are passed through in the runtime value, even if they are not declared as "canonical". These are not included in type suggestions, because they're not known upfront. But if present in the form definition, non-canonical/unknown appearances will also produce `true` on dot access and will be yielded on iteration.

- **Node specific aliases:** mostly included as a proof of concept (for now), each node type's `TokenListParser` may specify an `aliases` option, mapping any particular appearance to another "canonical" appearance. The behavior of aliases is **additive**: if a form definition specifies an appearance which is aliased to a "canonical" appearance, both the specified appearance and its corresponding "canonical" appearance will produce `true` on dot access, and both will be yielded on iteration (with the "canonical" appearance yielded first). This is currently only used to map a single select appearance (`search` -> `autocomplete`), but is intended to demonstrate the generalization. Ideally we will expand on these so that the engine can act as a source of truth for all known/supported appearance types, without pushing that responsibility separately onto every client.

[^1]: As discussed in more detail in the JSDoc for `TokenListParser`/`TokenList`, these names are a reference to a web standard `DOMTokenList`, as it is very similar conceptually. But I'm totally open to bikeshedding these names if they don't feel right.

#### Elimination of "repeat group"/"repeat" pairing in the parsed body representation

There has been some lingering awkwardness in the form parsing logic around the concept of "repeat groups". Earlier development introduced a pre-parse, in-memory normalization to ensure that every `<repeat nodeset>` element is wrapped in a corresponding `<group ref>` (where the `nodeset`/`ref` are the same. This is the most common production from XLSForms as produced by pyxform. The intent was to simplify the parsed **data model** by producing a consistent structure for repeats regardless of whether the form itself specifies a `<group ref><repeat nodeset>` pair. Even at the time, it was noted that the resulting data model was somewhat awkward, with the parsed representation of a repeat's body always including a corresponding pair of objects for what is ultimately treated as a singular _thing_.

Starting on the work to support appearances, this awkwardness resurfaced. I had also recognized that the awkwardness overlaps with **other coming feature work** (support for `jr:count` and `jr:noAddRemove` come to mind, and from recent prototyping on those I anticipate this change will bring some meaningful simplicity).

#### Rename remaining references to repeat "sequence" to "range"

Before #67, the concept of a set of zero-or-more contiguous repeat instances was called a "sequence". In that PR, we settled on the name "range" (`RepeatRange` &co). Since I was already doing some minor refactoring on the above "repeat group" concept, it felt appropriate to tackle this lingering naming inconsistency. So all references to the same "bag or vat of repeat instances" are now consistently referenced with the "range" terminology.

#### Open question regarding repeat appearances

As I understand the spec, appearances may be specified only on form controls and groups. In practice, I have noted many form fixtures (particularly in Enketo) define appearances directly on `<repeat>` elements. With the aforementioned change to normalize "repeat groups" into a singular concept, both of these would be treated as interchangeable:

```xml
<group ref="/data/rep" appearance="foo"><repeat nodeset="/data/rep"/></group>
<group ref="/data/rep"><repeat nodeset="/data/rep"/ appearance="foo"></group>
```

My instinct is that this is the most likely expectation, although that can also be framed as a question here. So: is that what users would most likely expect?

Perhaps more ambiguous in intent: what would users expect if **both nodes** specify appearances? I.e.

```xml
<group ref="/data/rep" appearance="..."><repeat nodeset="/data/rep"/ appearance="..."></group>
```

This is presently treated as an error. It's conceivable that:

- They should be treated differently **no matter what**
- They should be treated differently **iff the containing group has a `<label>`** (this is currently a presentational difference in the `web-forms` Vue UI client)
- They should be treated as a single definition **iff they match**
- They should be treated as a single definition **and merged/concatenated**
- ??? There's probably some other option I haven't even thought of

### Landing sequence

I'm opening in earnest, as I understand it will unblock quite a bit of UI client work. I'm opening it as draft, because I would like to land these other PRs first:

- #110 - so we can also add integration tests for this functionality
- #113 - as UI client progress is also blocked on this


eyelidlessness:
Somehow I forgot to mention in the initial writeup that this also adds support for body classes. The functionality is so conceptually similar that it made sense to me to generalize it and tackle both together.

It feels a bit awkward at the `RootNode` client interface to have `appearances: null; classes: EffectivelyTheSameDataType;`, but they're named different things in the spec and have slightly different semantic implications, so I erred on the side of "let's not be too cute with naming".

lognaturel:
> is that what users would most likely expect?

Yes, that sounds right to me.

> what would users expect if both nodes specify appearances?

I think starting with an error is a good idea because it will allow us to know whether it actually happens. As far as I know, the only appearance that can apply to groups or repeats is `field-list` so it shouldn't really be an issue currently. Unless maybe Enketo has a concept of `minimal` to show a group or repeat as collapsed initially? This is something I suddenly have a fuzzy memory of. Does it ring a bell, @eyelidlessness? I don't think that changes my answer here, though.

> this also adds support for body classes

I never registered that `class` was mentioned in https://getodk.github.io/xforms-spec/#body-attributes! As far as I know Collect would ignore those. Does Enketo apply them? Either way, seems something useful for Web Forms to support and the approach sounds good to me. I don't think it has to be restricted to only the `body` element necessarily.

lognaturel:
Re `minimal`: it looks like XForms recommends supporting `minimal` and `compact` for every control type as described in https://www.w3.org/MarkUp/Forms/wiki/Metadata_Appearance. I can't find that exact wording in the XForms 1.1 spec but I do see that most controls explicitly list full|minimal|compact as supported appearances. No action needed now I don't think but wanted to mention I'd noticed this. It could be relevant to upcoming work like maybe figuring out what we want to do about grids (https://github.com/getodk/web-forms/issues/16)

eyelidlessness:
> > is that what users would most likely expect?
> 
> Yes, that sounds right to me.
> 
> > what would users expect if both nodes specify appearances?
> 
> I think starting with an error is a good idea because it will allow us to know whether it actually happens.

👍

> As far as I know, the only appearance that can apply to groups or repeats is `field-list` so it shouldn't really be an issue currently. Unless maybe Enketo has a concept of `minimal` to show a group or repeat as collapsed initially? This is something I suddenly have a fuzzy memory of. Does it ring a bell, @eyelidlessness? I don't think that changes my answer here, though.

The other case I found is `table-list`.

The design is such that anything defined will be passed through, so anything that Enketo supports will be present, it just won't necessarily be explicit in the type suggestions.

> > this also adds support for body classes
> 
> I never registered that `class` was mentioned in https://getodk.github.io/xforms-spec/#body-attributes! As far as I know Collect would ignore those. Does Enketo apply them? Either way, seems something useful for Web Forms to support and the approach sounds good to me. I don't think it has to be restricted to only the `body` element necessarily.

Enketo uses body `class` to specify `pages`, as well as themes (`theme-*`; this also includes `theme-grid`). If I recall, it also applies all classes as defined in the transform, regardless of whether they have any special meaning. For the latter purpose, I agree it would potentially make sense to apply consistently across node types with UI. For the former, it feels more like the kind of functionality we could align with a hypothetical `appearance` on body (or have some other mechanism entirely, but consistent use of `appearance` would make the difference between node types feel a lot less awkward/go away).

lognaturel:
> The design is such that anything defined will be passed through, so anything that Enketo supports will be present, it just won't necessarily be explicit in the type-driven suggestions.

Right, that's perfect.

> Enketo uses body class to specify pages, as well as themes (theme-*; this also includes theme-grid).

Ah, right, of course.

lognaturel:
One thing that just came to mind is that labels cannot be directly nested in repeats according to the spec: https://getodk.github.io/xforms-spec/#repeats That’s why repeats are generally nested in a group bound to the same node. 

I think this just means that some of the work in the first commit won’t be used in practice. Seems ok to leave in for now, though.

packages/xforms-engine/src/body/appearance/selectAppearanceParser.ts
```diff
@@ -0,0 +1,38 @@
...
+		'columns-pack',
+		'autocomplete',
+
+		// TODO: these are `<select1>` only
+		'likert',
+		'quick',
+		'quickcompact',
+		'map',
+		// "quick map"
+	],
```
sadiqkhoja, line 32: ❓I am assuming that appearance (W1, W2... Wn) related to theme-grid of Enketo will be handled separately later on.

packages/xforms-engine/src/body/appearance/selectAppearanceParser.ts
```diff
@@ -0,0 +1,38 @@
...
+export const selectAppearanceParser = new TokenListParser(
+	[
+		// From XLSForm Docs:
+		'compact',
+		'horizontal',
+		'horizontal-compact',
+		'label',
+		'list-nolabel',
+		'minimal',
+
```
sadiqkhoja, line 12: I see following appearances from XLSForm template missing here:

- no-buttons
- list
- image-map

sadiqkhoja (approved):
Looks good to me. I really like the approach taken for handling appearance aliases 🎉

There are few inline comments, one about missing appearances of `select` may require a change, rest are just questions.

packages/xforms-engine/src/body/appearance/selectAppearanceParser.ts
lognaturel: Ah yes, `compact` is deprecated. It was split up into `horizontal-compact` and `no-buttons` which can be combined if desired.

packages/xforms-engine/src/XFormDOM.ts
```diff
@@ -41,7 +41,83 @@ const normalizeBodyRefNodesetAttributes = (body: Element): void => {
...
+				`Failed to normalize conflicting "appearances" attribute of group/repeat "${ref}"`
+			);
+		}
+
+		repeat.setAttributeNS(namespaceURI, nodeName, value);
+	}
+};
+
+const normalizeRepeatGroupLabel = (group: Element, repeat: Element): void => {
+	const groupLabel = Array.from(group.children).find((child) => {
```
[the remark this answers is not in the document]
eyelidlessness: Probably warn, and probably everywhere we look for 0-1 nodes (all label lookups, hints also come to mind). My understanding of the spec is that only one is “properly supported”, which to me strongly implies _ignoring_ >1.

Feels like a good issue to file, and/or category of things to add to a general discussion of how we’ll convey errors/warnings/notices about form definitions to clients.

packages/xforms-engine/src/body/appearance/selectAppearanceParser.ts
eyelidlessness: If there are any known/canonical appearances (or aliases to them) that we want to add right away, please feel free to add them in! My hope is that this approach makes that trivial. I’d be happy to add these when my power is back, but you’re welcome to add them at your convenience as well.

eyelidlessness:
> One thing that just came to mind is that labels cannot be directly nested in repeats according to the spec: https://getodk.github.io/xforms-spec/#repeats That’s why repeats are generally nested in a group bound to the same node.
> 
> I think this just means that some of the work in the first commit won’t be used in practice. Seems ok to leave in for now, though.

Ah I thought this was more common than it is. Turns out the only case I could find in Enketo was one I added myself 🙃, presumably when I was less familiar with these details of the spec.

lognaturel:
Have added changeset (that was fun), over to you @sadiqkhoja to either merge or align select appearances, up to you. 😊

packages/xforms-engine/src/XFormDOM.ts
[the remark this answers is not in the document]
lognaturel: Added to https://github.com/getodk/web-forms/issues/80

Issue #128, opened 2024-05-31
Adjust UI for mobile and tablet


sadiqkhoja:
Walk few steps towards making the web-forms responsive to different screen sizes, specifically for mobiles and tablets:

- Remove question card
- Make the questions container full width
- Move Form title in the row of print and language controls
- Show hamburger icon on mobile screens

[Figma](https://www.figma.com/design/KFi9hIQdrRgo0rLqIC7Sg8/ODK-user-flows?node-id=4952%3A19348&t=7iwnrIOJ3ZZvV7MW-1)
<img width="946" alt="Screenshot 2024-06-20 at 12 17 13 PM" src="https://github.com/getodk/web-forms/assets/7871340/a3d56380-8a8e-4263-a27f-dae0f1700c33">

**Design tweaks:**
- Title should be smaller on mobile (I think we may need to tweak the font sizes in general so it feels responsive. It might be too small on desktop right now)
- [Dialog ](https://www.figma.com/design/KFi9hIQdrRgo0rLqIC7Sg8/ODK-user-flows?node-id=5328%3A118478&t=stYrE0uZdWmXMJLz-1)for changing the language on mobile could wider so it doesn't feel so squished, and we can remove the 'X' because we have cancel at the bottom. Rounded edges might also make it more consistent with the tablet/desktop experience. 
-  Sent button needs more spacing on the right

alyblenkin:
Looking good! I think we can add more spacing between the radio button and text. As well as more spacing within the "Save" button so it's wider and easier to tap on mobile. 

![Image](https://github.com/getodk/web-forms/assets/7871340/aaa41995-8089-4922-8a75-dde87647a125)

I'm going to add a placeholder issue for adjusting the font size and spacing. I think we will need to play around with this after we implementing the outlines to the options.

PR #132, opened 2024-06-12
feat #128: first step towards responsive layout

sadiqkhoja:
#128


packages/web-forms/src/components/FormLanguageDialog.vue
```diff
@@ -0,0 +1,102 @@
...
+
+const selectedLanguage = ref<FormLanguage>();
+
+const initSelectedLanguage = () => {
+	if(!props.form.currentState.activeLanguage.isSyntheticDefault) {
+		selectedLanguage.value = props.form.currentState.activeLanguage;
+	}
+}
+
+initSelectedLanguage();
```
sadiqkhoja, line 21: maybe do a check in parent component about isSyntheticDefault

packages/web-forms/src/components/FormHeader.vue
```diff
@@ -28,4 +97,32 @@ defineProps<{title: string}>();
...
+	// this adds border to parent div so that bottom shadow of immediate child
+	// is visible.
+	border-bottom: 3px solid #fff;
+	
+	// to show ellipsis in the heading
+	// https://css-tricks.com/using-flexbox-and-text-ellipsis-together/
+	min-width: 0;
+
+	> div {
+		box-shadow: 0px 1px 2px 0px #0000004D;
```
sadiqkhoja, line 111: move color codes to a global variable

packages/web-forms/src/components/FormLanguageDialog.vue
eyelidlessness: One thing that I didn't anticipate when we were chatting about this yesterday, but also worth mentioning: when a form is **not translated**, the mobile presentation currently moves the print button into a menu that has no other items. That print button takes up the same space as the hamburger, so it's not really saving any space.

I don't think we necessarily need to change anything on this now, but it's one more thing to add to the discussion of what to do with the print button in mobile. And what we do if/when there are more header controls like view toggle.

packages/web-forms/src/components/FormHeader.vue
```diff
@@ -1,17 +1,86 @@
...
 		<template #content>
-			<h1>{{ title }}</h1>
+			<!-- TODO/q: should the title be on the definition or definition.form be accessible instead of definition.bind.form -->
+			<h1>{{ form.definition.bind.form.title }}</h1>
 			<!-- last saved timestamp -->
 		</template>
-	</Card>
+	</PrimeCard>
+
+	<!-- for mobile and tablet -->
```
eyelidlessness, line 48: Also not something we need to act on right now (unless you want to explore it!): if we are sure we want to preserve the single-row presentation and condense horizontal space, this feels like a potentially excellent use case for [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_queries). Using it well for this might have other design implications (i.e. we'd need to know how much space is needed upfront before we can query it).

The nice thing about container queries here would be the possibility of triggering space-saving as needed, which might allow more functionality to stay visible for e.g. forms with shorter titles.

packages/web-forms/src/components/FormHeader.vue
```diff
@@ -28,4 +97,32 @@ defineProps<{title: string}>();
...
+	min-width: 0;
+
+	> div {
+		box-shadow: 0px 1px 2px 0px #0000004D;
+	}
+
+	h1 {
+		padding-left: 10px;
+		text-overflow: ellipsis;
+		white-space: nowrap;
```
eyelidlessness, line 117: We discussed this on the call yesterday, and I don't want to lose the context: if we're truncating text, we might want to think about how to support expanding it. There's potentially a component generalization to consider there.

Adding to that... do we actually want to truncate form title text? In terms of use of space, might it be a better user experience to let it flow, possibly with smaller font size.

packages/web-forms/src/components/FormHeader.vue
```diff
@@ -28,4 +97,32 @@ defineProps<{title: string}>();
...
+		overflow: hidden;
+		font-size: 1.5rem;
+	}
+	
+	.btn-menu{
+		color: #000;
+		font-size: 1.5rem;
```
eyelidlessness, line 124: - This color seems to affect only the hamburger icon, and doesn't seem to be dependent on screen size. Can we apply the color directly to that element (with a class identifying it as such)?

- The font size does not seem to affect anything, as far as I can tell. The same size is applied with higher precedence on `.p-button.p-button-icon-only.p-button-rounded`.

packages/web-forms/src/components/FormHeader.vue
```diff
@@ -1,17 +1,86 @@
...
+		width: 2.5rem;
+		min-width: 2.5rem;
+		font-size: 1.5rem;
+
+		&:hover{
+			background: var(--primary-100);
+		}
+		&:active, &:focus {
+			background: var(--primary-50);
+		}
```
eyelidlessness, line 81: Unclear where to put this comment, but this is where it's most visible: the hamburger menu looks great in its default state, but when these background colors are applied it shows the background all the way to the right edge of the viewport. This is a minor nitpick, but it feels a little unpolished and maybe worth a little finesse.

packages/web-forms/e2e/vue.test.ts
```diff
@@ -76,7 +76,7 @@ test('All forms are rendered and there is no console error', async ({ page, brow
 		const langChanger = page.getByLabel('change language');
 
 		if ((await langChanger.count()) > 0) {
-			await langChanger.click();
+			await langChanger.first().click();
```
eyelidlessness, line 79: This feels like something that could churn quite a lot. I'm generally not fond of littering UI with classes just for testing purposes, but it seems reasonable to make an exception for:

- things that are global singletons (like this is)

- E2E testing specifically, where it's already challenging to relate the test code to the intended structure it's targeting

eyelidlessness (commented):
It's exciting to see this coming along! I have a few comments. I don't feel super strongly about most of them, and some probably warrant broader team discussion and possible UI/UX design iteration.

packages/web-forms/src/components/FormHeader.vue
sadiqkhoja: Good suggestion. For now I have run with the existing media query, I will try to use it next time we have any changes in the UI of the Form header.

packages/web-forms/src/components/FormHeader.vue
sadiqkhoja: I had a discussion with Aly and she suggested to add equal spacing on the right side as there is on the left side, which is 10px.

packages/web-forms/src/components/FormHeader.vue
sadiqkhoja: Add css variables for colors and box-shadow in theme.scss

packages/web-forms/e2e/vue.test.ts
sadiqkhoja: Added classes to make it easier to select language changer in e2e test.

packages/web-forms/src/components/FormHeader.vue
sadiqkhoja: good catch, I have removed it.

packages/web-forms/src/components/FormHeader.vue
sadiqkhoja: I have removed the truncation, instead Form title now wraps around.

sadiqkhoja:
While making changes related to color codes, I went on a tangent and removed the back button from the demo and used history API for navigation. I wanted to remove `back` to make it easy for me to compare the result with the Figma. I hope this change is acceptable.

.changeset/128-responsive-layout-init
```diff
@@ -0,0 +1,5 @@
+---
```
eyelidlessness, line 1: When we discussed this, I said I don't think the file name matters. But I do wonder if it matters whether there's a `.md` suffix. The changesets CLI always creates files with that suffix, and I noticed the same for the suggested filename when I used the GitHub flow in #135.

packages/web-forms/src/components/FormHeader.vue
```diff
@@ -1,20 +1,115 @@
...
-			<h1>{{ title }}</h1>
+			<!-- TODO/q: should the title be on the definition or definition.form be accessible instead of definition.bind.form -->
+			<h1>{{ form.definition.bind.form.title }}</h1>
 			<!-- last saved timestamp -->
 		</template>
-	</Card>
+	</PrimeCard>
+
+	<!-- for mobile and tablet -->
+	<div class="flex lg:hidden align-items-center smaller-screens">
```
eyelidlessness, line 62: I was curious to see how wrapping looks for forms with long titles. Pretty good! I think we may want to iterate later on...

- More consistent alignment of controls (icons in larger views, hamburger menu in smaller). The vertical centering makes those a moving target. I think we'll probably want some design guidance on that.

- Maybe slightly reducing text size/weight over a certain character length? (Honestly the weight feels pretty heavy regardless of length, wrapping.)

packages/web-forms/src/components/FormHeader.vue
eyelidlessness: Now that you mention this, it caught my eye that quite a lot of measurements in the styles are in `px`. I don't think we should worry about this now, but I would like to give some thought into where relative units make sense for accessibility/scaling flexibility.

eyelidlessness (approved):
Looking great! I'm glad to land the demo navigation functionality here. As mentioned when we talked about it on a call, I had made a similar change in the `ui-solid` demo at one point. I'd be cautious about expanding any kind of client navigation _beyond_ the demo, especially using native APIs directly (they're notoriously error prone). Anyway, it feels especially nice to be able to reload (or make changes and see hot reload) without getting bumped back to the form list.

packages/web-forms/src/components/FormHeader.vue
sadiqkhoja: Good points, I am quoting these in #139

packages/web-forms/src/components/FormHeader.vue
sadiqkhoja: Let's talk about this in Slack

.changeset/128-responsive-layout-init
sadiqkhoja: renamed it.

packages/web-forms/src/components/FormHeader.vue
sadiqkhoja: We talked about using `rem` instead of `px` in Slack and setting different font-size of root for different screen sizes. Tracking it in https://github.com/getodk/web-forms/issues/143

Issue #87, opened 2024-04-22
Implement jr:count


lognaturel:
* There's an early spike from the December demo
* Non-relevant repeat instances should be entirely hidden: https://github.com/enketo/enketo/issues/1028

eyelidlessness:
Should have closed this in #187.

PR #135, opened 2024-06-19
Several repeat-related fixes

eyelidlessness:
This PR includes a handful of fixes and improvements addressing issues found in #110, generally with a common repeat-related theme.

The commit notes go into some detail of each fix. I'll highlight some high level details here as well.

1. ad607dc Fix `InconsistentChildrenStateError` cases that currently affect `scenario`. This fix shouldn't affect the `web-forms` Vue UI client (as it only affects clients using Solid), but it was important to fix first so that the other fixes won't be obscured by introducing this same error.

2. 0212b9f Fix a variety of issues where behavior is correct for the first repeat instance in a range (or within that repeat instance, or referencing into it) but incorrect for subsequent repeat instances. The change doesn't address **all** such cases, but it is probably further reaching in real usage than the set of newly passing tests suggests. I expect that it will address quite a few cases where repeat functionality feels broken/incomplete.

    Note that this is followed by another commit (57883df) marking several form init smoke tests failing that did not previously. This may seem like a regression, but I don't think it really is. Those tests had previously satisfied a broad check for producing an error, and their previous passing status was essentially a coincidence. We'll likely address these more meaningfully in some of the work related to #80.

3. 19016e6 (and eea137d) Fix a few issues related to the order of reactive state initialization for each form node. The mechanism of the fix involves moving the initialization of `reference` (i.e. the effective XPath reference to the node, which may change for repeat instances and their descendants) as well as shared bind computations—`relevant`, `readonly`, `required`—earlier in node initialization... effectively before any other reactivity for all non-root nodes.

    The immediate benefit, as visible in the main commit for the fix, is that we now correctly clear default values of non-relevant nodes on form load. While _that_ isn't specifically repeat-related, the other benefit of this change is that it eliminates the aspect of form/repeat instance init where certain aspects of reactivity are not initially "ready" when they need to be initially referenced. As such, and reflected in followup commits, this fix also resolves any remaining asynchrony[^1] during form init (8642ed7). For repeats, this also fixes edge cases where certain aspects of reactive state lag after creating now repeat instances (90e1ad6).

    This broadly reaffirms our current intent that form state is reconciled fully, synchronously, for every form action (until a need to deviate arises).

4. b44beec Fix some cases where `relevant` state was incorrectly inherited when applied to repeat instances themselves. The nature of this issue was that `RepeatRange` nodes (which are a fiction for our runtime data model, not real nodes in the primary instance) were incorrectly evaluating `relevant` expressions intended for their `RepeatInstance` children.

    The first (correctness) problem with this was that the expression would be evaluated against the wrong context, as `RepeatRange.contextNode` references the repeat range's parent element in the primary instance tree. The second problem: if this expression evaluated to `false` for a `RepeatRange`, that would be inherited from **all of its `RepeatInstance` children** even if the expression properly evaluates `true` for the instance itself. These issues are resolved.

    There are also `scenario` tests which consider the relevance of the repeat range itself... or at least, they check the presence of a prompt to create new instances of the repeat range, and we've ported those tests to check for that range's relevance. We now determine this `RepeatRange` relevance state by a heuristic, where the range itself is relevant if either of these cases are true:

      - Does the `RepeatRange` have >=1 relevant `RepeatInstance` children?
      - Does the `RepeatRange` have 0 `RepeatInstance` children at all?

    It's entirely possible this is the wrong heuristic. It's also entirely possible that there is a more appropriate way to represent the presence or absence of a repeat "prompt" besides relevance state. I am certain that `scenario` will need different logic when we introduce support for `jr:count` (#87).

[^1]: Besides resource fetching, which is inherently async and which is at least presently handled **before** the now-synchronous `Root` initialization.

### DIFF
--- a/.changeset/neat-spoons-pull.md
+++ b/.changeset/neat-spoons-pull.md
@@ -1,0 +1,12 @@
+---
+"@getodk/scenario": patch
+"@getodk/xforms-engine": patch
+---
+
+Several repeat-related fixes:
+
+- Fix: most cases of inconsistent children state in Solid-based clients
+- Fix: many cases of incomplete functionality on/within N > 1 repeat instances
+- Fix: computation of bind states (`relevant` especially) before values, properly clear non-relevant default values
+- Fix: timing inconsistency of some computations on init, adding repeat instances
+- Fix: application of `relevant` state where expression is on repeat itself
--- a/packages/scenario/test/form-definition-validity.test.ts
+++ b/packages/scenario/test/form-definition-validity.test.ts
@@ -127,7 +127,7 @@ describe('TriggerableDagTest.java', () => {
 					// exceptionRule.expectMessage("Cycle detected in form's relevant and calculation logic!");
 
 					const init = async () => {
-						return Scenario.init(
+						await Scenario.init(
 							'Some form',
 							buildFormForDagCyclesCheck(bind('/data/count').type('int').calculate('. + 1'))
 						);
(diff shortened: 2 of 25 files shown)


sadiqkhoja (approved):
Looks good to me 👍. 

Really liked the simplification around `Root` construction 💜.

Issue #86, opened 2024-04-22
Make select choices easier to select


lognaturel:
**User story**

As a user on a tablet or desktop
I want to be able to easily select answers (without accidentally selecting something) 
so that I can quickly fill out my form without having to directly interact with the radio button or checkbox

Given I want to select an option on desktop
When I hover I see a light blue fill 
Then when I select the outline is ODK blue with a light blue fill  

Notes:
- We also decided if it's less than 50% then the container is only half and if it's more than 50% it's full width
- Needs discussion on likert scale, auto-filtering a list, selecting SVG shapes

[Figma](https://www.figma.com/design/KFi9hIQdrRgo0rLqIC7Sg8/ODK-user-flows?node-id=4952%3A19348&t=DGLHIxWze8CuXA8g-1)

PR #142, opened 2024-06-26
Select appearances

### LINKED ISSUE #141: Implement Select Appearances
Implement appearances for `select_one` and `select_multiple` in the UI client.

### Out of scope:
- Options with images (media)
- `map` and `image-map` appearances.

sadiqkhoja:
Closes #141 and #86 

- Implemented minimal, autocomplete, no-buttons, likert, list, list-nolabel, label, columns, column-pack, column-n for Select1
- Implemented minimal, autocomplete, no-buttons, list, list-nolabel, label, columns, column-pack, column-n for SelectN
- Implementation of map and image-map appearances will be handled separately
- Dynamic width of select options is achieved using javascript, we need to find a way to do it using css if possible
- Original SelectControl.vue has been decomposed into Select1Control and SelectNControl.
- Widgets directory contains low level components, mostly combination of a label and a primevue component
- Appearance directory contains wrapper components to acheive desire UI layout


packages/web-forms/src/components/widgets/LikertWidget.vue
```diff
@@ -0,0 +1,86 @@
...
+import type { SelectNode } from '@getodk/xforms-engine';
+import RadioButton from '../widgets/RadioButton.vue';
+
+defineProps<{question: SelectNode}>();
+
+</script>
+
+<template>
+	<div class="likert">
+		<RadioButton :question="question" />
```
eyelidlessness, line 11: I was a little surprised by the clickable area with the `likert` appearance. At first it seemed like the grey line was specifically made clickable, and then I realized that it is the full `<label>` containing each option... which even extends all the way to the left side of the leftmost option (and right/rightmost). It's really not obvious at all where the clickable area for one option ends, and the next begins. Worth considering:

- For better experience with pointer inputs (mouse), give some sort of indication of the target with a hover state
- For better experience with touch input, leave at least a small gap of inert area between each option

packages/web-forms/src/components/widgets/LikertWidget.vue
```diff
@@ -0,0 +1,86 @@
...
+			outline: none;
+		}
+
+
+		.p-radiobutton {
+			position: relative;
+			z-index: 10;
+		}
+
+		span {
```
eyelidlessness, line 46: Is there a reason not to assign a class to target this selector?

packages/web-forms/src/components/widgets/MultiselectDropdown.vue
```diff
@@ -0,0 +1,87 @@
...
+const props = defineProps<{ question: SelectNode, style?: string}>();
+defineEmits(['update:modelValue']);
+
+const setSelectNValue = (values: SelectItem[]) => {
+	for(const v of props.question.currentState.value){
+		props.question.deselect(v);
+	}
+	for(const v of values) {
+		props.question.select(v);
+	}
```
eyelidlessness, line 14: Awkward! This seems like it's kind of forced by the `PrimeMultiSelect` API? Yeah, it seems like we definitely want a better engine API for this kind of thing.

packages/web-forms/src/components/widgets/LikertWidget.vue
eyelidlessness: (Same goes for any bare tag name selector, really.)

packages/web-forms/src/components/widgets/SearchableDropdown.vue
```diff
@@ -0,0 +1,46 @@
...
+</template>
+
+<style scoped lang="scss">
+@import 'primeflex/core/_variables.scss';
+
+.dropdown {
+	width: 100%;
+	border-radius: 10px;
+	border-color: var(--surface-300);
+	border-radius: 10px;
```
eyelidlessness, line 36: What's going on here?

packages/web-forms/src/directives/FullWidth.ts
```diff
@@ -0,0 +1,22 @@
...
+	if (binding.value === false) return;
+
+	const items: Element[] = Array.from(el.children);
+
+	if (items.find((item) => item.scrollWidth * 2 > el.parentElement!.offsetWidth)) {
+		el.classList.add('full-width');
+	} else {
+		el.classList.remove('full-width');
+	}
+};
```
eyelidlessness, line 15: So... when we discussed this on chat the other day, we talked about maybe holding off if it feels like it's not doing what we want. After playing with it some, I feel pretty strongly that that's the best option. The behavior as implemented is surprising in a lot of ways:

- It's based entirely on comparing measurements at the time of mount, which are determined in part by the available space at that same time
- It adjusts appearance based on media query, rather than actual space used by the options
- It does not adjust based on resizing (which alleviates my performance concern, but doesn't seem to do what I'd expect)
- If resizing from a smaller viewport to larger, it makes otherwise narrow options full width in ways I wouldn't expect
- I didn't test it, but I don't expect it will adjust based on form language change

I think it might be worth holding off on this for now, and coming back to it with fresh eyes if it feels strictly necessary. If so, maybe we can pair on it and think about some of these nuances together.

eyelidlessness (changes requested):
Overall looking really great! I'm excited to see the appearances in action, and they generally do what I expect as I interact with them. This really does make the overall product progress feel a lot more tangible!

As mentioned on a call the other day, I'm also thrilled to see some of the component organization coming into focus. I think it might be worth aligning naming/hierarchy/compositional concepts throughout ODK (design, overlap with Collect, docs). That might be a pretty big task, but it's feeling plausible with this progress, and I think it'll be worth it to have that kind of a holistic view.

I have made a few comments inline. Some are just questions. I have some concerns about the `FullWidth` directive, commented in detail there.

packages/web-forms/src/components/widgets/LikertWidget.vue
sadiqkhoja: naming things is hard 🙃

packages/web-forms/src/components/widgets/MultiselectDropdown.vue
sadiqkhoja: yeah, if `select` method of future `SelectMultipleNode` could take `SelectItem[]` as an argument that would be great.

packages/web-forms/src/components/widgets/SearchableDropdown.vue
sadiqkhoja: I must have overlooked it, removed the duplicate style.

packages/web-forms/src/directives/FullWidth.ts
sadiqkhoja: The directive is listening to mount and update events, so it works with language change as well. I didn't handle resizing event, it felt more like a developer thing than end user use case.

In any case, I have removed the directive and created a separate issue to tackle this https://github.com/getodk/web-forms/issues/156. Yes, pairing on it will be helpful.

packages/web-forms/src/components/widgets/LikertWidget.vue
sadiqkhoja: Incorporated the suggested changes.

packages/web-forms/src/components/widgets/CheckboxWidget.vue
```diff
@@ -0,0 +1,87 @@
...
+			{{ option.label?.asString }}
+		</span>
+	</label>
+</template>
+
+<style scoped lang="scss">
+.value-option {
+	display: flex;
+	align-items: center;
+	outline: 1px solid var(--surface-300);
```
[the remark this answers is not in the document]
sadiqkhoja: yeah, changing border-width was causing whole Form to move up and down, even adjusting margins was causing odd effects.

eyelidlessness (approved):

PR #152, opened 2024-06-28
fix: include css in the built js file

sadiqkhoja:
After decomposing styles from theme.scss into SFC in #113, build process started generated separate css file instead of generating just a single `index.js` file. Because styles defined in SFC file scan't be accessed in `index.ts` as we do for other style files, I ended up using a plugin that post-processes styles and adds them in index.js file as `IIFE` that adds `<style>` tags in the document head. If we prefer `document.adoptedStyleSheets` instead I can use `injectCodeFunction ` of the plugin.

### DIFF
--- a/.changeset/swift-ducks-divide.md
+++ b/.changeset/swift-ducks-divide.md
@@ -1,0 +1,5 @@
+---
+"@getodk/web-forms": patch
+---
+
+fix: include css in the built js file, test build to prevent repeat regressions
--- a/packages/web-forms/build-preview/index.html
+++ b/packages/web-forms/build-preview/index.html
@@ -1,0 +1,85 @@
+<!doctype html>
+<html lang="en">
+	<head>
+		<meta charset="UTF-8" />
+		<meta name="viewport" content="width=device-width, initial-scale=1.0" />
+		<title>ODK Web Forms: Test Build</title>
+	</head>
+	<body>
+		<div id="app"></div>
+		<script type="module">
+			/** Basic form definition, to load for validation of build product */
+			const formXml = /* xml */ `<?xml version="1.0"?>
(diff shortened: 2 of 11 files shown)


eyelidlessness:
Can you give some guidance on how to evaluate this change? Maybe a sense of how to evaluate the build product generally.

I’d like to think about how we can automate checking for regressions of this kind. This is the second time we’ve effectively had a broken build due to unbundled styles, and it would be nice to make that less likely to happen again.

We effectively test the build product of the other published packages by testing downstream usage. I think it would give us some peace of mind about the build for this leaf package if it’s more clear how to validate it.

Even if automating that validation isn’t the highest priority, it might be as simple as running E2E tests against the build product (which seems like probably what we’d want anyway), and then adding a check that at least some stable subset of styling is effective.

sadiqkhoja:
I like your idea of running E2E test against built product. 

Another idea, quickest and not foolproof, is to build the package in CI and assert that there is only index.js and no style.css produced.

eyelidlessness:
For future reference, this morning we discussed E2E testing options for the build product. I offered to do a quick spike on it, and it turned out to be relatively straightforward to do (8ba788d). That spins up two servers for the E2E tests to hit:

- Dev demo (which extant test depends on to iterate through all fixtures)
- Simple preview server exercising the build

The latter is tested for presence of a specific CSS custom property defined in a Vue component, and then tests overriding another custom property which that same component uses to style the page. All of these checks validate that the build includes component-defined styles, and leaves some additional context if/when we want to expand testing of that sort.

sadiqkhoja (commented):
Have reviewed the recent commits, looks good to me 🚀

eyelidlessness (approved):
@sadiqkhoja approved, I approve. Let's go!

PR #158, opened 2024-07-08
CI: check formatting of the files

sadiqkhoja:
Currently there are 17 files in the main branch that violates prettier rules. I have added a step in CI to check the formatting of the files, so that we can catch those before merging.


eyelidlessness (changes requested):
How would you feel about ignoring everything in `packages/scenario/resources`, and reverting those changes? I'd prefer to keep their formatting as-is from JavaRosa, so we can easily see where changes come from upstream if/when we update anything.

eyelidlessness (approved):

PR #162, opened 2024-07-11
Added feature matrix in README.md

sadiqkhoja:
Added a quick script to regenerate it based on feature-matrix.json.

Preview of README.md: https://github.com/getodk/web-forms/tree/feature-matrix

@eyelidlessness  Feel free to update `feature-matrix.json` to reflect work-in-progress features with 🚧

PR #163, opened 2024-07-11
Post-release (0.2.0) dependency updates & related

eyelidlessness:
Not totally fuss-free, but less fuss than the last one!

More details in commit notes.


sadiqkhoja (approved):
Tried this PR locally, all seems to be working!

eyelidlessness:
Thanks for a first look @sadiqkhoja! Since you mentioned the conflict with #162, I just went ahead and reverted the `scripts` package removal and added a no-op `test` script to it. That way it won't cause similar hassles next time we do project wide dependency updates.

If you're still good with the change, please feel free to merge!

PR #166, opened 2024-07-12
Improve dependency resolution, XPath static analysis; support `current()` in computations, repeat-based itemsets, relative body references

eyelidlessness:
Whew, that title is a mouthful! As it suggests, there's a fair bit going on here. But it's all interrelated: all of the functional improvements (fixes and features alike) effectively call in to the same underlying abstractions (840937b).

Working out from the lowest conceptual level to the highest:

1. A simpler, more general, and more correct, approach to querying aspects of a given XPath expression's syntax—by [traversal of parsed XPath syntax trees](https://github.com/getodk/web-forms/commit/3180324f4e11f63f4637131f6b9226da34dd9a56#diff-d4084c8038e55eeedb29f038ec030f62a3d45f9a494878a2b6bf92021c1281a9)

2. [Semantic application](https://github.com/getodk/web-forms/commit/3180324f4e11f63f4637131f6b9226da34dd9a56#diff-b4d9c386b1f7bee4beb508d52cf07bfb56e90fd4b2bc17845163cfa6075af4f4) of that approach, used to identify various kinds of XPath path expressions and sub-expressions, including node-set returning functions; similar application replaces existing analysis of function calls, e.g. to identify translation expressions.

3. A more thorough (and thoroughly tested) approach to [path resolution logic](https://github.com/getodk/web-forms/commit/3180324f4e11f63f4637131f6b9226da34dd9a56#diff-286bad8b54c454947c62c9baa57dc2d522890652a75050a40e5979e15ba7aece). This is the most substantive aspect of the change, and I'll go into more detail about the approach below.

4. Analysis of [path predicates](https://github.com/getodk/web-forms/commit/3180324f4e11f63f4637131f6b9226da34dd9a56#diff-78aab177d5b3f3c41f8cf7bce7e5fb9f1c54a99cb4b7b1233faa12e90f632413) to similarly identify path references within those predicates, along with the sub-path context for each such reference. This will also be discussed in more detail below, as it relates to the next and previous points.

5. Combination of all of these aspects to perform [dependency analysis](https://github.com/getodk/web-forms/commit/3180324f4e11f63f4637131f6b9226da34dd9a56#diff-1c3fb12758b4aa18d981dac9f98839d6171803062422b6988863e9c9e312425c) on arbitrary sub-expressions, where (to the extent possible) the product is a set of contextualized `nodeset` references, which are ultimately used to establish reactive subscriptions for each of a form's computational expressions. More on this below as well.

## Note on XPath syntax terminology

There will be several references to aspects of XPath syntax, both in the text below and throughout the changes in this PR. These references always refer to one (or more) of:

- concepts directly named in [XPath 1.0](https://www.w3.org/TR/1999/REC-xpath-19991116/)
- concepts derived from that specification, typically where the spec does not explicitly name a concept
- convenient categorical concepts which reference multiple such underlying concepts (e.g. quite a few references in the change to "Step-like")

If any reference to such a concept is unclear, feedback on that is welcome and encouraged. My hope is that the end product of this PR is unambiguous on spec and spec-derived concepts, and easily understood if/when we iterate on these foundations downstream.

## Note on where this fits in the form initialization process

I want to make sure it's absolutely clear that all of the work described herein happens while **parsing a form**. All of this analysis and resolution is done once, upfront, before a form's instance state is initialized. Since #154, the distinction will hopefully be clearer for this and future work (as the code is clearly structured under the `parse` source directory). But it feels important to make the distinction clear now, as there's been some ambiguity in the past about when certain logic is effective, and what that might imply for e.g. performance, internal API usage, etc.

## Path resolution

Broadly, the approach to path resolution involves:

1. Build a partially flattened, list-like representation of the unresolved path expression, where each list item is one of these parsed syntax nodes:

    - Step (`StepNode`)
    - `/` (as a reference to the root node in an AbsoluteLocationPath; not explicitly named in the XPath grammar; internally named `AbsoluteRootLocationPathNode`)
    - `//` (shorthand for the sub-path expression `/descendant-or-self::node()/`, itself a Step plus the prefix to a following Step; also not explicitly named in the XPath grammar; internally named `RelativeStepSyntaxLiteralNode` [which is probably not ideal!])
    - FilterExpr (which is a reference to many different aspects of XPath syntax, but for our usage generally corresponds to a node-set returning function)

2. If a particular context path exists, build the same partially flattened representation and concatenate (e.g. `[...context, ...unresolved]`).

3. Iterate through the (now-contextualized, where applicable) flattened representation, resolving (where possible):

    - `current()` to the context path itself
    - `.` (and other syntax representations of `self` references) to the preceding Step-like path segment, by omitting the self reference
    - `..` (and other syntax representations of `parent` references), by omitting the parent reference as well as its preceding Step-like path segment

4. Serializing the remaining path segments. Once again, I'll give more detail on this under **Dependency analysis** below.

## Dependency analysis

The approach to dependency analysis replaces a much less thorough/mature approach, but has the same goal: produce a set of `nodeset` references identified in an arbitrary XPath expression. The previous approach did some portion of what will be described below, naively cutting a variety of corners (largely introduced in our early spike on the viability of this project).

Broadly, the more thorough and mature approach involves:

- With a given context expression (where one is known/available from a current or earlier parsing step), and
- Given an arbitrary XPath expression

Find all of the sub-expressions referencing a path in that arbitrary expression.

1. For every such path sub-expression found:
    1. Resolve the path itself, as described in **Path resolution** above
    2. **Predicate resolution:** for each path segment, and for each of its predicates: find any further path references in the predicate expression. For each predicate path sub-expression found: resolve **that** path such that `context()` references the same context as above, and `.`/self references the current path segment/Step context
3. **Serialization:** For all resolved path expressions, serialize the path segments _omitting predicates_.

This last point is important: as the work here happens during form parsing, the resolved dependencies it produces are (and have been for some time) `nodeset` references, as in references to an XForm _model node_. We will likely revisit this in one form or another, but this is the way all existing dependency analysis works (and has been improved in various ways under the same approach). This is also the mechanism that enables the same approach to fix tracking relative references as well as `current()` references.

Most importantly: the product of all this analysis is a set of `nodeset` references used in engine internals to identify `SubscribableDependency`s for reactive subscriptions. **The resolved expressions are never evaluated to produce results for the source expressions from which they were resolved.**

To really emphasize how much this is building on a set of assumptions already at work, [this is how little changed](https://github.com/getodk/web-forms/commit/5fe2a758761df9f97491610594b654ec8bdd8aef) to make these changes available to downstream instance state logic.

## Relative body references

This aspect of the change (3e425f1) largely builds on the same **path resolution** and **serialization** foundations above, with two key differences:

- `current()` (as the initial path segment of a given body reference) is treated as equivalent to `.`/self; I believe this is the expectation/intent, at least as represented by the ported JavaRosa `scenario` test which it fixes

- serialization of resolved paths preserves predicates; this ensures, for instance, that itemset nodesets retain any of the dynamic filtering logic defined in the form (and allows downstream dependency analysis from there)

An important note: while this aspect of path resolution is applied in nearly every aspect of body parsing where a reference might be expected, the `<value>` and `<label>` subexpressions of an `<itemset>` are exempted. Where these are defined as relative, it is important to preserve that so their evaluation is contextualized to the also-dynamic items produced by the `<itemset nodeset>` expression.

## `Scenario` updates

The first commit in this change takes a slightly different tack from #150 when it comes to the `scenario` tests with implicit repeat creation. Instead of implementing such implicit repeat creation, this change:

- Warns when implicit creation is likely expected at a given point in a test's execution, allowing the test to be updated to explicitly create the missing repeats
- Adds an API that allows such explicit additions easily identifiable, so similar changes may be adopted by JavaRosa at that project's convenience

Beyond that... well, there's a whole bunch of newly passing tests!


packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
```diff
@@ -0,0 +1,337 @@
...
+			expression: 'parent::*/bar',
+			expected: ['../bar'],
+		},
+
+		{
+			description: '// special cases',
+
+			contextNodeset: null,
+			expression: '//. | //./foo | //.. | //../foo',
+			expected: ['//.', '//foo', '//..'],
```
lognaturel, line 313: Why isn't the last one not `//../foo`?

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
```diff
@@ -0,0 +1,337 @@
...
+			expected: ['instance("foo")/bar', 'sel1/bat', 'sel1/quux'],
+		},
+
+		{
+			description:
+				'Relative context; instance() and relative path; extracted, instance call returned as-is, relative paths resolved to context',
+
+			contextNodeset: 'sel1',
+			expression: 'instance("foo")/bar and (bat or ../quux)',
+			expected: ['instance("foo")/bar', 'sel1/bat', 'sel1/../quux'],
```
lognaturel, line 108: 'sel1/../quux' -- subtle (in that it can't be `quux`) Can this actually be exercised in a form? Maybe in a repeat? Not super important but it piqued my curiosity.

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
```diff
@@ -0,0 +1,337 @@
...
+			expected: ['instance("foo")/bar', '/data/sel1/bat', '/data/quux'],
+		},
+
+		{
+			description:
+				'Context is instance("id") descendant. Expression references another instance (resolved as-is); relative sibling and relative child (resolved from context); relative sibling and child with current() calls (likewise resolved from context); absolute reference (resolved as-is)',
+
+			contextNodeset: 'instance("foo")/bar',
+			expression:
+				'if(position() > 2, instance("quux")/zig, ../zag | ./zag | current()/zag | current()/../zag) or /beep//boop',
```
lognaturel, line 126: `current()/zag` resolves to the same as `./zag` and same with the ../ pair. Is that intentional? Seems separating them would reduce the risk of this masking a future bug (seems very low given other test coverage but possible)

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
```diff
@@ -0,0 +1,337 @@
...
+				'Context is absolute. Expression includes multiple references to instance($some-id); each is extracted and returned as-is',
+
+			contextNodeset: '/data/quux',
+			expression: 'instance("first")/foo or instance("second")/bar',
+			expected: ['instance("first")/foo', 'instance("second")/bar'],
+		},
+
+		{
+			description:
+				'No context. Expression with instance("id") reference has its own parent steps resolved',
```
lognaturel, line 146: Should context be null?

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
```diff
@@ -0,0 +1,337 @@
...
+			description: 'Resolution without context: parent steps',
+
+			contextNodeset: null,
+			expression: '/data/foo/../bar and foo/bar/../bat or /data/quux/../zig',
+			expected: ['/data/bar', 'foo/bat', '/data/zig'],
+		},
+
+		{
+			description:
+				'Arbitrary FilterExpr with either a trailing Step or a Predicate (without any specialized handling for the FunctionName in use)',
```
lognaturel, line 163: This is exercising https://github.com/getodk/web-forms/pull/166/files#diff-b4d9c386b1f7bee4beb508d52cf07bfb56e90fd4b2bc17845163cfa6075af4f4R168 I believe. And so if a function were known and not one of the special ones any paths used as arguments would be resolved and returned? Maybe that's a test to add.

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
```diff
@@ -0,0 +1,337 @@
...
+			expect(actual).toEqual(expected);
+		});
+	});
+
+	// TODO! This was explored but it introduced a variety of complexities beyond
+	// an already fairly large chunk of work on this pass. Note also that
+	// parentheses can have surprising effects on a path expression's semantics!
+	// (For instance: see note on predicates and ordering in
+	// https://www.w3.org/TR/1999/REC-xpath-19991116/#node-sets).
+	it.fails('resolves parenthesized portions of sub-expressions', () => {
```
lognaturel, line 328: Feels low priority but good to have a test for!

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
```diff
@@ -0,0 +1,337 @@
+import { describe, expect, it } from 'vitest';
+import { resolveDependencyNodesets } from '../../../src/parse/xpath/dependency-analysis.ts';
+
+describe('Dependency analysis', () => {
```
lognaturel, line 4: I have tried hard to think of missing cases but this is feeling really complete to me. Maybe something with a known function? I would expect any arguments that are paths to be identified as dependencies and resolved I think but https://github.com/getodk/web-forms/pull/166/files#diff-cb3208eef759b81b33d971b0f6ccea7c8f2c0c041fdc98b2be1d9b515855cc13R163 makes me question that.

packages/xforms-engine/test/parse/xpath/semantic-analysis.test.ts
```diff
@@ -0,0 +1,91 @@
...
+				},
+				{
+					description: 'itext is substring of FunctionName, not full name',
+					expression: 'some-fn-with-itext-in-name()',
+					expected: false,
+				},
+
+				{
+					description:
+						'we do not currently check prefixes for function call analysis, typical usage has common namespace mapping',
```
lognaturel, line 51: This seems ok.

lognaturel (commented):
I found this quite overwhelming at first but started with the unit tests and those really helped me get oriented. After understanding the tests, I went back to the code and it felt much more comfortable. I read through it relatively quickly mostly to see whether it made me think of untested cases. The general approaches seem sound to me.

A few small questions inline.

> [this is how little changed](https://github.com/getodk/web-forms/commit/5fe2a758761df9f97491610594b654ec8bdd8aef) to make these changes available to downstream instance state logic.

Very cool!

> current() (as the initial path segment of a given body reference) is treated as equivalent to ./self

It still feels largely accidental that this is supported at all in JR but the interpretation is consistent.

packages/xforms-engine/src/parse/xpath/dependency-analysis.ts
```diff
@@ -0,0 +1,123 @@
...
+	 * that we will want similar special casing when we reintroduce explicit graph
+	 * cycle analysis.
+	 */
+	readonly ignoreReferenceToContextPath?: boolean;
+
+	/**
+	 * Treats the literal sub-expression `null` as if it were a keyword. This is a
+	 * deviation from XPath (where `null` would be a RelativeLocationPath
+	 * referencing a child node by that name). However, in real-world usage, some
+	 * forms evidently use `null` as a keyword representing a null and/or blank
```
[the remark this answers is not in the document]
eyelidlessness: I'm glad you asked about this! It gave me an opportunity to really think about the underlying reason it existed in the first place, and whether it makes sense to carry over into this change. In short: I don't think this is necessary anymore! I'll double check a couple things locally to be absolutely sure, but I expect to be able to remove this.

Since you asked for an example form, the cases I could find were in whova. See [here](https://github.com/getodk/web-forms/blob/cfde5ea53085cdae22ffcfa274cb8a48dfc3d699/packages/scenario/resources/smoketests/whova_form.xml#L14627C1-L14628C1) and [here](https://github.com/getodk/web-forms/blob/cfde5ea53085cdae22ffcfa274cb8a48dfc3d699/packages/scenario/resources/smoketests/whova_form.xml#L14634C1-L14635C1) (last argument to the inner `if()` in both).

- - -

I do have to ask whether a `'null'` string value is expected to be a special case. This would be very surprising to me!

Taking the whova usage as an example, the conditions that reference `null` as a NameTest will produce a blank value when they hit that condition. If those were a `'null'` string Literal, in strict XPath semantics, they'll produce the that literal string value.

A quick search of the `resources` ported from JavaRosa doesn't reveal any usage of `'null'` that I could find. A quick search in JavaRosa itself has a few hits for the string literal, but none that look like treating it as a special XPath value.

packages/xforms-engine/src/parse/xpath/semantic-analysis.ts
```diff
@@ -0,0 +1,290 @@
...
+ * Predicate to determine if a FilterPathExpr (as currently produced by
+ * `tree-sitter-xpath`) is one of:
+ *
+ * - `current()`
+ * - `current()/...` (where `...` represents additional steps)
+ * - `current()//...` (^)
+ *
+ * @todo XPath grammar technically also allows for `current()[some-predicate]`,
+ * and our `tree-sitter-xpath` grammar/parser also allow for this. But
+ * `@getodk/xpath` types do not currently acknowledge this possibility.
```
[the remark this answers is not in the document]
eyelidlessness: Yes, it's possible in a form, it's possible anywhere. It's part of the XPath grammar: [FilterExpr](https://www.w3.org/TR/1999/REC-xpath-19991116/#NT-FilterExpr). The grammar itself is overly permissive—it allows Predicates on things you can't predicate, such as numbers and string literals—but the inferrable intent is specifically to allow Predicates on node-set returning FunctionCalls and on parenthesized Exprs (implicitly those producing node-sets as well).

I'm curious what you mean by:

> The expression would need to appear in a place where it's bound to a repeat because only repeats can be filtered in this way. It would have to be in a relevance or read-only expression.

Are you saying that we should add special conditions to enforce this, restricting this aspect of XPath grammar to these specific areas of XForms functionality? Or just that expressions like that would only be produced (e.g. by XLSForm/pyxform) in those cases.

packages/xforms-engine/src/parse/xpath/semantic-analysis.ts
```diff
@@ -0,0 +1,290 @@
...
+
+/**
+ * Predicate to determine if a FilterPathExpr (as currently produced by
+ * `tree-sitter-xpath`) is one of:
+ *
+ * - `instance("id")`
+ * - `instance("id")/...` (where `...` represents additional steps)
+ * - `instance("id")//...` (^)
+ *
+ * @todo XPath grammar technically also allows for `current()[some-predicate]`,
```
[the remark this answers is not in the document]
eyelidlessness: Right. Missed in copypasta. I think it'd be better both this and [the other case](https://github.com/getodk/web-forms/pull/166#discussion_r1684703023)) to reference the grammar names for the syntax, rather than the specific function names. Which may also help clarify some of the questions raised in that other thread.

The point in both cases is that they're not accounting for predicates which may be present, because the current `SyntaxNode` types don't allow for it.

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
eyelidlessness: It's resolved to `//foo`. The result is set-like, so that result corresponds to both `//./foo` (second sub-expression) and `//../foo` (fourth/last sub-expression) in the UnionExpr. (This surprised me for a second too, when I wrote the test case.)

But now that I think about it a bit more, I suppose there's a slight chance I'm wrong about `//foo`, `//./foo`, `//../foo` all being equivalent. I think the self step case has to be, but there's a bit of nuance in the parent step case: it means "`foo` which has one or more child elements"[^1]. I don't know how much difference it'd make for this change now, as neither would actually resolve anything with our current downstream usage. But I'm happy to reconsider the handling of this case now, or add a comment about this nuance for future reference in case it becomes an issue when downstream can surface it.

[^1]: This kind of thing is why I find XPath expressivity both impressive and frustrating. It's kind of like regex that way.

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
eyelidlessness: I don't know if I can think of a form off the top of my head.

Connecting this to the subtlety I noticed in the [previous thread](https://github.com/getodk/web-forms/pull/166#discussion_r1684725331), I think both cases could be resolved more completely by moving the hierarchic check into a Predicate:

- I think this case would be semantically equivalent to (say) `quux[count(../sel1) > 0]`
- The other would be equivalent to `//foo[count(.) > 0]`

Moving those to predicates would definitely be more involved, and I'm not inclined to even think about it in this scope. But assuming this broad approach to dependency tracking survives long term (which I'm iffy on), I wouldn't be surprised if edge cases come along where we'd think about an approach like that.

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
eyelidlessness: I'm not sure whether you're asking:

1. Whether it's correct that each pair resolves to the same path expression
2. Whether, when multiple sub-expressions resolve to the same path, we should produce the same path for each sub-expression respectively versus the current set-like behavior

If the former, I think it's correct but feel free to clarify in case I'm missing something!

If the latter, I don't know what benefit we'd get from producing the same path multiple times. The set-like behavior is intentional: we're determining which dependencies are referenced in an arbitrary expression, within a particular context... and then using those resolved paths to look up `*Node`s in instance state, establishing reactive subscriptions to those nodes to determine when the expression should be re-evaluated. Producing multiple same-path strings would probably be pure overhead.

We might reconsider the set-like aspect in, say, a context like dev tools or other form analysis tooling, where it could be useful to match found sub-expressions up with their specific resolved paths. But even then we'd probably also want to produce the unresolved sub-expression in the same result (and possibly even character indexes or some other way to identify where in the overall expression the result is found).

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
eyelidlessness: It's exercising that, yes.

> And so if a function were known and not one of the special ones any paths used as arguments would be resolved and returned? Maybe that's a test to add.

If you're asking for a test for specifically resolving the paths as Arguments of a FunctionCall—something like `any-non-nodeset-fn(foo, /bar, //etc/., ..)`—does [this case calling `if(...)`](https://github.com/getodk/web-forms/pull/166/files#diff-cb3208eef759b81b33d971b0f6ccea7c8f2c0c041fdc98b2be1d9b515855cc13R126) suffice? If not, can you be more specific about what kind of expression you don't see covered?

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
eyelidlessness: Probably something we should get some data on. I don't know how much parenthesized Exprs are used in real world.

I do know that I'd want more coverage of it in `xpath` as well, particularly around the note linked in this comment.

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
eyelidlessness: Same response as there! I don't know if `if()` is what you're looking for. But it's specifically covered because I wanted to make sure at least one FunctionCall was included, and because that function in particular has been called out in the past for its conditional Argument evaluation. The point of including it is that all of the Arguments are analyzed, regardless of the branch they'll hit when evaluated.

packages/xforms-engine/test/parse/xpath/semantic-analysis.test.ts
eyelidlessness: Yeah... it's not my favorite thing, but I really wouldn't want to connect up form-derived evaluation context to static analysis unless/until we have a compelling reason, like a bug in the wild.

packages/xforms-engine/src/parse/xpath/dependency-analysis.ts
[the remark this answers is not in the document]
lognaturel: > See [here](https://github.com/getodk/web-forms/blob/cfde5ea53085cdae22ffcfa274cb8a48dfc3d699/packages/scenario/resources/smoketests/whova_form.xml#L14627C1-L14628C1) and [here](https://github.com/getodk/web-forms/blob/cfde5ea53085cdae22ffcfa274cb8a48dfc3d699/packages/scenario/resources/smoketests/whova_form.xml#L14634C1-L14635C1) (last argument to the inner if() in both).

🤯 Looks like the intent is indeed to have a blank value and because it had the desired outcome for the reasons you described it stayed in.

> I do have to ask whether a 'null' string value is expected to be a special case.

No, I don't think so!

> I expect to be able to remove this.

🎉🎉🎉

packages/xforms-engine/src/parse/xpath/semantic-analysis.ts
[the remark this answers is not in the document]
lognaturel: I just mean that it would only be actually useful/desirable to filter a nodeset with more than one node, I think. Basically the todo feels relatively low priority unless there's a use case I'm missing.

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
lognaturel: Ahh, yes, I understand. How about making it `//../bar` so it's a distinct case? I didn't notice there were only three items in the expected set. I got tripped up by that in another place as well.

> "foo which has one or more child elements"

Parent, right? But yeah, I don't think it needs to be addressed.

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
lognaturel: > I think this case would be semantically equivalent to (say) quux[count(../sel1) > 0]

Agreed

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
lognaturel: I don't mean to suggest any change to the code, just to the test so that each case is separate in the result set. It's not very important, it's really just to make reading the test a little easier. Here I wasn't confused reading the test like I was with the `//../foo` case above.

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
lognaturel: Sweet! I just missed it, probably because my brain doesn't tend to accept that `if` is a function.

lognaturel:
> Happy to confirm that if you think it'd be valuable.

No, I think it's very reasonable!

Maybe consider the naming changes to the two tests that resulted in overlap in expected results if you see why I found those confusing. Certainly doesn't feel blocking to me.

@sadiqkhoja will likely want to take a look at some point but I will give approval in case we may want to merge before then.

lognaturel (approved):

packages/xforms-engine/src/parse/xpath/semantic-analysis.ts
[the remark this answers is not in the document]
eyelidlessness: It's mostly acknowledging that this is a gap in `@getodk/xpath`, both at the type level for `SyntaxNode` variants and (consequently) at the implementation of that aspect of the spec. Agree it's low priority! I don't want to lose the context completely, as it's one of very few areas `xpath` deviates from the spec.

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
eyelidlessness: > Ahh, yes, I understand. How about making it //../bar so it's a distinct case? I didn't notice there were only three items in the expected set. I got tripped up by that in another place as well.

Despite both of us being tripped up on the same thing, I think it's valuable to have tests specifically around the set-like behavior. My instinct is:

- where there are cases that exhibit the set-like behavior, break them up so there is a clearer connection between each individual sub-expression and the resolved path

- add one or more dedicated cases specifically around the set-like behavior, where there are _only_ multiple sub-expressions which would resolve to a single path

Do you think that would be clearer for your understanding too?

> Parent, right? But yeah, I don't think it needs to be addressed.

Neither actually! I started to walk through the steps to explain my reasoning, but that helped me realize it _is equivalent_ to `//foo` (probably retreading similar reasoning when I originally arrived at the same result):

`//../foo` is equivalent to:
`/descendant-or-self::node()/../foo` is equivalent to:

1. All nodes in the document except attributes and namespaces (which are not children/descendants):
    - `/`
    - every element
    - every text node
    - every comment node

2. The subset of 1 which are parents of any other subset of 1:
    - `/`
    - every element with any child node

3. The children of 2 which pass the `foo` NameTest

That's equivalent to `//foo`, because `foo` can't be a child of anything filtered out in step 2.

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
eyelidlessness: Yeah. It might be a little easier to reconcile if you've worked in a lisp or something else where `if` can be implemented in userland (but even in a lisp it'll be a macro, not a function).

packages/xforms-engine/test/parse/xpath/dependency-analysis.test.ts
lognaturel: > Do you think that would be clearer for your understanding too?

Yes!

> That's equivalent to //foo, because foo can't be a child of anything filtered out in step 2.

Ah, ok.

sadiqkhoja:
My questions/confusions/observations after going through unit tests:

1- Why following two are different. If contextNodeset is relative, shouldn’t current() to resolve as relative nodeset:
```
const contextNodeset = 'whatever';
const expression = '/data[current() = 1]/bar';
const expected = ['/data/bar', 'whatever']; // shouldn't this be '/data/whatever'??
//vs
const contextNodeset = 'whatever';
const expression = '/data[whatever = 1]/bar';
const expected = ['/data/bar', '/data/whatever'];
```

2- I was assuming arguments to the functions would be added as dependencies and this would return ‘/bat’ as well?
```
description: 'Arbitrary FilterExpr with either a trailing Step or a Predicate (without any specialized handling for the FunctionName in use)',
contextNodeset: null,
expression: 'foo("bar", /bat)/quux and zig()[zag]',
expected: ['foo("bar", /bat)/quux', 'zig()', 'zig()/zag']
```
3- Parent’s sibling works when contextnode is absolute but not when it is relative:
```
const contextNodeset = 'foo';
const expression = '../bar';
const expected = ['bar']; // fail
//vs
const contextNodeset = '/foo';
const expression = '../bar';
const expected = ['/bar']; // pass
```
4- How are we handling text()? I am thinking parent of the text node (i.e. the element node) should be added to the dependencies
```
const contextNodeset = '/data/foo';
const expression = 'child::text()';
const expected = ['/data/foo'];
```
5- shouldn’t this be an error - text node can’t have children, right?
```
const contextNodeset = '/data/child::text()';
const expression = 'child::node()';
const expected = ['/data/child::text()/child::node()'];
```
6- I think isTranslationExpression should return true for the following expression:
```
/foo[jr:itext("id") = 1]
```
7- would be nice to have one or two unit tests for `ignoreReferenceToContextPath`

packages/xforms-engine/test/parse/xpath/semantic-analysis.test.ts
```diff
@@ -0,0 +1,106 @@
...
+describe('Semantic analysis', () => {
+	describe('FunctionCall', () => {
+		describe('translation', () => {
+			interface IsTranslationExpressionCase {
+				readonly description: string;
+				readonly expression: string;
+				readonly expected: boolean;
+			}
+
+			it.each<IsTranslationExpressionCase>([
```
sadiqkhoja, line 17: I was expecting following case to pass:

```
{
  description: 'argument is a string',
  expression: 'jr:itext(string(1))',
  expected: true,
},
```

sadiqkhoja (changes requested):
Overall this is an excellent work 🎉

I have added few small suggestions inline, otherwise this is great.

packages/xforms-engine/src/parse/xpath/dependency-analysis.ts
```diff
@@ -0,0 +1,99 @@
...
+	contextNodeset: string | null,
+	expression: string,
+	options: PathResolutionOptions = {}
+): readonly string[] => {
+	const { ignoreReferenceToContextPath } = {
+		...defaultPathResolutionOptions,
+		...options,
+	};
+
+	const contextNode = getPathExpressionNode(contextNodeset ?? '"not a path, null fallback"');
```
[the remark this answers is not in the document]
eyelidlessness: Yeah, this was a little wonky and I think I meant to change it but forgot 😅. We do need to accept a null `contextNodeset` (and call it where that occurs[^1]), because the results can still be contextualized as the ancestor tree is resolved. In any case, as you've seen in the test suite, there is meaningful (albeit more limited) resolution logic applied even in the absence of a `contextNodeset`.

I think I'd originally done it this way to minimize the amount of branchy logic, but that was in an earlier pass which had... quite a bit more branchy logic in quite a few more places. (A huge chunk of the time spent on this change was focused on trying to tame the implementation complexity, so it could potentially be understood and maintained by others as well as my future self.) I think making it more explicitly conditional instead of this weird fallback is probably the most appropriate solution for now.

I do think it's likely we'll be able to make `contextNodeset` non-optional eventually, probably in the (er) context of fixing other aspects of groups without a `ref`.

[^1]: At a glance, I think the only current case is the expression parts of a `<label>` in a `<group>` without a `ref` attribute. And we have a variety of work to do to improve that case.

packages/xforms-engine/src/parse/xpath/syntax-traversal.ts
```diff
@@ -0,0 +1,129 @@
...
+ * This may be useful for performing semantic analysis on expressions, for
+ * instance identifying when an expression **is a FunctionCall**, and producing
+ * the {@link SyntaxNode} representing that FunctionCall.
+ *
+ * In contrast, this would produce `null` for an expression **containing a
+ * FunctionCall** in some sub-expression position (e.g. the call to `position`
+ * in `foo[position() = 2]`).
+ */
+export const findTypedPrincipalExpressionNode = <const Type extends AnySyntaxType>(
+	types: readonly [...Type[]],
```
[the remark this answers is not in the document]
eyelidlessness: Whoops, actually meant this to be...

```suggestion
	types: readonly [Type, ...Type[]],
```

To ensure at least one is provided. Otherwise it does weird things with inference if you mistakenly pass an empty array.

packages/xforms-engine/src/parse/xpath/syntax-traversal.ts
```diff
@@ -0,0 +1,129 @@
...
+	xpathNode: XPathNode
+): TypedSyntaxNode<Type> | null => {
+	const [first, ...rest] = collectTypedNodes(types, xpathNode);
+
+	if (first == null || rest.length > 1) {
```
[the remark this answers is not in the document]
eyelidlessness: Yikes good catch!

packages/xforms-engine/test/parse/xpath/semantic-analysis.test.ts
eyelidlessness: My first thought looking at this comment was "well yeah, and it should also accept anything else returning or producing a string". And then it occurred to me that... that's everything. Even the test cases where we currently reject some argument types (e.g. Number), because the `xpath` evaluator will cast them to a string. So the more I think about this, the more I think we shouldn't bother checking anything about the Argument's expression at all, except that it's there.

sadiqkhoja: true, I was not sure if we expect non-core functions to be doing explicit casting of the arguments.

sadiqkhoja (approved):

PR #167, opened 2024-07-12
Fix/vitest flakiness

sadiqkhoja:
On main branch when I run e2e tests, they work for every other execution. So run the test with `--ui` to have an opportunity to see what's happening. With that I saw `[504 (Outdated Optimize Dep)]` in browser console, google that error and found this https://stackoverflow.com/questions/75883720/504-outdated-optimize-dep-while-using-react-vite and https://vitejs.dev/guide/dep-pre-bundling.html#browser-cache

So I added `--force` to playwright.config.ts in webServer.command, which fixed the e2e tests. 

That gave me a clue that issue in component tests for webkit could also be because of vite's optimizer, so tried disabling that with `optimizeDeps.noDiscover` in vitest.config.ts. That solved the problem locally and in CI as well, as many times I have tried.

I have also reversed `sleep` and double execution of test-browser:webkit that introduced in [#d1d31ea](https://github.com/getodk/web-forms/commit/7e787086829c7d1007407fda59fc3cb6a9399203)

### DIFF
--- a/packages/web-forms/vite.config.ts
+++ b/packages/web-forms/vite.config.ts
@@ -37,6 +37,10 @@ export default defineConfig({
 			},
 		},
 	},
+	optimizeDeps: {
+		// Without this flag, `test:e2e` fails for every other execution
+		force: true,
+	},
 	css: {
 		postcss: {
 			plugins: [


eyelidlessness:
Ironically, I now see the simplified commit failed in CI too. Which has me questioning my local repro approach, and doubting the mitigation rollback altogether. Maybe for now we just keep the `--force` flag and consider more holistic changes down the line?

sadiqkhoja:
I saw the failure of simplified commit as well, the difference is `noDiscovery: true` vs `force: true`. 

`noDiscovery` totally disables the optimizer completely and `force` ignores the previously cached optimized dependencies. I am reading [this document](https://vitejs.dev/config/dep-optimization-options.html#optimizedeps-disabled). 

I am thinking to use simplified commit but add `noDiscovery: true` as well - keeping both flags, what do you say?

eyelidlessness:
> I saw the failure of simplified commit as well, the difference is `noDiscovery: true` vs `force: true`.
> 
> `noDiscovery` totally disables the optimizer completely and `force` ignores the previously cached optimized dependencies. I am reading [this document](https://vitejs.dev/config/dep-optimization-options.html#optimizedeps-disabled).

Right. As I mentioned, when I tried these options before I opened #165, I had still observed flakiness (reduced, but still observable). It's also worth looking at [`optimizeDeps.holdUntilCrawlEnd`](https://vitejs.dev/config/dep-optimization-options.html#optimizedeps-holduntilcrawlend). That said...

> I am thinking to use simplified commit but add `noDiscovery: true` as well - keeping both flags, what do you say?

My instinct now is that the mitigations from #165, while adding some noise, are relatively harmless and probably worth keeping around for now. Maybe we make an exception to only apply the delay in CI (with an `env` check or whatever). The double run is already restricted to CI, should only occur on first-run failure, and will still fail if there's a **meaningful** failure.

PR #168, opened 2024-07-15
Fix: use of AxisName as abbreviated child NameTest Step

eyelidlessness:
Fixes use of `child` (and other [AxisName](https://www.w3.org/TR/1999/REC-xpath-19991116/#NT-AxisName) strings) as a [NameTest](https://www.w3.org/TR/1999/REC-xpath-19991116/#NT-NameTest) (as an abbreviated child axis test Step).

Nearly all of this change is tangentially related:

- 53c0a91 escaping `<` and `>` in XForm DSL attribute values, so the fix will actually show up in a passing `scenario` test (and 721c82c updating another test that would have passed otherwise). I believe this is a reasonable adjustment. Making the change in the engine would mean supporting malformed XML, and I think that would be a snake pit of complexity and potential source of bugs.

- 7c219a3 removing the "E2E" tests in `@getodk/tree-sitter-xpath`. These tests had some value early in the project for demonstrating A) how the grammar could be consumed downstream and B) a sense of how we might perform syntactic analysis on XPath expressions. Neither has been especially valuable for a while since, and the latter is especially moot after #166. Note that these test were not impacted in any way by the fix, this just felt like an opportune time to clean up some dead weight.

- 78d1a92 adding parse grammar tests specific to the change, starting with the expression causing an actual failure in `scenario`, a few variants on that, and then a test of every AxisName-as-NameTest

- **The actual substantive change achieving the fix:** 03305c1. Not much to note here, other than the fact that nothing downstream needs to change to benefit from this fix. It just ensures that expressions like `/child` are handled the same way as expressions like `/notchild`. (And 0afb4a2 is the `scenario` test newly passing with the change.)


sadiqkhoja (approved):
Some of the vocabulary used in the description of this PR and commit messages goes over my head, I probably need to go through the xpath spec. But I understand the problem this change-set fixes i.e. xpath parser was not correctly parsing expression that contained any axis name as step (node name).

One thing I noticed that we don't have any scenario that uses `child::anything` like expression, is it worth adding such a scenario 🤷‍♂️ ? I can't imagine anyone using child axis explicitly.

eyelidlessness:
> Some of the vocabulary used in the description of this PR and commit messages goes over my head, I probably need to go through the xpath spec.

Probably a good idea at some point! Even if only to get familiar with the grammar and the names used for various parts of syntax. I've found it really helpful to be familiar with these, both in implementing the `xpath` evaluator and in being able to describe particular kinds of expressions/sub-expressions with precision.

> But I understand the problem this change-set fixes i.e. xpath parser was not correctly parsing expression that contained any axis name as step (node name).

Right. And if you want to see the impact of the fix, you can compare the results of `yarn workspace @getodk/tree-sitter-xpath test` on 78d1a92 and after. What you'll see before is that every case where a Step references an element by name which could be an AxisName, you'll see a bunch of `(node_test)` where `(node_test (unprefixed_name))` should be.

> One thing I noticed that we don't have any scenario that uses `child::anything` like expression, is it worth adding such a scenario 🤷‍♂️ ? I can't imagine anyone using child axis explicitly.

That's sort of in the opposite direction of this PR, but a good question to raise nonetheless! You can see that there are tests for evaluation [in `xpath`](https://github.com/getodk/web-forms/blob/cfde5ea53085cdae22ffcfa274cb8a48dfc3d699/packages/xpath/test/native/axis.spec.ts#L275). If we were to introduce `scenario` tests exercising expressions using that explicit (non-abbreviated) syntax, I would expect it to break dependency tracking, causing any downstream computation update to fail. That's something that we could cover either in #166 or in a followup building on the abstractions introduced there.

PR #173, opened 2024-07-19
Design tweaks (#139)

### LINKED ISSUE #139: Design tweaks
Starting a list of small changes to improve the over all experience

**Spacing, weights, text fields**
- [x] Vertical spacing after text boxes (can we leverage the vue components here)
- [x] Make the spacing between selects consistent with the text fields (selects seems more spaced apart) 
- [x] Reduce text size/weight of title
- [x] Text field needs to be a lighter grey because it looks disabled (try #F7F8F9)
- [x] Hover state for text fields can be lighter 
- [x] For displayed values and conditionally disabled fields, it's hard to tell it's updating because the text is too light - Needs design (~~Idea: when it's dynamic maybe add a light blue highlight animation~~)

**Icons**
- [x] More consistent alignment of controls (icons in larger views, hamburger menu in smaller). Vertical centering makes those a moving target. 
- [x] more menu still appears pretty small 
- [x] Chevrons for groups are not aligned in the middle with the text

**Elevation**
- [x] [Remove boarder and elevation](https://www.figma.com/design/KFi9hIQdrRgo0rLqIC7Sg8/ODK-user-flows?node-id=5258-108702&t=7E9o9DdPAwWqHTRs-1) so it's flat
- [x] Flat buttons

**Repeats**
- [x] Add a "+" icon to the [add button for repeats with label](https://www.figma.com/design/KFi9hIQdrRgo0rLqIC7Sg8/ODK-user-flows?node-id=5023-34916&t=V2GBznX4TxQWpKlP-1). 
- [x] Adding the repeat label (really the label of the outer group) in the button text. That would be similar to Collect (see comment below)
- [x] The number should match the label text weight

sadiqkhoja:
Closes #139 

Should be merged after #164.

Last 17 commits are for this PR: [diff](https://github.com/getodk/web-forms/pull/173/files/62c785970283fa2c826e62664aca78e42b517f82..4da4eaa8788ca5072f751efb7b7ccf0463f51d00#diff-5b50bc3ac22db3a7f3c7b3c5cc8762b7be81ede134f1a35765a07070a1d13bb5)


packages/web-forms/src/components/RepeatRange.vue
```diff
@@ -15,7 +15,12 @@ const label = computed(() => props.node.currentState.label?.asString);
 		<RepeatInstance v-for="(instance, index) in node.currentState.children" :key="index" :instance="instance" :instance-index="index" @remove="node.removeInstances(index)" />
 
 		<div class="flex justify-content-start flex-wrap">
-			<Button label="Add" rounded outlined class="w-2 btn-add" @click="node.addInstances()" />
+			<Button rounded outlined class="btn-add" @click="node.addInstances()">
+				<span class="p-button-label" data-pc-section="label">
+					<span class="icon-add" />
+					<span class="btn-add-label">Add {{ label }}</span>
```
eyelidlessness, line 21: This would improve vertical alignment of the icon/text:

```suggestion
					<span class="flex btn-add-label">Add {{ label }}</span>
```

packages/web-forms/src/components/FormHeader.vue
```diff
@@ -124,22 +150,113 @@ const handleLanguageChange = (event: FormLanguage) => {
...
+				}
+				.print-button {
+					display: none;
+				}
+				.language-changer {
+					display: none;
+				}
+			}
+
+			@container formOptionsContainer (min-width: 240px) {
```
eyelidlessness, line 233: Hey look at that, it's a container query! Cool to see it in action. We may want to up this `min-width` (and the `.language-changer` `max-width`) some, but otherwise 🎉

packages/web-forms/src/components/FormPanel.vue
```diff
@@ -69,23 +69,27 @@ const toggleMenu = (event:  Event) => {
 </template>
 
 <style scoped lang="scss">
+.chevron {
+	padding-top: 2px;
+}
+
```
eyelidlessness, line 75: Quite a lot of what's going on in here feels like it might be less fussy and less full of magic tiny pixel spacing if the containers of affected elements were flexbox/vertically centered. I don't know if it'd cover everything changing here, but it did seem to have a similar effect in a couple places where I tried it.

packages/web-forms/src/components/FormQuestion.vue
```diff
@@ -13,11 +13,56 @@ const isSelectNode = (n: AnyLeafNode) : n is SelectNode => n.nodeType === 'selec
...
+
+.question-container {
+	display: flex;
+	flex-direction: column;
+	padding: 0.5rem 1rem 0 1rem;
+	scroll-margin-top: 60px;
+}
+
+:global(.p-panel-content .question-container) {
+	margin-left: -1rem;
```
eyelidlessness, line 41: What's happening here? It seems like it's accommodating spacing for the form-wide validation background, specific to questions in a group (and/or repeat?).

packages/web-forms/src/components/widgets/RadioButton.vue
```diff
@@ -45,8 +45,13 @@ const setSelect1Value = (item: SelectItem) => {
   outline: 1px solid var(--surface-300);
 	border-radius: 10px;
   padding: 15px;
-	margin: 20px 0;
+	margin: 0 0 20px 0;
   cursor: pointer;
+	background: var(--surface-0);
+
+	&:last-of-type{
+		margin-bottom: 0;
```
eyelidlessness, line 53: image

I think I mentioned in #164 concerns about using `margin` for spacing. This is a really good example of why `gap` might be a better choice.

(I think I recall there was a fix for this previously, but I saw it in #164 as well, and now I see it's moved here.)

eyelidlessness (changes requested):
A few mostly small things. I don't want to totally derail this into broader concerns, but I do want to re-emphasize concerns about margin for spacing, and `px` unit for quite a lot of things.

packages/web-forms/src/components/RepeatInstance.vue
```diff
@@ -52,7 +52,7 @@ const menuItems: MenuItem[] = [
 
 </script>
 <template>
-	<FormPanel :title="label" :menu-items="menuItems" class="repeat" :label-number="instanceIndex + 1">
+	<FormPanel :title="label" :menu-items="menuItems" class="repeat-instance" :label-number="instanceIndex + 1">
```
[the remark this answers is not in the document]
sadiqkhoja: > With scoped, the parent component's styles will not leak into child components. However, a child component's root node will be affected by both the parent's scoped CSS and the child's scoped CSS. This is by design so that the parent can style the child root element for layout purposes.

packages/web-forms/src/components/RepeatRange.vue
sadiqkhoja: It looks like this: 
image

packages/web-forms/src/components/FormQuestion.vue
sadiqkhoja: that's right.

packages/web-forms/src/components/RepeatRange.vue
eyelidlessness: Weird! I think I added it to the wrong line. Even then, I think something else changed that affects horizontal alignment. Anyway, as of the current commit ([8382ffa](https://github.com/getodk/web-forms/pull/173/commits/8382ffa83987656a4f0bf74af5fd4a5b0664d40a)), this has the same effect as what I meant to suggest:

```diff
-				<span class="p-button-label" data-pc-section="label">
+				<span class="flex justify-content-center p-button-label" data-pc-section="label">
					<span class="icon-add" />
					<span class="btn-add-label">Add {{ label }}</span>
```

packages/web-forms/src/components/FormQuestion.vue
eyelidlessness: Feels like this could use a comment. In general, I think spacing adjustments would probably benefit from some context/explanation when they use negative values, and values which are small or otherwise not obvious fractional proportions.

eyelidlessness:
This is now superseded by #164.

Issue #21, opened 2024-02-18
Show read-only input without a value without a control or space for a value


lognaturel:
As a form designer
I want to write some text without a value
So that I can provide information or guidance to form users

This is the `note` type in XLSForm. If there is no value, there should be no control or vertical space for a value.

Typically a `note` in XLSForm only has a label. But because XLSForm makes `note` an alias to type `string` with the `readonly` bind attribute set to `true()`, everything that's allowed for a `string` is allowed for a `note`. [Form showing what could be done with the XLSForm `note` type](https://docs.google.com/spreadsheets/d/12EQhUKACxbBs4OLs8pIkFkywFO0oNjmxw7S7ZIAWHQI/edit?gid=1068911091#gid=1068911091).

sadiqkhoja:
done in #181

PR #174, opened 2024-07-19
Engine support for notes

eyelidlessness:
Part of #21.

## Added engine/client node API: `NodeNode`

The main goal (and the bulk of substantive work) of this change is to distinguish "notes" from other inputs, in support of #21. The change introduces a distinct `NodeNode` (16cd6b3) client-facing type.

### What _is_ a `NoteNode`?

A `NoteNode` is modeled as:

1. Any leaf node
2. Associated with a `<bind readonly>` expression that is both:
    1. constant[^1]: referentially transparent, with no references to instance state, no implicit state dependencies
    2. known to produce a truthy value
3. Associated with an `<input>` containing at least one of:
    - `<label>`
    - `<hint>`

[^1]: I'd previously been phrasing this as "statically readonly", but I think there's some room for confusion in that terminology. "Constant" in this case is consistent with usage for the same concept in many programming contexts, for instance FP languages often have a notion of `constantly` as producing a thunk which always returns the same value.

This node type will be a subset of the nodes identified as `StringNode` before this change.

From a client perspective, a `nodeType: 'note'` should be sufficient to signal that the node is not expected to be user-interactive, and that it is expected to produce some user-facing text. There are a few distinctions between the `NoteNode` and `StringNode` types, apart from `nodeType`:

- `NoteNode.currentState.readonly` will always produce `true`
- A note-specific `NoteNode.currentState.noteText` property is used to convey the guarantee that at least one label or hint is present
- `NoteNode` has no write methods, as being constantly `readonly` is one of a note's defining characteristics

### What is a "constant truthy" `readonly` expression?

This is implemented in f2e00e7, which builds on some of the abstractions from #166. Neither aspect is exhaustive, just enough to satisfy the most obvious cases I could think of.

Currently, a constant expression is one where the complete expression consists of a syntax type:

- FunctionCall `true()`
- FunctionCall `false()`
- Number (number literal)
- Literal (string literal)

As of 4adc07b, this determination is also referenced in `createComputedExpression`, expanding on an existing optimization which already made a more naive version of the same check.

The check for whether a `readonly` expression is "constantly truthy" first checks that the expression is constant, and then narrows that to:

- FunctionCall `true()`
- Number which casts to boolean `true` (at runtime, not evaluated or using `xpath` casting logic, but JS runtime casting rules are the same)
- Literal (string literal) which is non-empty (this logic is redundant to the casting in `xpath`, but simple enough I think that's okay)

### What's not in `NoteNode`

We'd had some discussion about possibly accounting for more than a string data type in this work, but I decided to defer that for now. I think we'll be better off addressing the broader set of data types holistically, and I think that the expanded set of node types in this change will likely help inform that design.

We also discussed some consideration for determining the various ways a note's value state may or may not change. Nothing in this current pass addresses the note's value state other than ensuring it is produced consistently with other node types.

## Split engine/client API: `ModelValueNode` from `StringNode`

Since `NoteNode` is already being split off from `StringNode`, I took the opportunity (9baaa25) to address the awkwardness of `StringNode` representing both `<input>` and model-only nodes. This dual responsibility has made several client concerns more complex than necessary, and I've wanted to address it for some time. While it's not strictly necessary here, it felt like an appropriate context to situate the change. And it felt like it provides some good context on how the more pertinent `NodeNode` split looks from a client interface perspective.

As with `NoteNode`, a `ModelValueNode` has no write methods. Since it is not associated with a form control, user-facing scenarios will/should not typically provide a means to set model-only nodes' value state. (There could be future scenarios which would warrant writing model-only values, but those would likely be either internal or perhaps involve form design workflows).

### What's not in this `ModelValueNode` introduction

As with `NoteNode`, and with the same reasoning, I decided to defer consideration of non-string data types for `ModelValueNode`.

## Gradually reducing pertinence of `AnyNode.definition`

This section is mostly intended to make explicit that some of the above design is _also intended_ to support reducing client reliance on a given node's `definition`. This has been a long-standing goal, and these aspects help move us in that direction:

1. The `ModelValueNode`/`StringNode` split removes any need to distinguish whether a `StringNode` has a `definition.bodyElement`
2. The client-facing `NoteNode.definition` type is intentionally minimal. The actual internal definition is much more involved, with several distinguishing factors to support the "What _is_ a `NoteNode`?" characteristics described above. I believe there would be no value in exposing that aspect of the `definition` to clients... even negative value, by emphasizing information clients don't need to care about.

At any rate, I've catalogued the remaining client usage of the `definition` property of any node.

<details>
<summary>Remaining client `definition` references</summary>

- `web-forms`:
  - ~~check for truthy `node.definition.bodyElement`~~ (removed in 40af8ba)
  - form title (temporary, will go away when we introduce an _intentional API_ to access form title API)
  - some questionable heuristics about whether a `RepeatInstanceNode` has a single group child, partly branched on the group's conceptual type (checking for `'logical-group'` or `'presentation-group'`); all of this would be better served by dedicated engine/client API design
  - distinction between `<select>`/`<select1>` (likely temporary, would go away if/when we split those into distinct node types)
- `scenario`:
  - Checks for whether a repeat range allows clients to add or remove repeat instances (temporary, will go away when we support `jr:count` and `jr:noAddRemove`) 
- `ui-solid` (not high priority to support, but may be instructive to understand other potential client usage scenarios):
  - form title (same as `web-forms`)
  - distinction between `<select>`/`<select1>` (same as `web-forms`)
  - **debug only:** view serializing form state (likely moot when we introduce an API to access submission state)
  - **debug only:** view describing unsupported controls/body elements (we should consider a dedicated node type or some other dedicated interface for this!)
  - **debug only:** view highlighting when inputs have no label (this was really only useful in early dev, could go away whenever)
</details>

Unless there is something I haven't yet seen in other open PRs/work in flight, it seems pretty clear that nearly all usage is either going to become moot or serves only internal purposes. At this point, I'm **highly inclined** to include a `@deprecated` JSDoc tag on every node's `definition` property type, and would love to include that **in this change** if there's no objection.

## Internal-ish refactor: `ValueNodeDefinition` -> `LeafNodeDefinition`

In an effort to minimize drift between our semantic concepts and the names associated with them, I also took this opportunity to rename this aspect of the parse definition associated with leaf nodes. While it's _technically true_ that notes may have a value, their user-facing role is presentational and their value-bearing quality is both secondary and variable depending on the form design.

I say that this change is "interal-ish" because it is technically client-visible in each of:

- `StringNode.definition.type`
- `SelectNode.definition.type`

I did not find any client code referencing these properties, and as detailed above I'm inclined to discourage further client reliance on `definition` aspects of the API.

## Client updates

Since there are some breaking aspects of these changes, clients are updated to account for them. I've made a point to break these client-side accommodations into separate per-client commits for easy review.

This change **does** include a basic note implementation in `ui-solid`, which I used to spot check that it generally works as intended. It **does not** include any note implementation in the `web-forms` Vue client. I understand there may be more involved UI/UX considerations, so I'll leave that for a followup PR.

One thing worth calling out is that in 454c824 I was able to confirm that, in the `scenario` WHOVA smoke test where writes to `readonly` fields could fail (unless skipped in the test parameterization), each respective case is now known to correspond to a `NodeNode`.

## Testing

I haven't added any dedicated `scenario` integration tests for either `NoteNode` or `ModelValueNode`. I don't feel strongly either way about whether that would be valuable or superfluous.

I believe `ModelValueNode` is well exercised by the fact that its `nodeType` is now part of what determines whether a given engine node produces a corresponding JavaRosa-based `QuestionEvent`.

I feel comfortable enough with the WHOVA smoke test change as an exercise of `NoteNode`. It is clear automated validation that the functionality is working and has an impact on client integration. But I recognize that this is less discoverable than dedicated tests, so I'd be happy to add those if there's a preference for that.


packages/xforms-engine/test/parse/xpath/semantic-analysis.test.ts
```diff
@@ -109,4 +111,49 @@ describe('Semantic analysis', () => {
...
 		});
 	});
+
+	interface ExpressionPredicateCase {
+		readonly expression: string;
+		readonly expected: boolean;
+	}
+
+	describe('Constant expressions', () => {
+		it.each<ExpressionPredicateCase>([
```
sadiqkhoja, line 121: following cases fail:

```
{ expression: 'not(false())', expected: true },
{ expression: 'number("1.2")', expected: true },
{ expression: 'boolean(1)', expected: true },
```

Similarly `isConstantTruthyExpression` would fail as well when a boolean returning function is used with constant arguments.

To me this is low priority and can be handled in a followup PR, I don't think people would be doing this kind of stuff in their Form definitions. Just documenting the limitations of these two functions would be sufficient.

packages/xforms-engine/src/lib/assertions.ts
```diff
@@ -0,0 +1,15 @@
...
+	}
+};
+
+type AssertNonNull = <T>(value: T) => asserts value is NonNullable<T>;
+
+export const assertNonNull: AssertNonNull = (value) => {
+	if (value == null) {
+		throw new Error('Expected value');
+	}
+};
```
sadiqkhoja, line 15: Looks like this is not used anywhere

packages/xforms-engine/src/lib/reactivity/createAssertedMemo.ts
```diff
@@ -0,0 +1,18 @@
...
+	return scope.runTask(() => {
+		return createMemo((): U => {
+			const value = fn();
+
+			assert(value);
+
+			return value;
+		});
+	});
+};
```
sadiqkhoja, line 18: looks like unused code

sadiqkhoja (changes requested):

packages/xforms-engine/src/client/ModelValueNode.ts
```diff
@@ -0,0 +1,40 @@
...
+import type { BaseNode, BaseNodeState } from './BaseNode.ts';
+import type { GeneralParentNode } from './hierarchy.ts';
+import type { RootNode } from './RootNode.ts';
+import type { LeafNodeValidationState } from './validation.ts';
+
+export interface ModelValueNodeState extends BaseNodeState {
+	get label(): null;
+	get hint(): null;
+	get children(): null;
+	get valueOptions(): null;
```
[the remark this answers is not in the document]
eyelidlessness: I'm not sure! I think we might be in new territory having a concept of distinct model-only nodes. I did a quick spike to see if any `scenario` tests would fail, expecting _at least_ one or two failures on `relevant` checks, but I don't see any.

Keep in mind that there's already another (mostly[^1]) model-only node type: `SubtreeNode`. We don't presently make a special case for these state properties on `SubtreeNode.currentState`, and if we were to do that here, I'd expect to do it there too.

I think it's also worth reconsidering whether model-only nodes of any type are even pertinent to a client. More precisely, whether they're pertinent to **every client**. I did another quick spike to jog my memory, and there are definitely several `scenario` tests which fail if we filter them out[^2].

But I definitely think it's an open question whether they even provide any value to the `web-forms` Vue client. I think that question would have a similar answer for _some_ other presentational client use cases, depending on their use case: a hypothetical native app for form filling would line up with `web-forms`, but tooling for form design or debugging would almost certainly need a way to access model-only form aspects[^3].

Right now, my instinct on this one is to defer all of these questions, and follow up on them with a more thoughtful design around how we handle model-only aspects of form state at the engine/client boundary overall. I think it would be worthwhile to consider different "views" for different client use cases.

[^1]: I suspect there are edge cases where we designate a node a `SubtreeNode` even if it ultimately has descendants which aren't model-only. Probably worth filing an issue to confirm, and decide whether that's right.

[^2]: It's worth mentioning that this made me doubt that some of the affected tests are "really integration tests". I tend to think about model-only form features as basically "implementation details" of the form, at least while filling an instance. A stricter conception of "integration test" would probably only test their submission state.

[^3]: Tooling use cases would also almost certainly have more reason to access to `definition` aspects of a form. Now that I'm giving all of this more thought, that's something I want to keep in mind as I otherwise try to phase those out for our main user-facing client use case.

packages/xforms-engine/src/client/ModelValueNode.ts
```diff
@@ -0,0 +1,40 @@
...
+ *   the form's `<model>` as either an {@link Element} or {@link Attr})
+ */
+export interface ModelValueNode extends BaseNode {
+	readonly nodeType: 'model-value';
+	readonly appearances: null;
+	readonly definition: ModelValueDefinition;
+	readonly root: RootNode;
+	readonly parent: GeneralParentNode;
+	readonly currentState: ModelValueNodeState;
+	readonly validationState: LeafNodeValidationState;
```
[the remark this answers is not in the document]
eyelidlessness: As with the last question, my instinct is to defer this for now. While we'll definitely have different behavior on your example form, I would want to see how it affects real user-defined forms before tailoring the behavior to contrived scenarios.

I think it's worth emphasizing that this will not be new behavior introduced by this PR. The same nodes will have the same `validationState` (as well as `currentState.readonly`, etc) with this change as they do on `main` today. The main things that are changing about them now are their `nodeType` and the absence of write methods.

I think I'll ultimately agree with you on the expected behavior here, but I think it's probably better treated as a separate scope of work. And I think it would be good to think about it in the context of the [questions raised above](https://github.com/getodk/web-forms/pull/174#discussion_r1693358593).

packages/xforms-engine/test/parse/xpath/semantic-analysis.test.ts
eyelidlessness: Yeah. The intent isn't really to be exhaustive, mainly to bring the existing usage (which is even more incomplete, and much more naive) in line with the overlapping functionality this change needed.

I'd also say that this feels like functionality we'll want to push down the stack. The usage in `createComputedExpression` is an optimization that would be beneficial to any user of `@getodk/xpath`. There are already quite a few internal optimizations in that package where certain statically known things are evaluated at "compile" time (in the sense that XPath expressions are parsed into evaluable representations).

For the non-optimization `isConstantTruthyExpression` usage, I think some of the cases introduced in this PR are already a bit of a stretch from how real world forms will be defined (typical usage being `readonly="true()"`). While it's definitely more "correct" to handle the cases you posted here, I don't know that they'd make any difference either way on real forms.

packages/xforms-engine/src/lib/reactivity/createAssertedMemo.ts
eyelidlessness: Yeah, this and `assertions.ts` were part of a naive first pass, but didn't turn out to be needed. A little bit of how the sausage wasn't made 🙃.

As mentioned in the commit removing them, I think it might be worth keeping the original commit intact for slightly easier future lookup, because they're general enough we might find them useful for something else.

sadiqkhoja (approved):
I agree with the responses to my earlier comments, creating separate issues to track them is the best thing to do.

Let's get this PR in 🎉

PR #183, opened 2024-08-12
Vue Route in demo App

sadiqkhoja:
The existing mechanism of using History API to handle routing is not foolproof, I noticed it breaks when demo app is run from a subdirectory. Hence, I have install vue-router. This was also suggested in https://github.com/getodk/web-forms/pull/164#discussion_r1690270770


eyelidlessness (changes requested):
I'm excited to get this in. I do have several changes to suggest, on the `review/features/ui/vue-router` branch. See the [diff](https://github.com/sadiqkhoja/web-forms/compare/features/ui/vue-router...getodk:web-forms:review/features/ui/vue-router)), with each suggestion and reasoning in a separate commit. Feel free to cherry pick as you like, or make changes with adjustments as you see fit!

I know the separate branch/diff isn't necessarily an ideal format for review. I was finding it difficult to make some of the suggestions inline, and to make some it clear where some related changes would be implied across modules or with larger gaps between affected lines within the same module.

Hopefully none of these are particularly controversial. Looking forward to getting this in. I think it'll reduce a lot of little daily paper cuts that the manual history management approach has.

sadiqkhoja:
Thanks for the review, I have merged all the changes you suggested.

eyelidlessness (approved):
🚀

PR #184, opened 2024-08-12
Build demo app for distribution / preview

sadiqkhoja:
This enables us to deploy demo app anywhere as a static web-site. For example: if you run `npx live-server packages/web-forms/dist-demo ` after building the project, you will see the demo App.


eyelidlessness (approved):
This is good as-is, so you can feel free to merge if you prefer. But I have a branch with some related _additive_ changes that I think would be good to bring in if you'd like to incorporate them: [diff](https://github.com/sadiqkhoja/web-forms/compare/features/ui/build-demo-app...getodk:web-forms:review/features/ui/build-demo-app).

As with #183, I broke this down by commit-with-reasoning. The general goal of the changes are:

- Reduce differences between the `web-forms` Vite/Vitest configs and other packages (also fixes some confounding config type issues in the process)
- Eliminate the internal build (for E2E testing, originally introduced in #152), since this new build now satisfies that need and it redundant

I did also make some slight changes to the branchy logic in the config, I think it makes the intent clearer. But the resulting config is unchanged.

PR #204, opened 2024-08-28
xpath: update README to reflect `indexed-repeat` support

eyelidlessness:
This should have been done in #195 but I forgot!

I could also update `feature-matrix.json` here, but I was thinking maybe we'll be doing that in release prep?


eyelidlessness:
Closing in favor of #208

Issue #205, opened 2024-08-28
Form design pattern fails: Repeating as long as a condition is met


eyelidlessness:
The form design pattern is described in the XLSForm Reference section titled _[Repeating as long as a condition is met](https://docs.getodk.org/form-logic/#repeating-as-long-as-a-condition-is-met)_. This failure can be seen in [CI failures](https://github.com/getodk/web-forms/actions/runs/10600400699) since merging both #187 and #195.

Failure presents as a long delay while computations rerun (blocking the main thread), and finally a reactive infinite loop error thrown by Solid. I've confirmed that the failure is not specific to `indexed-repeat`, and the same behavior occurs with the child-vaccination form's `indexed-repeat` expression rewritten to its LocationPath equivalent.

Since this is affecting CI now, we will:

- [ ] Temporarily skip failing child-vaccination smoke test
- [ ] Fix affected functionality

PR #206, opened 2024-08-28
Temporarily skip child-vaccination smoke test

eyelidlessness:
Part of #205.

I really don't want to risk forgetting that this test is skipped. So I've conditioned the skip to be within a two week window, which we can easily extend as needed. I also included a non-code escape hatch to quickly unblock CI if rerun.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ defaults:
 
 env:
   TZ: 'America/Phoenix'
+  GITHUB_ACTIONS_RUN_ATTEMPT: ${{ github.run_attempt }}
 
 jobs:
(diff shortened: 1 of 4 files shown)


eyelidlessness:
After opening this I realized the form definition itself is the culprit. I'll update the issue with a more detailed explanation for posterity. For this PR, I think it's best to close. I'll have a replacement PR soon which takes the tack we discussed in Slack yesterday: creating an alternate form definition/test with a change to avoid the pathological case triggered by the form design as-is. I expect that replacement PR will also involve updating the child vaccination smoketest to account for further progress, since it was `indexed-repeat` and then `jr:count` blocking its progress previously.

Issue #203, opened 2024-08-27
External secondary instances: Engine representation, XPath support


eyelidlessness:
This design issue is part of broader support for external secondary instances:

- [I/O support for `jr:` URLs][io-design]
- [Error conditions][error-conditions]
- **[Engine representation, XPath support][engine-xpath]**

[io-design]: https://github.com/getodk/web-forms/issues/201
[error-conditions]: https://github.com/getodk/web-forms/issues/202
[engine-xpath]: https://github.com/getodk/web-forms/issues/203

This issue is **mostly** focused on engine internals—including aspects of both `@getodk/xforms-engine` itself, and its integration with `@getodk/xpath`.

We **may** consider minor, optional changes to the engine/client API as a part of one of the approaches described here. While optional, there may be some compelling reasons to consider them anyway (possibly overlapping with design around [error conditions][error-conditions]).

## New concepts

Supporting external secondary instances introduces two concepts that don't currently exist in the Engine/XPath portion of the Web Forms stack. Rather, two facets of the same concept:

1. Data outside of a form definition
2. Data formats beyond XML

The concept unifying both is that the data can be _referenced by a form's XPath computations_, and is treated by those computations **as if** it...

1. ... were part of the form defintion
2. ... had the same XML semantics as the form definition

Once we've addressed the I/O and error condition aspects of this functionality, the main requirement of supporting external secondary instances is to satisfy those facets.

## Option 0: negate the external and non-XML facets of the problem

This is, more or less, the Enketo option:

1. Parses and processes supported non-XML formats into an XML/DOM representation suitable for expected XPath evaluation

2. Inserts that representation into the form as if it were an [internal secondary instance](https://getodk.github.io/xforms-spec/#secondary-instances---internal)

This is a cromulent, naive, solution! However, I have some concerns with it.

First and foremost, it is likely to **exacerbate the worst performance challenges we currently face**. Second, it would squander an excellent opportunity to begin addressing that problem with a solution frequently at the center of many other goals.

## Option 1: evolve `@getodk/xpath` to support these concepts

This option is... to decouple `@getodk/xpath` from the browser/XML DOM (#39)!

I've prototyped a solution to this that I think is actionable, suitable for the feature, and likely to deliver on other benefits I'd hoped for.

[...]

Issue #39, opened 2024-02-20
Decouple @odk/xpath API from XML/browser DOM


eyelidlessness:
I'd like to explore the possibility of decoupling the @odk/xpath API from the XML/browser DOM API. Specifically for the purposes of this issue, the decoupling would involve accepting a different type for `contextNode` with the following considerations:

- Provide a significantly more **limited** interface, restricted to only those semantics of a node which satisfy the implemented XPath semantics. This interface would likely still be "a DOM", in the sense of representing a traversable tree of nodes consistent with the XPath node types, but it wouldn't necessarily be "the DOM".

    A simpler interface would allow us to more easily reason about potentially substituting node implementations. And being able to more easily reason about this, we could better evaluate if and how to reap other potential benefits (discussed below).

    It's an open question whether this simpler interface should attempt to maximize **compatibility** with "the DOM" (i.e. so that an unmodified XML/browser DOM reference is **assignable**), but my gut instinct is that it should not be a requirement though it may be nice to have. If we pursue this, and choose not to maximize compatibility, we'd still have the option of providing a compatibility layer e.g. wrapping the underlying XML/browser DOM APIs.

- Minimize incidental complexities of different interfaces for different node types. This is a source of significant incidental complexity in the evaluation implementation as well, and it's highly error prone. A trivial example of this kind of incidental complexity is the lack of consistent, direct access to certain properties, such as `ownerDocument`. Another good example is the combinatory explosion if slightly different ways to access node data (`textContent`, `innerText`, `children`, `childElements`, `data`, `value`, ...).

- Require (whether enforced or not) certain guarantees which aren't assumed in the XML/browser DOM compatible API we currently provide. Top of mind here:

    - Require that the entire context is read-only for the entire scope of a given evaluation. This would unlock a wide variety of potential performance improvements without baking in a lot of domain-specific assumptions.

    - Require that evaluation is always performed against a context node which is fully intact/connected to its root. This would overlap both with API simplification and performance.

- Provide specific affordances for additional optimizations, e.g. the ability to designate certain aspects of a context tree permanently immutable.

- Make callable aspects of the API explicit, either by choosing to express them as methods rather than as property accessors, or by designating them as getters with a clear indication that they are anticipated to have an associated runtime behavior. In the most minimal version of this, it would help us reason about where certain accesses are more expensive than others.

[...]

PR #207, opened 2024-08-30
Fix: evaluate `jr:count` in context of affected repeat; fix/update child vaccination smoketest, other Scenario tests

eyelidlessness:
Addresses issue discussed in #205, updates other related "indefinite repeat" tests.

I'm not sure whether we should keep #205 open or not. I'll note that I did do the brief timeboxed investigation described near the end of that issue. The fix I had in mind did not work as described. It's possible there are adjustments to the approach which could make it work, but I think it's best to prioritize separately so we can get CI passing and move forward.

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Not applicable

I doubt it's applicable, but I'll gladly check if there's any doubt.

The vast majority of the change is testing.

Regarding the `jr:count` evaluation context fix: it is definitely not the best possible solution! It is a solution consistent with how we've handled other presentation of the same problem. The better solution will come with #203 if we go with the option to address #39.

Regarding the rest, we knew we needed to update the child vaccination smoketest. Skipping the pathological case seems like the right call, and adding an alternate without it helps demonstrate our progress on supporting that form. I was tempted to include a fix for the new known point of failure in this change as well, but decided to hold off (for reasons discussed in commit notes on c4b8168).

No regression risks I'm aware of.

Only the ones under test.

## What's changed

This is sort of a few things glommed together because they're conceptually related:

1. Skips the child vaccination smoketest, as directly ported from JavaRosa. As described in #205, the form includes a `jr:count` which causes an infinite loop, where each newly added repeat instance triggers another new one to be added.
2. Introduces an alternate version of that child vaccination fixture, without the pathological behavior.
3. Updates the child vaccination smoketest to reach the current point of failure before fixing `jr:count` evaluation context.
4. Fixes `jr:count` evaluation context.
5. Updates the child vaccination smoketest further, to account for the next (current as of this PR) known failure.
6. Fixes a mistake in #187's updates to `scenario`, where controlled repeats (either `jr:count` or `jr:noAddRemove`) produce a `PROMPT_NEW_REPEAT` positional event. They should not, and this was already known, documented in porting notes on other ported JavaRosa tests.
7. Updates the first of those other tests, which passes immediately with (6).
8. Updates the last of those other tests, which passes when parameterized with a minor correction of the `jr:count` expression.

I think this list will serve as a good summary for a merge commit. For review, I'd recommend taking some time to read through each commit's notes to get more detailed context.

### DIFF
--- a/.changeset/quiet-donkeys-tie.md
+++ b/.changeset/quiet-donkeys-tie.md
@@ -1,0 +1,8 @@
+---
+'@getodk/xforms-engine': patch
+'@getodk/web-forms': patch
+'@getodk/scenario': patch
+'@getodk/ui-solid': patch
+---
+
+Fix: evaluate `jr:count` in context of affected repeat
--- a/packages/scenario/resources/smoketests/child_vaccination_VOL_tool_v12-alt.xml
+++ b/packages/scenario/resources/smoketests/child_vaccination_VOL_tool_v12-alt.xml
@@ -1,0 +1,1201 @@
+<?xml version="1.0"?>
+<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa">
+    <h:head>
+        <h:title>child_vaccination_VOL_tool_v12</h:title>
+        <model>
+            <instance>
+                <data id="VOL_CVT_0627" version="1">
+                    <today/>
+                    <start/>
+                    <end/>
+                    <deviceid/>
+                    <phonenumber/>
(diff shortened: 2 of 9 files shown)


sadiqkhoja (approved):
I tried a simple [Form](https://github.com/user-attachments/files/16821749/PlantCount_xml.txt) with relative path in `jr:count` function to confirm the fix made in https://github.com/getodk/web-forms/pull/207/commits/f6738ecbf440a281d6001459039cc366421458be and it works.

Other changes in tests also make sense to me.

PR #159, opened 2024-07-10
Prepare release

eyelidlessness:
I did a little fiddling with changesets:

- f18d1d9: retcon the changeset for engine `appearances` support, overriding `patch` -> `minor` where appropriate; break out separate changeset for `@getodk/common`
- 4d09e42: manually override commit reference to that retcon

I hope and expect this sort of fiddling will become less frequent as I/we get more of a rhythm using the changesets tool.

PR #208, opened 2024-09-04
Prepare release: `@getodk/web-forms` 0.3.0

eyelidlessness:
Supersedes #204, which I'll close after opening this.

A few quick notes:

- I'm not sure how best to title release prep PRs. We're bumping each package's version based on its own semver changes, as determined by that package's outstanding changesets. I wanted to look back at previous release prep PRs and wasn't thrilled by the less specific #159; the more specific formats of #116 and #125 don't make sense when we're releasing different versions per package. I think since `web-forms` is generally the package most end-users will use, this PR's title format might be a good one to follow in the future. But open to suggestions!

- As in #204, I updated the `xpath` README to reflect `indexed-repeat` support. That README section may be better removed, but I want to hold off for the work designed in #203. We may want to move XForms functionality out of `xpath` altogether, and into `xforms-engine`.

- I _think_ the feature matrix is up to date. Another pair of eyes to scan it for anything I might have missed would be welcome.

- I also added a top-level script to only format the top-level README, and added its invocation to the end of the feature matrix script. It seems kinda goofy to automate part of that process and then leave a little bit of manual followup work to do after the automation is run.

- I want to move `feature-matrix.json` into the directory with the script that controls it, but held off to ask about it here. Is the intent of having it top-level so that hypothetical contributors outside the ODK team have clear direction for updating it? If so, I think that's a pretty unlikely contribution scenario, but we could move it and then add a symlink at the root.


sadiqkhoja (approved):
LGTM 🚀

PR #210, opened 2024-09-09
Apply Prettier in .vue `<script>`/`<style>` blocks, for more consistent code style formatting throughout the project

eyelidlessness:

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [x] Not applicable

At least I don't think it should be applicable. If it is, we have bigger problems!

Spot check the automated formatting applied. Nothing looks like it could possibly change behavior.

Ideally we wouldn't have this special casing at all, but this is the nature of layers upon layers of custom language tooling with partially cyclic inter-dependency 🙃 

This shouldn't have any regression risk.

N/A.

## What's changed

We had previously excluded .vue modules from prettier formatting, due to conflicts between formatting applied in `<template>` blocks, by prettier and eslint-plugin-vue. This change avoids that issue by:

1. Applying formatting to .vue modules _through ESLint_…
2. … with the `prettier-vue` plugin…
3. … configured specifically to exclude `<template>` blocks
4. Disabling no-undef in .vue, consistent with the rest of the project

With these configuration changes, the current set of automated formatting changes can be seen in 1c7f219.

- - -

I previously mentioned the possibility of also making some subjective changes. I think I'll hold off on that and defer to @sadiqkhoja on when/whether to make specific syntax choices in .vue modules, in order to get the most (or least worst) out of Prettier's opinions.


sadiqkhoja (approved):
Thank you! This makes the code beautiful 🎉

PR #213, opened 2024-09-12
Chore (xforms-engine): finish reorganization of parse-related file system structure

eyelidlessness:

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [x] Not applicable

I haven't done any manual verification at all, apart from running `yarn workspace @getodk/xforms-engine test:types` a zillion times as I made incremental changes. If anything is broken that wasn't caught by that, we have much bigger problems. And even if so, there's a lot of downstream CI to alert us if that's the case.

**Edit** (before I even had a chance to open the PR!): turns out downstream CI did alert us to an issue. More on that in the change description below.

I relied **heavily** on VSCode to do nearly all of the work.

We agreed that this change was desirable some time back. I finally decided to finish the task, because it was beginning to irritate me introducing more new parsing functionality with this partially done.

I don't believe anything in this change is a regression risk which won't also be caught by CI.

N/A

## What's changed

This completes a task, begun in beb4d38 (#154), to reorganize engine parsing logic together in a coherent way. In more detail:

- As a first step, all remaining parsing logic not already under `src/parse` was moved there in its current file system structure, with no source code changes apart from updating import paths.
- Some subjective changes were made to rename the various `DependentExpression` subclasses to reflect their expression-ness, and to restructure those together. This was done in several steps, as I wanted to feel out the implications without getting too deep into such a broad change. 
- Likewise, the `DependentExpression` class was made explicitly `abstract`. I expected to do this when the concept was introduced, but I deferred commiting to it until we had more usage in place to determine whether it made sense to proliferate subclasses for every usage. (It does!)

All of the above were done with heavy reliance on VSCode and the TypeScript language server to make sure the appropriate imports and name references found their way to all affected call sites. Finally, one issue did arise in CI that didn't show up in type checks:

- A cyclic module dependency between `BodyDefinition` and group/repeat element definitions resulted in some runtime errors parsing repeats. Importantly, **this change did not introduce the problem**, it only caused it to finally be surfaced. I don't entirely know why that is, but we would likely have encountered issues with it eventually anyway. I went ahead and picked a narrow fix: assigning a `body` reference to each body element definition, updating these recursive child constructor calls to reference the same logic on a `BodyDefinition` instance method rather than the static method which precedes it. A more thoughtful fix would likely involve something more like the child construction logic in `src/instance/children.ts`, but that felt like more of a scope stretch.

- - -

I would recommend reviewing this commit by commit, I think that'll at least make it easiest to see which changes were most/least assisted by TS/VSCode. And I think the commit notes will give some insight into why some of the steps were taken as they were (which will hopefully be helpful for future git history quests as well).


eyelidlessness:
I should mention: this shouldn't result in any _meaningful_ for clients. It does rename some internals which are observable _at runtime_ in clients, so it could be construed as a "breaking change". But I don't believe any of the affected types are exposed as directly client-importable. And anything transitive a client could access (by way of a node's `definition`) has long been discouraged.

If we were post-1.0.0 I'd want to discuss the semver implications of this. I don't think it's worth even a changeset in our current release posture, but I do think it's worth mentioning so 
that in the future we'll have some precedent of discussing the impact of observable but not-intentionally-contractual changes of this sort.

sadiqkhoja (approved):
Went through this commit by commit, don't see anything problematic. Happy to see all parsing code under parse directory.

Also manually tried few Forms, all are working!

PR #215, opened 2024-09-13
Stub unsupported control nodes: `range`, `odk:rank`, `trigger`, `upload`

### LINKED ISSUE #214: `<trigger>` (currently unsupported) incorrectly treated as `ModelValueNode`
Filing a quick issue so we have something to track. This is definitely a bug!

In chat we're discussing filling in gaps in unsupported question types, by parsing them to stub nodes with `unknown` for anything pending further design. I know there are a few other cases, and expect they'll have the same behavior.

Will fill in more details if/as they come up, otherwise the intent of this issue is to track:

1. **Do not** produce `ModelValueNode` for form nodes associated with known body elements
2. **Do** produce something for those nodes, which for now will just be a node type which clients can't do anything with (other than observe their presence)

eyelidlessness:
Closes #214.

- [x] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [x] Not applicable

I would be shocked if there are any cross browser implications to this change.

The vast majority of it was guided by:

- Type checking
- Existing tests

**Note:** I have not added any _new tests_. I am certainly open to it. If there's interest, the tests I think would be most valuable would be engine-level unit-ish tests verifying that all of the stubbed node types:

- exist
- have the expected `nodeType` (i.e. not `'model-value'`, reported as a bug in #214)

I think that testing beyond that would ultimately have negative value: anything else we would be inclined to test will definitely change when we implement real support for these node types.

I briefly considered actually implementing engine support for `<trigger>`, since that lack of support was also what prompted this work. But I think this is a good unit of work, and I believe it would be trivial to follow up with that additional (now even more trivial) change.

Any form with any of the following controls should now present those nodes as unsupported in the `web-forms` UI:

- `<range>`
- `<odk:rank>`
- `<trigger>`
- `<upload>`

## What's changed

Anything I'd add here would be redundant to notes on each commit.

### DIFF
--- a/.changeset/nice-bags-sneeze.md
+++ b/.changeset/nice-bags-sneeze.md
@@ -1,0 +1,8 @@
+---
+"@getodk/scenario": patch
+"@getodk/web-forms": patch
+"@getodk/xforms-engine": patch
+---
+
+- Fix: unsupported controls treated as `ModelValueNode`
+- Stub unsupported control nodes: `range`, `odk:rank`, `trigger`, `upload`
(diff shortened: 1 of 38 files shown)


sadiqkhoja (approved):
LGTM

PR #219, opened 2024-09-18
Consolidate XForms fixtures

eyelidlessness:

- [x] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [x] Not applicable

There is some branchy logic for browser/Node differences. Otherwise I think if there were any potential for browser-specific issues we'd probably see test failures.

Ran tests and dev modes for each downstream package to validate.

I think the separation between special directories (`test-javarosa`, `test-scenario`) and regular category directories (all the others) could be more clear! But I was already getting wary of this becoming an over-engineered yak shave, and feel this is probably good enough to at least have clear iteration points if we need better guardrails.

The only users it should affect are us (tests, dev mode). CI says tests are unaffected, I've manually verified both pertinent dev modes.

They're all here, that's what this PR is about 🙃

## What's changed

- Moved all fixtures into `@getodk/common`
- Introduced a generalized `XFormFixture` interface for metadata about a given fixture, and providing an (asynchronous) method to load the fixture's XML content
- Introduced a few shared points to access the project's fixtures
- Migrated all usage (`@getodk/scenario`, `@getodk/web-forms`, `@getodk/ui-solid`) to these access points and interfaces

As mentioned, the `XFormFixture` design intentionally makes fixture loading asynchronous. This addresses a couple of issues that have been lurking and/or growing:

1. As we add more (and especially larger) fixtures, the UI client demos suffer degraded load performance. This is frustrating during dev, and tends to give a misleading picture of actual form load time.
2. Synchronous loading (`import.meta.glob` with `eager: true`) is a constant risk of mistakenly bundling fixtures in published packages.

The same design is backed by an abstract class meant to provide roughly the same interface for loading remote XForms, e.g. in the upcoming preview functionality. Ideally this will be trivial to consume for that purpose, and allow isolation of user-facing code from preview IO/XLSForm integration.

Beyond that, there are a few lint/formatting config updates to address changes to the filesystem structure, and a few lint/formatting fixes to address modules which were previously ignored by mistake.


eyelidlessness:
I don't _think_ this warrants a changeset in its current state. I think we'd probably want to consider it post-1.0, but for now it should be entirely internal.

Issue #217, opened 2024-09-16
Make version information accessible to make troubleshooting easier


lognaturel:
As a user (system administrator, data collector, etc)
  I want to include version information when reporting an issue
  So that the development team can fix it

ODK tools typically provide version information from `git describe --tags --dirty --always`. This information is available from different places:
- At the bottom of the landing screen in Collect. This is so data collectors can quickly verify that they're on the expected version while training, etc
- From the help menu in Central
- From a system summary screen and at the bottom of the offline menu in Enketo

Some ideas for Web Forms:
- tooltip when hovering over "powered by"
- bottom of the screen

alyblenkin:
I think bottom of the page works well! We could display it under the "powered by ODK" label so the user can see it if they are  desktop or on their phone. [Figma link](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/Web-Forms?node-id=234-5069&t=FX1DIthW728DuhfP-1)

PR #223, opened 2024-09-23
Web Forms Preview service

### LINKED ISSUE #151: Public service where anyone can try their XLSForm with Web Forms
User story
---
As someone interested in web forms for the ODK ecosystem
I want to try my XLSForm with ODK Web Forms
So that I can provide feedback and determine whether it's ready for my use

We had initially discussed adding this as a button on https://getodk.org/xlsform/ but we've decided we'd rather have a dedicated page with more information. XLSForm online may link to that page.

Requirements:
- upload an XLSForm and preview it
- list of demo forms to preview 
- clear way to give feedback
- mark an uploaded form as ok for the ODK team to look at and use for development purposes

On the ODK website, I think it should go under resources as "Try new ODK Web Forms" or "Web Forms"

[Figma](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/Web-Forms?node-id=120-1616&t=nhahbntZEAL3n873-1)

Image

## User flow:
-  User click "upload file" link or drops an XLSForm file
- System shows loading spinner while file is being uploaded and converted
- If file has errors/warnings those are listed below the upload section otherwise system shows "Preview Form" and "Upload new Form" buttons
- On pressing "Preview Form" button, Web Form is previewed in a new tab
- If user clicks "Upload new Form" then system shows the initial state, user can upload another file and see its preview.
- System should allow user to preview multiple Form at the same time in different tabs

Future considerations 
----
In the designs we have some of the future state considerations like:  
- Couple of paragraphs about ODK Web Forms
- The status and timeline
- Ways to find out more

sadiqkhoja:
Closes #151 and #217

## What's changed

- The "Demo" app inside `@getodk/web-forms` that was used for preview Forms during development has been enhanced to include:
  - Ability to upload any XLSForm or XForm for preview
  - List of real world Forms so that users who don't have any Form handy can preview the progress of ODK Web Forms
- Added a Feedback button in the `FormPreview` component of demo app. 
- Updated GitHub workflow to upload built artifact of demo app.
- Added version number in the footer of `OdkWebForms`.

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Not applicable

- Manually tested the functionality by uploading XLSForm and XForm.

- Also tested Feedback button by providing test feedback. 

- Also deployed to nginx as static website to ensure everything works: http://odk.sadiqkhoja.com/demo/xls-preview/#/

If it is valuable I can add some component/unit tests in a followup PR.

**Conversion of XLSForm**

For conversion of XLSForm into XForm (xml), I have integrated with `xlsform-online` service, which can convert and validate XLSForms, it also has ability to persist the Form so that we can analyze and use after anonymizing them.

Other approach I tried was running `pyxform` directly in the browser using [pyodide](https://pyodide.org/en/stable/), but it had following drawbacks:

- It added ~20 MB of additional resources
- It didn't included validation functionality because (pyxform depends on JRE for validation)
- This solution would have still required some place to persist user Forms

**Configuration of base URL of XLSForm-online service**

As discussed in our meeting, I have added a `config.json` file that is fetch on the page load. Right now it only contains one field i.e. `xlsform-online-url`, more configuration can be added in the same file later on if we need them. The `config.json` file is copied using `viteStaticCopy` plugin, because I wanted this file to be at the root directory of the build output so that it could be easily modified while deploying. I couldn't figure out how to achieve that using vite's default behaviour. I tried `new URL(url, import.meta.url)` but the data of config file was bundled inline.

Before implementing `config.json` approach, I was using environment variable. That approach required xlsform-online-url to be set during build time, which makes the value part of the output bundle and hard to modify during deployment.

**Icons**

I have updated `icomoon` font and its configuration to include required icons. I have started to remove `icomoon` and use [`@kalimahapps/vue-icons`](https://github.com/kalimahapps/vue-icons), which I will submit in a separate PR.

**GitHub Artifact**

I am uploading `dist-demo` (build output of demo app) as workflow artifact instead of Release artifact to avoid creating confusion for the developers who wants to use ODK Web Form in their own applications. This artifact can easily be downloaded from 'Actions > run > Artifacts section' of GitHub webpage or using [GitHub API](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-repository)

**Version number**

I am retrieving version of the web-forms using the command suggested in the issue #217 with slight modification and set the output to vite `define` as suggested in one of our meetings.

**Preview functionality in the demo app**

The motivation to enhance existing "demo" app to include functionality required for preview service is to use the same product during development that we ship for enduser. Also to keep the end product closer to our development workflow. But currently it is hard to see that `@getodk/web-forms` package has two different responsibilities/build output 1) produce Vue component to be consumed by a Vue host application; 2) produce a static website that consumes Vue component to show ODK Web Forms. I think it is better to extract out demo app from the `@getodk/web-forms` package into a separate package/directory and adjust tooling such that changes in `@getodk/web-forms` are watched so that development experience is not negatively affected.

**XLS Fixtures**

I have added XLS files for "real world"/demo Forms in the common package and used the similar pattern established for consuming XForm (xml) to get and serve XLS Fixtures by adding a separate module `xlsforms.ts` in `@getodk/common`. These files can easily be serve from xlsform-online service but I found it nice to keep these files in this repository so that we can use/update them for development/test purposes.

**Extra fixtures in build output**

The way `xform.ts` in `@getokd/common/fixtures` is setup, all the xml files are included in the build output of demo app (`dist-demo`) even though they are not used for production. I see this as just a small nuisance because end users won't be impacted as these files are not requested during runtime.

This let users to use their own Forms to see the workings of new Web Forms, and to see if currently implemented features are sufficient for them to adopt it. 

I have just added version number in the footer of OdkWebForm component, which is entry/main component of `@getodk/web-forms` package. Other than that nothing is changed in the components of the package that might cause regression.

Any XLSForm or XForm can be used.


packages/web-forms/env.d.ts
```diff
@@ -3,3 +3,6 @@
 /// <reference path="../../vendor-types/primevue/events.d.ts" />
 /// <reference path="../../vendor-types/primevue/dropdown.d.ts" />
 /// <reference path="../../vendor-types/primevue/multiselect.d.ts" />
+/// <reference path="../../vendor-types/@vue/input-html-attributes.d.ts" />
+
+declare const __WEB_FORMS_VERSION__: string;
```
eyelidlessness, line 8: Fine for now, but for future reference, I have found it better to do something like:

```ts
declare global {
	// eslint-disable-next-line no-var
	declare var __WEB_FORMS_VERSION__: string | undefined;
}
```

... and then:

1. Reference the value as a property on `globalThis`
2. Do an explicit non-nullish check

All other variables using this pattern are using a regular `CONST_CASE` naming convention (without prefix/suffix). Let's not change this now, I think it'll be good to consider whether there's a better naming convention across the board.

packages/web-forms/src/demo/FormPreview.vue
```diff
@@ -11,11 +12,17 @@ const formParam = route.params.form as string;
 
 const formXML = ref<string>();
 
-const demoFixture = xformFixturesByCategory.get(categoryParam)?.find((fixture) => {
-	return fixture.identifier === formParam;
-});
+let xformResource;
```
eyelidlessness, line 15:
```suggestion
let xformResource: XFormResource<'local'> | XFormResource<'remote'> | undefined;
```

packages/web-forms/src/demo/FormUpload.vue
```diff
@@ -0,0 +1,334 @@
...
+};
+
+const toggleHighlight = () => {
+	// short-circuit if a file is being uploaded
+	if (formUploading.value) {
+		return;
+	}
+
+	highlighted.value = !highlighted.value;
+};
```
eyelidlessness, line 106: Nothing for now, but since this is the handler for both `dragenter` and `dragleave`, there are a zillion ways this can produce surprising states. It'll make sense to follow up with:

- distinct enter/leave handlers which set clear states
- (likely depending on need, to be discovered in manual testing) potentially handling for viewport blur, other conditions where `dragenter` state could get "stuck"

packages/web-forms/src/demo/FormUpload.vue
```diff
@@ -0,0 +1,334 @@
...
+
+const reset = () => {
+	uploadedFilename.value = '';
+	error.value = '';
+	warnings.value = [];
+	xformUrl.value = null;
+};
+
+// Optional: Prevent file drop on rest of the page
```
eyelidlessness, line 115:
```suggestion
// Prevent file drop on rest of the page
```

🙃

packages/web-forms/src/demo/HomePage.vue
```diff
@@ -0,0 +1,279 @@
...
+		<section v-if="isDev" class="dev-forms">
+			<FormList />
+		</section>
+	</div>
+</template>
+
+<style scoped lang="scss">
+@import 'primeflex/core/_variables.scss';
+.home-page-component {
+	--font-family: 'Hanken Grotesk', sans-serif;
```
eyelidlessness, line 133:
```suggestion
	--font-family: 'Hanken Grotesk', Roboto, sans-serif;
```

packages/web-forms/src/demo/composables/xlsform-online.ts
```diff
@@ -0,0 +1,44 @@
...
+		formData.append('file', file);
+
+		const response = await fetch(`${baseUrl.value}/api/xlsform`, {
+			method: 'POST',
+			body: formData,
+		})
+			.then((res) => res.json())
+			.then((json: XlsOnlineResponse) => json)
+			.catch((fetchError: string) => {
+				error = fetchError;
```
eyelidlessness, line 32: For future iteration: there are whole classes of errors where we just spit out `FetchError`, and we could message more clearly what went wrong based on the actual kind of error. Top of mind (not exhaustive) from local testing:

- Failure to reach server at all
- CORS failure
- Empty response (saw this while trying with `pyxform-http`, before I realized I needed to use `xlsform-online`)
- 404 (and maybe other common HTTP errors, or known errors if we can enumerate them)

packages/web-forms/src/demo/demo.ts
```diff
@@ -4,6 +4,9 @@ import { createApp } from 'vue';
 import { webFormsPlugin } from '../WebFormsPlugin';
 import OdkWebFormDemo from './OdkWebFormDemo.vue';
 
+import hankenGrotesk400 from '@fontsource/hanken-grotesk/400.css?inline';
+import hankenGrotesk600 from '@fontsource/hanken-grotesk/600.css?inline';
+import hankenGrotesk700 from '@fontsource/hanken-grotesk/700.css?inline';
```
eyelidlessness, line 9: Should these also be configured in `vite.config.ts`? For Roboto, the config comment says:

> Load only a subset of the font family.

I can't remember if that's copypasta from something I found online, or if I made that note myself based on the effect of the config. If it does make a difference to build weight, I think it'd be worth it.

packages/web-forms/vite.config.ts
```diff
@@ -50,20 +57,36 @@ if (webkitFlakinessMitigations) {
 	globalSetup.push('./tests/globalSetup/mitigate-webkit-flakiness.ts');
 }
 
+const copyConfigFile = viteStaticCopy({
+	targets: [
+		{
+			src: 'src/demo/config.json',
```
eyelidlessness, line 63: Why not just have this in assets and `import` it like other static files (e.g. `import demoConfig from 'whatever/demo/config.json?raw`)? Its inclusion in each build would be determined by the effective module graph, without need for an extra build plugin and out of band filesystem action.

eyelidlessness (changes requested):
## Here's what I learned I need to test locally

I tested against the latest build artifact (found at bottom of the most recently run GitHub Actions result page).

I needed:

0. Web Forms Preview running in a local server. I used `yarn vite serve dist-demo`.
1. xlsform-online... **not pyxform-http** as I expected! Makes sense in hindsight, if I had realized they were distinct projects.
2. Changes in https://github.com/getodk/xlsform-online/pull/83
3. I also found the Dockerfile (https://github.com/getodk/xlsform-online/pull/84) helpful. Commands to run (fill in your own `$PATH_TO_TMP_DIR`):

    ```sh
		docker build --tag xlsform-online-wfpreview .
		docker run -e DJANGO_SECRET_KEY=secret -v $PATH_TO_TMP_DIR:/tmp_home -v $PATH_TO_TMP_DIR:/persistent_home -p 8000:8000 xlsform-online-wfpreview:latest
    ```

## Success...

I am able to preview XLSForms and XML XForms locally!

### ...mostly

On my first attempt with an XML XForm, I thought that part of the functionality was broken. But I saw the same error (below) with the same form. It turns out the error is caused by the form's reference to a GeoJSON external secondary instance. Here is the error I get:

```
Error: ODK Validate Errors:
Exception in thread "main" java.lang.NoClassDefFoundError: com${ObjectMapper}
Caused by: java.lang.ClassNotFoundException: com.fasterxml.jackson.databind.ObjectMapper
... 15 more
```

It's unclear whether there should be special accommodations for external secondary instances. I stopped debugging once I understood the root cause of the error, so I could focus on finishing a review pass here.

#### Workaround

Per @sadiqkhoja, a temporary workaround is to change `xlsforms-online` > `requirements.txt`:

```diff
- pyxform@git+https://github.com/xlsform/pyxform@master
+ pyxform@git+https://github.com/xlsform/pyxform@2c565d7
```

(Then for Docker, re-run `docker build` and `docker run` commands.)

@sadiqkhoja also guessed the commit with the root cause to be https://github.com/XLSForm/pyxform/commit/2ff0f89f3e24ae2fb3b429a3ebebdee867fc52aa.

## Feedback

I think it would be helpful to have clear dev instructions for how to run an integrated Web Forms Preview locally. This is a cross-project concern with some external changes still in flight, so I'm happy to defer and address our side in a follow up PR once the external dependencies are ready.

Otherwise this looks good and is pretty much what I was expecting!

I suggested a handful of minor changes. I think the ones with diff suggestions are pretty non-controversial. I would definitely like to minimize both the page weight and tooling impact if we can.

packages/web-forms/src/demo/FormUpload.vue
sadiqkhoja: done

packages/web-forms/src/demo/HomePage.vue
sadiqkhoja: done

packages/web-forms/src/demo/demo.ts
sadiqkhoja: I have made the change in https://github.com/getodk/web-forms/pull/223/commits/b3ae435863f5c09ea8068dcdc26396abd59e3bb7

https://www.npmjs.com/package/unplugin-fonts#import-generated-font-rules-css says to import `unfonts.css` if using custom or fontsource providers which we are.

So I have removed the manual import of fonts from `demo.ts` and `index.ts`; and imported `unfonts.css` in both. I hope this is the the right way. There is no impact on the bundle size of this change. I have also verified that fonts are loaded at runtime by running Demo app and @getodk/web-forms lib (`dist`) in Central.

packages/web-forms/vite.config.ts
sadiqkhoja: Doing that puts the content of the json file in `dist-demo/assets/index-<hash>.js`, which makes it difficult to update configuration value in the built artifact.

packages/web-forms/src/demo/FormPreview.vue
sadiqkhoja: done

packages/web-forms/src/assets/images/demo-screenshot@2x.jpg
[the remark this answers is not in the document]
sadiqkhoja: done

packages/web-forms/vite.config.ts
eyelidlessness: Gotcha. Fine for now, I'll think about it for a followup!

packages/web-forms/src/demo/demo.ts
eyelidlessness: As mentioned in the Slack thread, this change unfortunately broke font loading in the build artifact. Feel free to cherry pick [this commit](https://github.com/getodk/web-forms/commit/56f64ee59db553beae3e3a37691801ae066275b7) reverting the change and removing `unplugin-fonts` entirely. Sorry for the diversion on this one!

eyelidlessness (approved):
As mentioned, the most recent change to font loading caused them to break in the build artifact. Other than that, LGTM!

I'm approving in anticipation of any of these totally acceptable outcomes:

- incorporate my proposed fix ([commit](https://github.com/getodk/web-forms/commit/56f64ee59db553beae3e3a37691801ae066275b7))
- more narrowly roll back the font imports, defer the dependency change for later
- ignore the problem, shaving 100 KB off the page weight for now

PR #224, opened 2024-09-24
Update the feature matrix

lognaturel:
I noticed that XLSForm `trigger` was marked as in progress but that corresponds to XForms `setvalue` which is not yet in progress. ODK XForms `trigger` is exposed in XLSForm as `acknowledge`. I don't know the full history of this weirdness!

While I was there, I also noticed that we do have a full implementation of `likert` and that `table-list` is an XLSForm alias so doesn't need to be here.

I had some trouble figuring out how to run the script so I added a few more breadcrumbs. I also added a section header to try to improve the Github table of contents view.

This doesn't need a changeset because it's not end-user-facing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ This section is auto generated. Please update `feature-matrix.json` and then run
   <summary>
 
 <!-- prettier-ignore -->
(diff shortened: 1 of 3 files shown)


sadiqkhoja (approved):
Table of content still shows Latex syntax ☹️

PR #225, opened 2024-09-26
Minor changes to Web Form Preview

sadiqkhoja:

- [x] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Not applicable

Paired with @alyblenkin and @yanokwa to make sure they are satisfied with the changes.

All changes are related to copy and font sizes/weight. 

None

NA

## What's changed

- Font sizes and weight on the HomePage of Web Forms Preview
- Copy / text on HomePage

### DIFF
--- a/packages/web-forms/src/demo/DemoForm.vue
+++ b/packages/web-forms/src/demo/DemoForm.vue
@@ -79,7 +79,7 @@ const formXls = computed(() => {
 	h2 {
 		margin: 0.5rem 0;
 		font-size: 1.5rem;
-		font-weight: 700;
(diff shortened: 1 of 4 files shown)


eyelidlessness (approved):
I took a quick look at the areas I understand to be affected. LGTM!

While looking, I popped open the dev tools to quickly compare the before/after on the beta "chip". Definitely looks better. I did happen to notice that there's no top padding on the landing page for viewports smaller than the largest breakpoint. Not a huge deal, but could use a quick polish if there's interest.

PR #228, opened 2024-09-27
Fix: width overflow on mobile

eyelidlessness:

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [x] Safari (iOS)
- [ ] Not applicable

I mentioned in Slack I want to consider some minimal snapshot testing with screenshots. But as I say under **what's changed**, I don't want to delay this fix for that. I also don't know exactly what making that effective would look like, and I'd want to flag a couple more points of reference (especially in the actual form view).

It's definitely not the best possible solution. But it's the most minimal change I could find.

It's possible I missed some nuance in how the affected styles are overridden at larger breakpoints, but I am pretty confident that I didn't.

N/A

## What's changed

Looking at the Web Forms Preview as deployed on getodk.org, on my phone I noticed:

1. Many of the blocks were significantly wider than the viewport. This is especially noticeable for the blocks in the “Try real forms” section, where their description text overflows the viewport.

2. The non-iframe presentation is marginally better (i.e. you can actually scroll to see the overflowing content). But because the preview page **on getodk.org** is displayed in an iframe with `scrolling=“no”`, it is also impossible to scroll horizontally to view the overflowing content/read the demo form descriptions.

I believe this is the absolute minimum (albeit a little weird) fix possible for the issue. Longer term I think it would be nice to do another pass on how we’re handling this sort of layout, but I think it’d require reconsidering some of the page structure and I don’t want to delay this fix for that.

### DIFF
--- a/packages/web-forms/src/demo/HomePage.vue
+++ b/packages/web-forms/src/demo/HomePage.vue
@@ -151,7 +151,7 @@ const isDev = import.meta.env.DEV;
 
 	.container {
 		width: 100%;
-		max-width: 540px;
+		max-width: min(100%, 540px);
 		margin: 0 auto;
 		display: flex;
 		flex-direction: column;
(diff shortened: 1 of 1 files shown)


sadiqkhoja (approved):
Thanks, verified by deploying the artifact and tested on a phone. 🚀

PR #231, opened 2024-10-02
Add details about demo forms

lognaturel:
When looking at the preview site, I wanted to know more about why I would care about each form. This attempts to give a little information on what to expect.

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Not applicable

Ran it locally.

Wording changes only.

No regression risk.

No.


lognaturel:
Looks like CI might be getting rate limited: https://github.com/volta-cli/volta/issues/1523#issuecomment-1732187410 Will try to rerun later

lognaturel:
@yanokwa is rolling these into another PR.

PR #239, opened 2024-10-08
demo: enable .xml form upload

brontolosone:
## What & why

It's just a demo, but now at least the implementation allows the filetypes that the upload dialog allows: `.xml`, `.xlsx`, `.xls`.

This allows one to upload an `.xml` into the demo. Which is nice if you hack about in your `.xml`s by hand, in which case you you'll not have a correspondent `.xlsx` at hand (if there even is one, for that `.xml`).
It's also handy for when you don't have the xlsconverter server on hand and up and running.

- [X] Chromium
- [X] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Not applicable

I thought this would be the simplest thing that could possibly work, and indeed it turned out to work, and then I must admit to not considering any further approaches.

### DIFF
--- a/packages/web-forms/src/demo/FormUpload.vue
+++ b/packages/web-forms/src/demo/FormUpload.vue
@@ -27,6 +27,10 @@ const { data: config, error: configErrors } = useConfiguration();
 
 const xlsformOnlineBaseUrl = computed(() => config.value?.['xlsform-online-url']);
 
+const inDevMode = import.meta.env.DEV;
+
+const bypassConverterForXml = ref<boolean>(false);
+
 watch(configErrors, (value) => {
 	if (value) {
 		error.value =
@@ -77,8 +81,8 @@ const uploadFile = async (file: File) => {
 	reset();
 	uploadedFilename.value = file.name;
 
-	if (file.name.endsWith('.xml')) {
-		xformUrl.value = URL.createObjectURL(file);
+	if (bypassConverterForXml.value && file.name.endsWith('.xml')) {
+		xformUrl.value = URL.createObjectURL(file); // leaks; in non-demo code, we might want to deallocate the object URL at some point.
 	} else {
 		const { data: response, error: conversionError } = await convertXlsForm(file);
 
@@ -149,6 +153,14 @@ document.addEventListener(
 						<span class="icon-insert_drive_file" />
 						Drag and drop XLSForm or <a href="javascript:;" class="upload-file-link" @click="fileInput.click()">upload form</a>
 					</span>
+					<template v-if="inDevMode">
+						<label>
+							<input
+								v-model="bypassConverterForXml" type="checkbox"
+							>
+							Bypass converter for <code>XML</code> upload
+						</label>
+					</template>
 				</template>
 				<template v-else>
 					<PrimeProgressSpinner class="spinner" />


eyelidlessness (commented):
I see you had the same instinct as I did: we should be able to upload the format we directly support! That instinct did lead to a change to accept `.xml` files (though I don't recall exactly what the change involved).

I'm curious which of these (if either) prompted this change:

1. Did you encounter a problem with an `xlsform-online` integration not accepting a specific XML XForm?

2. Do you want to upload arbitrary XML XForms in your dev environment, without setting up and running `xlsform-online`?

- - -

If (1): I would suspect there might be either a bug somewhere in either of the projects, or in the integration between them. But I do think it's worth clarifying that we do actually want to upload XML XForms to the `xlsform-online` service. At least for the [Web Forms Preview](https://getodk.org/web-forms-preview/) public deployment. While that may be a bit unintuitive, we want to apply the same policies regardless of the form's source type.

If (2): I agree! Let's put our heads together on how to address that while preserving the intent in (1).

eyelidlessness:
> 2), as in the PR description

That's what I get for a rushed first look! Sorry I missed that on first read.

> 1): yes, remember that form I made that had cycles in its calculation graph? I couldn't upload that one to the [public demo](https://getodk.org/web-forms-preview/)

Ah, this is a usage scenario I hadn't considered! And your question is an excellent prompt for how I think it fits into my newly expanded mental model:

> So one question is: do we let it process `.xml` straight away, without it first passing through `ODK Validate`, already? Not just for devs but also for the public demo, even though `web-forms`' validation might be incomplete in comparison with `xlsform-online`'s?
> 
> I'd say "yes, why not".

I think the dev/public demo distinction is probably the most reasonable place to branch on this, at least for now. Our own use cases are somewhat different from likely users of the public-facing service. We have many very good reasons for wanting to interact with Web Forms—not just the form upload interface, but the entire stack—under conditions which are much less ideal for users.

It's worth emphasizing that, originally, this demo UI wasn't even conceived with support for uploading XML XForms. I advocated for adding the functionality, to expand our ability to dogfood the full project; and because it's quite convenient for dev flow to be able to try out ad hoc forms without adding them in the file system fixtures. For public usage, my understanding is that the intent is to be relatively close to what users will experience _in Web Forms, as deployed_, e.g.:

- in an equivalent to [XLSForm Online](https://getodk.org/xlsform/)
- Central's form preview functionality
- any other downstream embedding context with similar form preview functionality.

In these use cases, it's pretty likely the current validation functionality is what we want.

I can certainly see a case for some option like _I understand there's something wrong with this form, proceed anyway_. But I think we'd want to have a broader conversation with the team about the implications and design for something like that. In any case, I think it's also reasonable to revisit for a separate change/scope.

### Proposed dev solution

For our own dev use case, I think a more reasonable workflow would:

1. Make the `xlsform-online` integration optional; hopefully without extra steps, e.g. fallback when not configured and/or available
2. Make validation optional, even if integration is available; presumably this involves detecting that failure mode, then going through the same integration-optional flow
3. **In a separate change/scope:** If a validation error is encountered, while we don't want it to block, we likely want to know about it. We have near term work to introduce known error conditions, and this could be among them. That work also at least has a _notion of severity_, so this might be a good candidate to trial a "warning" or "info" condition.

As mentioned in our Slack chat, we can presumably leverage the dev-mode branching approach already in use to trigger these behaviors. I think 1-2 are relatively narrow in scope. If you feel good about the approach and feel like you have what you need to proceed, go for it! Or as mentioned in Slack I'd be happy to pair on it.

eyelidlessness (approved):
Thanks! The change is clear, but I validated each of these just to be totally sure it works as expected:

- Dev mode:
  - Checkbox option is present
  - Checked allows upload and preview of XML XForms without a running `xlsform-online` server
  - Unchecked works as before
- Build artifact:
  - Checkbox option is not present
  - Works as before

Issue #240, opened 2024-10-14
String to number conversion: Differences between JavaRosa and Webforms


brontolosone:
While working on #49, I stumbled upon some implementation differences between webforms and JavaRosa, that came about while thinking about "how and why does JR allow a randomization seed from a node with text, is that intentional or accidental, and should Webforms do the same" (so far I only have answers to #1 and maybe #2). Anyway:

The equivalent of a JS `Number` as used in Webforms is the Java `Double`. Same thing, same precision, so in theory this should allow identical outcomes on both platforms.

- [JavaRosa implementation](https://github.com/getodk/javarosa/blob/3efbb35b86cf22359eef85d426ca9b39af005714/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java#L702-L746) (think of it as a `toDouble`, as that's what it does)

- [The correspondent code site in Webforms might be here](https://github.com/getodk/web-forms/blob/1f868e80801e37e9d849337b1f2d9ffedd554adf/packages/xpath/src/evaluations/NodeEvaluation.ts#L60)

Evidently the JR implementation is much more extensive; and there's talk about "xpath type conversion rules" — which I haven't looked up, but those sound as if they could be relevant for webforms.

Poignantly, JR allows for `"123,123"` -> `123.123`, while Webforms makes this a `NaN`.
Conversely, Webforms allows `"0x12"` -> `18`, while JR would make this a `NaN`.

The question is thus: which one (if any) of the two is right? And if JR is right, what'd be the appropriate site for the Webforms implementation of same?

PR #241, opened 2024-10-14
choice randomization: better approximation of JR behaviour, fixes #49

### LINKED ISSUE #49: Seed to randomize should not require an integer
Per [spec](https://getodk.github.io/xforms-spec/#fn:randomize), `randomize` accepts an optional seed. The current implementation currently [checks that the seed is an integer](https://github.com/getodk/web-forms/blob/d37223250af06b57c73c4a57d4d76379a0b95fd5/packages/odk-xpath/src/lib/collections/sort.ts#L48-L49), and throws (a string, a tangentially related my-bad) if it's not. This was [derived](https://github.com/enketo/openrosa-xpath-evaluator/blob/master/src/utils/shuffle.js#L18-L20) from the implementation in OpenRosa XPath Evaluator, so I'd expect this bug affects Enketo as well. Which is somewhat surprising, because I discovered the bug by using XLSForms' [prominent example](https://xlsform.org/en/#randomize-choices) for stable-randomization.

### Steps to reproduce

1. Use the aforementioned XLSForms example in a form

### Expected behavior

1. A form using a decimal/float seed to `randomize` will stably randomize options, without error
2. Even if there is an error due to an actually invalid seed value, it will be surfaced to the user in some meaningful way
3. Possibly some sort of recovery behavior for certain invalid seed values (like `NaN` or `""`?), but this would potentially involve spec clarification

### Observed behavior

(From memory, happy to repro again if needed) Form load fails without explanation.

brontolosone:
Closes #49

- [x] Chromium
- [x] Firefox

Some related problems remain to be solved:
- #240 
- getodk/javarosa#800

This brings what Webforms does more in line (barring #240) with what Javarosa does, and as such fixes the immediate problem of #49.

I felt it was worth it to be verbose with the comments here, so check those out.

This story is not over yet. Depending on whether we deem it OK to change the seed derivation algo, I'd like to make it value type/length agnostic and would just hash the input in its textual form and derive a seed from that hash - see getodk/javarosa#800. And in that case this code will need to be altered again.

### DIFF
--- a/packages/xpath/src/functions/xforms/node-set.ts
+++ b/packages/xpath/src/functions/xforms/node-set.ts
@@ -410,14 +410,16 @@ export const randomize = new NodeSetFunction(
 );
 
 const toBigIntHash = (text: string): bigint => {
-	// Hash text with sha256, and interpret the first 64 bits of output
-	// (the first and second int32s ("words") of CryptoJS digest output)
-	// as an int64 (in JS represented in a BigInt).
-	// Thus the entropy of the hash is reduced to 64 bits, which
-	// for some applications is sufficient.
-	// The underlying representations are big-endian regardless of the endianness
-	// of the machine this runs on, as is the equivalent JavaRosa implementation
-	// at https://github.com/getodk/javarosa/blob/ab0e8f4da6ad8180ac7ede5bc939f3f261c16edf/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java#L718-L726
+	/**
+	Hash text with sha256, and interpret the first 64 bits of output
+	(the first and second int32s ("words") of CryptoJS digest output)
+	as an int64 (in JS represented in a BigInt).
+	Thus the entropy of the hash is reduced to 64 bits, which
+	for some applications is sufficient.
+	The underlying representations are big-endian regardless of the endianness
+	of the machine this runs on, as is the equivalent JavaRosa implementation.
+	({@link https://github.com/getodk/javarosa/blob/ab0e8f4da6ad8180ac7ede5bc939f3f261c16edf/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java#L718-L726 | see here}).
+	*/
 	const buffer = new ArrayBuffer(8);
 	const dataview = new DataView(buffer);
 	SHA256(text)


eyelidlessness (changes requested):
This is great. I **really appreciate** how the commentary here tells the _why_ story!

As discussed a bit in Slack, I think a couple of adjustments would make the change clearer. And would either make the need for some of the commentary moot, or make the remaining commentary more useful.

1. There's a clear "JavaRosa compatibility" responsibility here. While it is inherently coupled to the `seededRandomize` implementation, it is also very specifically a mapping to a more general concept: `longValue`. I think it would help immensely for future understanding of what's going on here if we make that an explicit function, with the same name.

2. In general, I've found liberal use of JSDoc comments (i.e. `/** ... */`) really helpful. The comment style provides support for all sorts of editor functionality.

    Here, we'd get a lot of benefit from inline linking support (i.e. `{@link $URL_OR_REFERENCE}` and/or `{@link $URL_OR_REFERENCER | more specific title}`). In particular I think a permalink to the JavaRosa `resolveRandomSeed` method would be useful, and probably also links to the pertinent issues. This ties directly to point 1: giving the JavaRosa-compatible thing a name corresponding to the Java thing it emulates, also gives a clear place for that JSDoc to reference it and clarify the nuances it's addressing.
    
    I think a few tweaks to [this comment](https://github.com/getodk/web-forms/pull/241/files#diff-66bb4539d9279bf5b487e8f56ffbd07498f8400e356ed42c0932c8e1441f6335R49-R56) would be pretty much perfect.

3. We can eliminate the divergences from JavaRosa by using BigInt values for several of these cases. This, combined with their usage in a clearly articulated `longValue` JR-equivalent would also eliminate the need for commentary on those cases. In my local exploration of this, what I found made the most sense with the least fuss was to change `type Int = number` to `type Int = bigint | number`, then have the `longValue` equivalent also produce that `Int` type. Pretty much everything else that would need to change falls out of that (i.e. any mixed-type operators producing fractional values do the appropriate explicit `Number()` casts to preserve those mathematical semantics).

    It'd also be useful for the `Infinity`/`-Infinity` cases to be bound to constants with clear names. Insofar as there's still benefit to commentary on those, JSDoc on those constants is a good place.


Aside from making some of the intent clearer here, I suspect we may find there are other edge cases where we want to cordon off JavaRosa-compat/Java-isms in a general and reusable way. Even if that seems like a premature abstraction, doing it in this case is a direct, 1:1 linkable reference to the existing abstraction we'll be emulating.

packages/xpath/src/functions/xforms/node-set.ts
```diff
@@ -1,3 +1,5 @@
+import sha256 from 'crypto-js/sha256';
+
 import type { XPathNode } from '../../adapter/interface/XPathNode.ts';
```
eyelidlessness, line 3: Minor nits:

```suggestion
import { SHA256 } from 'crypto-js';
import type { XPathNode } from '../../adapter/interface/XPathNode.ts';
```

This change will result in a slightly smaller bundle size, both because it's consistent with the project's other import of the same function and because it potentially improves the bundler's ability to do tree shaking. (In general, named imports are better for that.)

Of course you'd need to update the `sha256` call to uppercase as well. Also note removal of the blank line between imports. Won't matter here, but in general that allows the editor (or whatever in your workflow handles auto-formatting) to auto-sort imports.

- - -

Normally I'd prefer to handle code style issues like the blank line with automated tooling (maybe we can follow up to find an ESLint rule that addresses this?). But it felt worth mentioning while there's a more substantive change to address in the same place.

packages/xpath/src/functions/xforms/node-set.ts
```diff
@@ -384,8 +386,39 @@ export const randomize = new NodeSetFunction(
 
 		const nodeResults = Array.from(results.values());
 		const nodes = nodeResults.map(({ value }) => value);
-		const seed = seedExpression?.evaluate(context).toNumber();
-
-		return seededRandomize(nodes, seed);
+		if (seedExpression === undefined) return seededRandomize(nodes);
```
eyelidlessness, line 389: **No need to change here** (unless you agree/feel like it), but this tripped me up. I don't have a strong preference against single-line conditionals like this, but I'd generally prefer they're surrounded by a blank line so it's clear where they begin and end.

(But please feel free to ignore for now! I would definitely prefer to automate code style issues like this. I'm just mentioning it because it made me pause longer than I anticipated to make sure I understood the code.)

packages/xpath/src/functions/xforms/node-set.ts
```diff
@@ -384,8 +386,39 @@ export const randomize = new NodeSetFunction(
...
 );
+
+function toBigIntHash(text: string): bigint {
+	// hash text with sha256, and interpret the first 64 bits of output
+	// (the first and second int32s ("words") of CryptoJS digest output)
+	// as a BigInt. Thus the entropy of the hash is reduced to 64 bits, which
+	// for some applications is sufficient.
+	// The underlying representations are big-endian regardless of the endianness
+	// of the machine this runs on, as is the equivalent JavaRosa implementation
+	// at https://github.com/getodk/javarosa/blob/ab0e8f4da6ad8180ac7ede5bc939f3f261c16edf/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java#L718-L726
```
eyelidlessness, line 417:
```suggestion
/**
 * Hash text with sha256, and interpret the first 64 bits of output (the first
 * and second int32s ("words") of CryptoJS digest output) as a BigInt. Thus the
 * entropy of the hash is reduced to 64 bits, which for some applications is
 * sufficient. The underlying representations are big-endian regardless of the
 * endianness of the machine this runs on, as is the
 * {@link https://github.com/getodk/javarosa/blob/ab0e8f4da6ad8180ac7ede5bc939f3f261c16edf/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java#L718-L726 | equivalent JavaRosa implementation}.
 */
const toBigIntHash = (text: string): bigint => {
```

As a JSDoc comment, this allows the same documentation to be accessed at the call site.

Switching to an arrow function is somewhat a nit, but it's generally preferable to avoid unnecessary `function` functions as they have confusing behavior. (Maybe that's also a thing we could lint?)

packages/xpath/src/lib/collections/sort.ts
```diff
@@ -15,11 +15,18 @@ class UnseededPseudoRandomNumberGenerator implements PseudoRandomNumberGenerator
...
-		let initialSeed = seed % SEED_MODULO_OPERAND;
-
+	constructor(seed: Int | bigint) {
+		let initialSeed: number;
+		if (typeof seed === 'bigint') {
+			// the result of the modulo operation is always smaller than Number.MAX_SAFE_INTEGER,
+			// thus it's safe to convert to a Number.
+			initialSeed = Number(BigInt(seed) % BigInt(SEED_MODULO_OPERAND));
+		} else {
+			initialSeed = Number(seed) % Number(SEED_MODULO_OPERAND);
```
eyelidlessness, line 28:
```suggestion
			initialSeed = seed % SEED_MODULO_OPERAND;
```

Both operands are already `number` here.

packages/xpath/src/lib/collections/sort.ts
```diff
@@ -38,17 +45,32 @@ class SeededPseudoRandomNumberGenerator implements PseudoRandomNumberGenerator {
...
+		// In Java, a NaN double's .longValue is 0
+		if (Number.isNaN(seed)) finalSeed = 0;
+		// In Java, an Infinity double's .longValue() is 2**63 -1, which is larger than Number.MAX_SAFE_INTEGER, thus we'll need a BigInt.
+		else if (seed === Infinity) finalSeed = 2n ** 63n - 1n;
+		// Analogous with the above conversion, but for -Infinity
+		else if (seed === -Infinity) finalSeed = -(2n ** 63n);
+		// A Java double's .longValue drops the fractional part.
+		else if (typeof seed === 'number' && !Number.isInteger(seed)) finalSeed = Math.trunc(seed);
+		else finalSeed = seed;
+		generator = new SeededPseudoRandomNumberGenerator(finalSeed);
```
eyelidlessness, line 73: I think this is where I was focused when I suggested cordoning off the JavaRosa compatibility concern. I also think there's a way to do it that feels both natural and idiomatic: as a subclass of `SeededPseudoRandomNumberGenerator`, with a name that clarifies that responsibility beyond the base Park-Miller algorithm. This logic would then fit neatly in its constructor. I'd be thrilled if we could find a way to translate some of these inline comments into a meaningful JSDoc block too, but I wouldn't push too hard for it. A named class would go a long way toward clarity of where the responsibility boundaries are.

(And perhaps it would be clearer still if that base class was renamed to indicate its Park-Miller-only responsibility?)

packages/xpath/test/xforms/randomize.test.ts
```diff
@@ -18,6 +18,9 @@ describe('randomize()', () => {
 	});
 
 	const SELECTOR = '//xhtml:div[@id="FunctionRandomize"]/xhtml:div';
+	const MIRROR = 'mirror';
+	const MIRROR_HASH_VALUE = 5989458117437254; // in Python: "from struct import unpack; from hashlib import sha256; unpack('>Q', sha256(b'mirror').digest()[:8])[0]"
```
eyelidlessness, line 22: What is this Python reference to?

packages/xpath/src/functions/xforms/node-set.ts
```diff
@@ -384,8 +386,39 @@ export const randomize = new NodeSetFunction(
 
 		const nodeResults = Array.from(results.values());
 		const nodes = nodeResults.map(({ value }) => value);
-		const seed = seedExpression?.evaluate(context).toNumber();
-
-		return seededRandomize(nodes, seed);
+		if (seedExpression === undefined) return seededRandomize(nodes);
+		const seed = seedExpression.evaluate(context);
+		const asNumber = seed.toNumber(); // TODO: There are some peculiarities to address: https://github.com/getodk/web-forms/issues/240
```
eyelidlessness, line 391: I'm not sure this comment belongs here. It isn't specific to this cast, it's specific to casting to XPath `number` throughout. Fine to leave since we have an issue tracking it, but we'll probably just find it went stale some time after we address the issue.

packages/xpath/src/functions/xforms/node-set.ts
```diff
@@ -384,8 +386,39 @@ export const randomize = new NodeSetFunction(
...
+			const seedString = seed.toString();
+			if (seedString === '') {
+				finalSeed = 0; // special case: JR behaviour
+			} else {
+				// any other string, we'll convert to a number via a digest function
+				finalSeed = toBigIntHash(seedString);
+			}
+		} else {
+			finalSeed = asNumber;
+		}
```
eyelidlessness, line 405: Doesn't "special case: JR behavior" apply to all of this?

eyelidlessness (changes requested):
Thanks! I think this is really close. Most of my remaining feedback is around code _clarity_ (naming, separation of responsibilities, accessibility and applicability of comments).

packages/xpath/src/functions/xforms/node-set.ts
eyelidlessness: I'll try to make the case for JSDoc. If you're open to reconsidering here, that would be excellent. If not, I think we should come back to this discussion as a team.

- - -

JSDoc comments are a standard designed to encode structured documentation about any symbol they're attached to.

The editor support alone goes well beyond linking to URLs. For example, the ability to reference documentation across modules are invaluable. Linking to _other symbols_ (both within and across modules) is also invaluable, both as a navigation tool and because **they can be kept up to date** as those symbols change.

Beyond editor support, being a standard for structured documentation and association with symbols, JSDoc can be used for _documentation output_. We are already using this to generate documentation for `@getodk/xforms-engine`. I'd quite like to expand that to other use cases.

I share your distaste for some of the syntax minutiae of JSDoc, and `@link` is particularly weird (I suspect this is because inline tags are relatively rare). But that distaste doesn't outweigh the overwhelming benefits of an extensible documentation standard which is widely adopted in tooling we already use. It is also widely adopted throughout this project, and across the ecosystem; which is to say, it is both locally and globally idiomatic.

I also noticed GitHub's odd presentation in a couple diff suggestions in this PR. It's worth noting that:

- That's not representative of how GitHub presents JSDoc in complete source
- It's not representative of how GitHub presents JSDoc in diffs broadly
- GitHub's syntax highlighting is notoriously inconsistent across various views
- The highlighting is applied to an incomplete (i.e. syntactically invalid) chunk of code, which likely exacerbates potential issues

Lastly, I am sensitive to poor authoring ergonomics. I'm a bit surprised to hear that your editor doesn't make adding/editing JSDoc comments _easier than single line comments_, as that's my experience in the editors I'm familiar with. If this is a major hangup, I'd be happy to help look into ways to make the authoring experience nicer for you.

eyelidlessness (approved):

Issue #202, opened 2024-08-27
External secondary instances: Error conditions


eyelidlessness:
This design issue is part of broader support for external secondary instances:

- [I/O support for `jr:` URLs][io-design]
- **[Error conditions][error-conditions]**
- [Engine representation, XPath support][engine-xpath]

[io-design]: #io-design
[error-conditions]: #error-conditions
[engine-xpath]: #engine-xpath

This issue is focused on:

- establishing a goal to define error conditions specific to the external secondary instance feature
- designing the engine/client interface to document and convey such error conditions
- anticipating nuances of error conditions as such, and other error conditions not covered by this feature but affected by this design decision

## Anticipated error conditions

The error conditions I expect we'll need to handle fall into three buckets:

- I/O failures (such as poor/intermittent connectivity, verifiable absence of referenced resources)
- Resources available but malformed
- Unsupported resource data types

I don't intend to detail all of the specific cases _yet_, but I think we should be prepared to define a specific set of expected error conditions as a prerequisite for implementation. This way we can plan and design accordingly for the various downstream user flow and/or messaging suitable for each case.

Worth mentioning on that note: depending on how we prioritize functionality, we may want to take special care to distinguish between unsupported data types _we know about, and plan to support_ versus those we simply don't recognize.

## Deeper consideration of "error" as a binary state

There's probably room for nuance in both buckets. We might want to also consider some partial and recoverable failure cases:

- Should unavailability of some subset of form-referenced resources always block form load?
- Can we detect poor/intermittent connectivity and retry requests?
- Can we defer failure caused by unavailable resources, until they're required for progress filling/accessing specific subsets of a form?
- Are there known recovery techniques for certain kinds of malformed data?

I would categorize these as **future considerations**, out of scope for now unless there's a compelling reason to prioritize otherwise.

But I mention them both because it's good to anticipate future goals, and because they're important to keep in mind as we consider the implications of the next section.

## Goal: represent error conditions on par with success conditions

This is something I wish I had made more explicit when I wrote up #188, to avoid confusion on why some of the proposal is shaped the way it is. To be very clear:

- **Documentation-first**: an ongoing goal of designs for `@getodk/xforms-engine` is to make documentation a core work product, in a way that is and remains consistent with the documented functionality.

- **Documentation of failure modes** (and other non-happy path cases): a long-discussed goal—now coming into focus as we take on functionality with inherent fallibility—is to ensure that failure modes and error conditions are documented in a way that's consistent with success modes.

[...]

PR #244, opened 2024-10-25
Distance function: multiple arguments

### LINKED ISSUE #118: Add multiple-argument distance variant
https://github.com/getodk/javarosa/pull/761

brontolosone:
Closes #118

One can now call distance() with multiple arguments.

- [X] Chromium
- [X] Firefox

- Ran `yarn test-node:jsdom xforms` in the `packages/xpath` directory.

It might not be; the argument count error handling is not ideal. When I create a form that calls `distance()` (so 0 arguments) an error is thrown (good) but the message is "Argument 1 not allowed" which is rather odd. But I've been told the errors are going to get reworked, some rethinking is slated to happen in #202.

The below is handy. Didn't work before, but now it does.
[distance.xlsx](https://github.com/user-attachments/files/17526386/distance.xlsx)


packages/xpath/src/evaluator/functions/FunctionImplementation.ts
```diff
@@ -127,6 +127,12 @@ export class FunctionImplementation<Length extends number> {
 							max: Infinity,
 						};
 
+					case 'variadic+':
```
lognaturel, line 130: My naive outsider expectation would be that the parameter would still be `variadic` and that the implementation of the function would explicitly throw (or maybe eventually return an error result) for the 0 case. It feels to me not like the parameter type is not variadic but like the specific function has an additional requirement.

Something related that comes to mind is `indexed-repeat`. It needs to be given a path first and then pairs of paths and indexes. That means it is not allowed to have an even number of total parameters. I'm pretty sure an exception will be thrown if pairs aren't supplied https://github.com/getodk/web-forms/blob/9d83384cc10fee25ccfe2b5fd60cae62cde6481d/packages/xpath/src/functions/xforms/node-set.ts#L127 It feels similar to me as here because there's a function-specific constraint on "variadic".

JavaRosa makes these checks explicit: https://github.com/getodk/javarosa/blob/0d6c2eaa3704d28a5be4c3fc723baf56725f7bff/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java#L182 and https://github.com/getodk/javarosa/blob/0d6c2eaa3704d28a5be4c3fc723baf56725f7bff/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java#L498 That does mean the user can get a specific, helpful error message.

PR #245, opened 2024-10-31
XPath support for external secondary instances

### LINKED ISSUE #203: External secondary instances: Engine representation, XPath support

eyelidlessness:
Closes #203.

I'd normally include a more comprehensive writeup, especially for such a large change. But #203 effectively covers the design/approach, and I believe the extensive commit notes better represent this set of changes than anything I would write here. As we discussed this morning, we'll let those commit notes speak for the change.


lognaturel (approved):
We've decided to merge this now even though it has only received cursory review so far. It's a conceptual change that @eyelidlessness has been planning for a long time and there's enough testing coverage that it's unlikely to have broken anything.

If we don't merge now, we'll instead stack commits on top. Even if there are changes to be made, we think these commits will be valuable to have access to in the future.

Next steps:
- @eyelidlessness will remove no-op review cutpoint commits and merge
- @eyelidlessness will add support for XML, CSV and GeoJSON external secondary instances
- @brontolosone will interact with the implementation as he's working on the host application integration and will file issues or PRs as appropriate

PR #259, opened 2024-11-26
Support for external secondary instances

### LINKED ISSUE #201: External secondary instances: I/O support for `jr:` URLs
This design issue is part of broader support for external secondary instances:

- **[I/O support for `jr:` URLs][io-design]**
- [Error conditions][error-conditions]
- [Engine representation, XPath support][engine-xpath]

[io-design]: https://github.com/getodk/web-forms/issues/201
[error-conditions]: https://github.com/getodk/web-forms/issues/202
[engine-xpath]: https://github.com/getodk/web-forms/issues/203

The intent in this issue is to decide on a direction for how we want to handle retrieval of form attachments _generally_, with _specific goal_ of supporting external secondary instance functionality.

These are the pertinent aspects of the ODK XForms spec.

- Broadly: [Secondary Instances - External](https://getodk.github.io/xforms-spec/#secondary-instances---external)
- More specifically: [File Endpoints](https://getodk.github.io/xforms-spec/#file-endpoints)

The issue will focus on the engine/client interface to support the retrieval of `jr:` URLs. As I tend to do for engine/client interface design, I will present a few options which we can choose from or iterate on. Implementation in the engine will be derived from there.

**Note:** it is expected that the design we choose here _will also_ lay groundwork for supporting other `jr:` URL use cases—i.e. media attachments—so I've tried to be mindful of that (as we should in discussion).

**Note:** this issue does not currently address the [`jr://instance/last-saved` virtual endpoint](https://getodk.github.io/xforms-spec/#virtual-endpoints), but I believe nothing in any of the proposed options would block or impede that functionality, when we're ready to address it.

## Option 0: client/host application handles resource resolution

This is a true null option: in the narrowest sense, we **could** claim this work is already done with the provision of a [`fetchResource`](https://github.com/getodk/web-forms/blob/1ce92a838b2a9928d3833cf8c3da5025cd917ef9/packages/xforms-engine/src/client/EngineConfig.ts#L82) configuration option. This option is _technically sufficient_ to satisfy the engine's spec responsibilities.

How this would work:

- The engine will call `fetchResource` with any external secondary instance's `jr:` URL
- If the client has provided a `fetchResource` option, the client is then responsible for resolving that `jr:` URL to the referenced resource, for the active form instance
- If the client has not provided a `fetchResource` option, the engine will produce a well-defined error result (with errors to be discussed in a separate design)

At least until we support offline mode (or any other functionality that would imply runtime-level caching/persistence), this would leave resource resolution entirely to clients.

This is sort of the opposite of a "pit of success" option, with its primary appeal being limited engine-side work for this aspect of the targeted feature.

Beyond obvious non-"pit of success" drawbacks, I'll specifically note that it's the most likely option to result in disparities and drift between clients. It's also likely to promote disparities/drift between different functionality which intersect with it.[^1]

[^1]: This has been a particular pain point in Enketo. Support for `jr:` URLs is spread across three packages, and difficult to iterate on even after moving the projects to a monorepo.

## Option 0.1: engine does not handle this aspect _at all_

Another null option variant.

This would effectively mean that clients must resolve `jr:` URLs **before initializing a form**. They'd probably supply the resources as `data:` or `blob:` URLs, substituted directly in the form definition provided by clients to the engine.

This option does not appeal to me, but I think it's worth mentioning so we can make a thoroughly informed decision.

## Option 0.5: engine provides resolution handler(s) for common cases

An extension of option 0, this is similar in spirit to the submission API proposal (#188), and some of the discussion ongoing there. The idea would be that we recognize one or more typical resource mapping schemes, and expose default `fetchResource` implementations to address those (likely as some kind of factory function so clients can parameterize them for per-instance usage appropriately).

I would imagine starting with handlers for:

- the OpenRosa Protocol [media manifest](https://docs.getodk.org/openrosa-form-list/#the-manifest-document)
- perhaps also Central's [form attachment list API](https://docs.getodk.org/central-api-form-management/#listing-form-attachments)

## Option 1: engine provides one or more explicit mechanisms for form attachment resolution, tailored to feature-specific use cases

Instead of the engine calling a generalized `fetchResource` option _with a `jr:` URL_, the engine would instead accept a configuration mapping between specific `jr:` URLs to one of:

- the resource's _real URL_, or some other `fetch`-able URL (`blob:`, `data:`, ?); this mapped URL would **then** be accessed by the same `fetchResource` option
- anything else capable of resolving a `Blob` of the resource's data (`Promise<Blob>`, `() => Promise<Blob>`, ?)

The mapping itself could be any of:

- the OpenRosa manifest format
- the response type from Central's form attachment list API
- a simple `Map`-like object (or even `Record<string, T>` if we're feeling really loosey goosey about it)

eyelidlessness:
Closes #201

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [x] Safari (iOS)
- ~~[ ] Not applicable~~

- Some existing tests ported from JavaRosa now pass
- At least one happy path test for each format (XML, CSV, GeoJSON)
- Added a slew of CSV parsing tests based on conversation yesterday with @lognaturel
- Added/updated integration in `@getodk/web-forms` dev/preview modes:
  - Web Forms Preview: when a user uploads a form referencing an external secondary instance, previewing the form will behave as if those instances are blank/empty.
  - Dev mode: when loading a fixture from the file system, it will load any referenced external secondary instances from files of the same name, in the same directory as that fixture; if any referenced file does not exist, the form will fail to load.

  These cases effectively represent the behaviors an integrating host application can expect. The latter is the default behavior for missing external secondary instance resources. The former requires a configuration option, and should be suitable for e.g. Central's form preview mode.

Worth noting: there is still a failing test ported from JavaRosa, where an error is expected when referencing a node that doesn't exist in a secondary instance. This test technically references an external secondary instance, but it isn't clear whether the behavior would be applicable to all secondary instances.

Another test ported from JavaRosa is now marked failing, representing the same default behavior for missing external secondary instances described above. A companion test has been added representing the optional/configured alternate behavior.

This builds on the work in #245, and _mostly_ follows the design laid out in #201. It deviates somewhat from that design, mostly addressing some of the later discussion on that thread. I'll add more detail on that deviation below.

The main regression risk I can think of is that the newer default behavior is technically stricter for forms with external secondary instance references. Prior to this PR, those forms will always treat all such instances as blank. As of this PR, by default they'll be loaded and parsed by default, and evaluated as appropriate for any `instance(...)/...` expression defined in the form.

The other regression risk is that this does introduce breaking changes for host applications. Those will also be detailed below.

This is pretty much covered by existing/added test fixtures. If you want to try it with your own form/attachments, they can be placed together in any fixtures subdirectory and accessed in dev mode.

### Breaking changes: host application <-> `@getodk/web-forms`

1. To support **loading** external secondary instances, a host application must implement a `fetchFormAttachment` function.
2. To provide a graceful fallback for loading a form **without loading** external secondary instances, a host application must override the default `missingResourceBehavior` option.

Both of these are specified (likely in `kebab-case`) as props on the `OdkWebForm` component.

### Breaking changes: client <-> engine

The same host/client breaking changes apply to integration between clients and the engine. These breaking changes have been implemented in `@getodk/web-forms` to support host application integration. They've also been implemented in `@getodk/scenario` to reflect the functionality in existing/added tests.

**Note:** I am only realizing now that I've neglected to integrate these changes in `@getodk/ui-solid`. I am honestly not sure how valuable the package is at this point, so I'm hesitant to address this unless anyone feels strongly otherwise. If we all agree that the package itself isn't providing much value now, I'd like to consider removing it sometime soon (so it doesn't languish in a weird semi-broken state).

### Engine configuration: `fetchFormAttachment`

The interface for this configuration is (simplified for reading inline):

```ts
type FetchFormAttachment = (resource: URL & { protocol: 'jr:' }) => Promise<FetchResourceResponse>
```

... where `FetchResourceResponse` is the same subset of `Response` already exposed for loading remote XForm XML resources, slightly expanded to allow an optional `headers` property (which in turn is a subset of the native `Headers` type).

This deviates from the design in #201: rather than providing a `Map` (or other map-like structure) for resolving `jr:` URLs, clients are expected to provide a `fetch`-like function. This deviation from the design is intended to address discussion in that issue around the appropriate _keys_ for such a map-like structure.

In spirit, it makes sense that the keys should be the full `jr:` URL (which I believe reflects @brontolosone's view in that discussion). In practice, a client and/or host will almost certainly be resolving those `jr:` URLs from a file name without additional information. In order for a caller to **be able** to satisfy a mapping between `jr:` URL and something resolving it, we'd need to provide a parsed set of URLs to satisfy. We can certainly do this! And we've discussed doing it. But as a primary API, it would complicate the form initialization process by introducing a step I believe is unnecessary.

This design change is a compromise recognizing that these are conceptually equivalent concepts:

- map-like structure whose values may be accessed by a suffix of its canonical keys
- function of canonical key to value

I briefly considered providing a sample implementation for the two most common formats we'd expect to derive this configuration: OpenRosa Manifest documents and the Central Form Attachments API. But I decided that implementing them is trivial enough that it shouldn't hold up the PR. I'll be happy to assist with integration if it turns out to be less obvious downstream than I hope.

### Engine configuration: `missingResourceBehavior`

This configuration is mostly described above. It reflects a duality of expected behaviors:

- By default, a form is not complete without its external secondary instances
- There are specific use cases where that default is unsuitable (e.g. various use cases conveniently all called "preview")

The configuration options for this are exposed as constants, but a bare string is acceptable as well.

### Notable engine internal implementation details

This is mostly covered in 0fbc3c5 and its commit notes. The most important detail is that parsing and loading is intentionally staged/layered:

1. I/O
2. Data format detection
3. Per-format parsing logic
4. Common representation of all **external** secondary instances
5. Finally, common representation of **all** secondary instances

That final common representation is then provided through `PrimaryInstance` to its instance of `XFormsXPathAdapter`. From there, evaluation of a form's `instance(...)/...` expressions makes no distinction between any secondary instance's type (internal vs external), source (XForm definition, `FetchFormAttachment` response) or format (XML DOM, XML text, CSV, GeoJSON).

There is also a bit of groundwork in the I/O layer anticipating support for _media attachments_. Not much is done there other than recognizing the entrypoint for that functionality will likely flow through a common `FormAttachmentResource` abstraction (as do external secondary instance resources now).

### Acknowledgement: some of the `StaticNode` interfaces are _super awkward_

(Where `StaticNode` and its subclasses are part of the engine's DOM adapter representation introduced in #245.)

Parsing logic for both CSV and GeoJSON introduce some indirection around these representations to make them a little bit easier to reason about/more ergonomic to use. I think there's some room to simplify the base constructor signatures further, but I decided to put that off and live with that bit of tech debt for now. I'm eager to land the functionality, and expect others on the team are as well. So let's call this ready for review!

### DIFF
--- a/packages/common/src/fixtures/xform-attachments.ts
+++ b/packages/common/src/fixtures/xform-attachments.ts
@@ -1,0 +1,137 @@
+import { UpsertableMap } from '../lib/collections/UpsertableMap.ts';
+import { UnreachableError } from '../lib/error/UnreachableError.ts';
+import type { ImportMetaGlobLoader } from './import-glob-helper.ts';
+import { toGlobLoaderEntries } from './import-glob-helper.ts';
+
+/**
+ * @todo Support Windows paths?
+ */
+const getFileName = (absolutePath: string): string => {
+	const fileName = absolutePath.split('/').at(-1);
+
+	if (fileName == null) {
+		throw new Error(`Failed to get file name for file system path: ${absolutePath}`);
+	}
+
+	return fileName;
+};
+
+// prettier-ignore
+const xformAttachmentFileExtensions = [
+	'.csv',
+	'.geojson',
+	'.xml',
+	'.xml.example',
+	'.xlsx',
+] as const;
+
+type XFormAttachmentFileExtensions = typeof xformAttachmentFileExtensions;
+type XFormAttachmentFileExtension = XFormAttachmentFileExtensions[number];
+
+const getFileExtension = (absolutePath: string): XFormAttachmentFileExtension => {
+	for (const extension of xformAttachmentFileExtensions) {
+		if (absolutePath.endsWith(extension)) {
+			return extension;
+		}
+	}
+
+	throw new Error(`Unknown file extension for file name: ${getFileName(absolutePath)}`);
+};
+
+const getParentDirectory = (absolutePath: string): string => {
+	const fileName = getFileName(absolutePath);
+
+	return absolutePath.slice(0, absolutePath.length - fileName.length - 1);
+};
+
+const xformAttachmentFixtureLoaderEntries = toGlobLoaderEntries(
+	import.meta,
+	import.meta.glob<true, 'url', string>('./*/**/*', {
+		query: '?url',
+		import: 'default',
+		eager: true,
+	})
+);
+
+export class XFormAttachmentFixture {
+	readonly fileName: string;
+	readonly fileExtension: string;
+	readonly mimeType: string;
+
+	constructor(
+		readonly absolutePath: string,
+		readonly load: ImportMetaGlobLoader
+	) {
+		const fileName = getFileName(absolutePath);
+		const fileExtension = getFileExtension(fileName);
+
+		this.fileName = fileName;
+		this.fileExtension = fileExtension;
+
+		switch (fileExtension) {
+			case '.csv':
+				this.mimeType = 'text/csv';
+				break;
+
+			case '.geojson':
+				this.mimeType = 'application/geo+json';
+				break;
+
+			case '.xml':
+			case '.xml.example':
+				this.mimeType = 'text/xml';
+				break;
+
+			case '.xlsx':
+				this.mimeType = 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet';
+				break;
+
+			default:
+				throw new UnreachableError(fileExtension);
+		}
+	}
+}
+
+type XFormAttachmentFixtureEntry = readonly [absolutePath: string, fixture: XFormAttachmentFixture];
+
+type XFormAttachmentFixtureEntries = readonly XFormAttachmentFixtureEntry[];
+
+const xformAttachmentFixtureEntries: XFormAttachmentFixtureEntries =
+	xformAttachmentFixtureLoaderEntries.map(([absolutePath, load]) => {
+		const fixture = new XFormAttachmentFixture(absolutePath, load);
+
+		return [absolutePath, fixture];
+	});
+
+type XFormAttachmentFixturesByAbsolutePath = ReadonlyMap<string, XFormAttachmentFixture>;
+
+const buildXFormAttachmentFixturesByAbsolutePath = (
+	entries: XFormAttachmentFixtureEntries
+): XFormAttachmentFixturesByAbsolutePath => {
+	return new Map(entries);
+};
+
+export const xformAttachmentFixturesByPath = buildXFormAttachmentFixturesByAbsolutePath(
+	xformAttachmentFixtureEntries
+);
+
+type XFormAttachmentFixturesByDirectory = ReadonlyMap<string, readonly XFormAttachmentFixture[]>;
+
+const buildXFormAttachmentFixturesByDirectory = (
+	entries: XFormAttachmentFixtureEntries
+): XFormAttachmentFixturesByDirectory => {
+	const result = new UpsertableMap<string, XFormAttachmentFixture[]>();
+
+	for (const [absolutePath, fixture] of entries) {
+		const parentDirectory = getParentDirectory(absolutePath);
+		const subset = result.upsert(parentDirectory, () => []);
+
+		subset.push(fixture);
+	}
+
+	return result;
+};
+
+export const xformAttachmentFixturesByDirectory = buildXFormAttachmentFixturesByDirectory(
+	xformAttachmentFixtureEntries
+);
--- a/packages/common/src/jr-resources/JRResourceURL.ts
+++ b/packages/common/src/jr-resources/JRResourceURL.ts
@@ -30,7 +30,7 @@ export class JRResourceURL extends URL {
 	}
 
 	static isJRResourceReference(reference: string | null): reference is JRResourceURLString {
-		return reference?.startsWith('jr:') ?? false;
+		return reference?.startsWith(JR_RESOURCE_URL_PROTOCOL) ?? false;
 	}
 
 	declare readonly protocol: JRResourceURLProtocol;
--- a/packages/common/src/jr-resources/JRResourceURL.ts
+++ b/packages/common/src/jr-resources/JRResourceURL.ts
@@ -1,0 +1,46 @@
+const JR_RESOURCE_URL_PROTOCOL = 'jr:';
+type JRResourceURLProtocol = typeof JR_RESOURCE_URL_PROTOCOL;
+
+export type JRResourceURLString = `${JRResourceURLProtocol}${string}`;
+
+interface ValidatedJRResourceURL extends URL {
+	readonly protocol: JRResourceURLProtocol;
+	readonly href: JRResourceURLString;
+}
+
+type ValidateJRResourceURL = (url: URL) => asserts url is ValidatedJRResourceURL;
+
+const validateJRResourceURL: ValidateJRResourceURL = (url) => {
+	if (import.meta.env.DEV) {
+		const { protocol, href } = url;
+
+		if (protocol !== JR_RESOURCE_URL_PROTOCOL || !href.startsWith(JR_RESOURCE_URL_PROTOCOL)) {
+			throw new Error(`Invalid JRResoruceURL: ${url}`);
+		}
+	}
+};
+
+export class JRResourceURL extends URL {
+	static create(category: string, fileName: string): JRResourceURL {
+		return new this(`jr://${category}/${fileName}`);
+	}
+
+	static from(url: JRResourceURLString): JRResourceURL {
+		return new this(url);
+	}
+
+	static isJRResourceReference(reference: string | null): reference is JRResourceURLString {
+		return reference?.startsWith('jr:') ?? false;
+	}
+
+	declare readonly protocol: JRResourceURLProtocol;
+	declare readonly href: JRResourceURLString;
+
+	private constructor(url: JRResourceURL);
+	private constructor(url: JRResourceURLString);
+	private constructor(url: URL | string) {
+		super(url);
+
+		validateJRResourceURL(this);
+	}
+}
--- a/packages/scenario/test/secondary-instances.test.ts
+++ b/packages/scenario/test/secondary-instances.test.ts
@@ -1,3 +1,4 @@
+import { xformAttachmentFixturesByDirectory } from '@getodk/common/fixtures/xform-attachments.ts';
 import { JRResourceService } from '@getodk/common/jr-resources/JRResourceService.ts';
 import {
 	bind,
@@ -552,11 +553,6 @@ describe('Secondary instances', () => {
 				 * Potentially test elsewhere and/or as integration test.
 				 */
 				it.todo('itemsFromExternalSecondaryCSVInstance_ShouldBeAvailableToXPathParser');
-
-				/**
-				 * **PORTING NOTES** (speculative addition)
-				 */
-				it.todo('can select itemset values');
 			});
 		});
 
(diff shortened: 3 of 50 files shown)


lognaturel:
Yay!

The tests you added following our conversation look great. I haven't gotten to adding a test for JR around behavior when CSV tokens have whitespace around them. The behavior at https://github.com/getodk/web-forms/pull/259/commits/6b1f1454749c6ab5226d3ba6654d15f254639645#diff-0ae498d66cc2114687268adaea646f23d17985f3209cc8596693a1e55251eeffR918 looks like it matches the JR behavior and I'm not inclined to make any changes to it currently. For posterity -- Enketo strips those spaces.

I'd be inclined to have an `option` or `config` prop for the client <> host interface for optional things like the missing resource behavior configuration. Is this kind of thing good to consider now? We could certainly also refine it later.

I am surprised by `MissingResourceBehavior` in general. I expected to see a boolean related only to external secondary instances. Is this intended to apply to other media attachments? Any other kinds of resources? I think we've talked about how external secondary instances are typically more critical than image/audio/video attachments. Someone may want to require external secondary instances to be there but accept that some other media could be missing. Is that maybe why you have the flexibility of a non-boolean type? But then `BLANK` feels like it doesn't leave room for the variations I've mentioned.

eyelidlessness:
> Yay!
> 
> The tests you added following our conversation look great. I haven't gotten to adding a test for JR around behavior when CSV tokens have whitespace around them. The behavior at [6b1f145#diff-0ae498d66cc2114687268adaea646f23d17985f3209cc8596693a1e55251eeffR918](https://github.com/getodk/web-forms/commit/6b1f1454749c6ab5226d3ba6654d15f254639645#diff-0ae498d66cc2114687268adaea646f23d17985f3209cc8596693a1e55251eeffR918) looks like it matches the JR behavior and I'm not inclined to make any changes to it currently.

🎉

> For posterity -- Enketo strips those spaces.

It might be interesting to track these specific divergences as they come up. If nothing else it could be valuable documentation. It could also theoretically be part of some kind of config-based way to trigger more Enketo-like behaviors, at least for behaviors users express interest in.

> I'd be inclined to have an `option` or `config` prop for the client <> host interface for optional things like the missing resource behavior configuration. Is this kind of thing good to consider now? We could certainly also refine it later.

Unless I'm misunderstanding, I think that's covered by [this prop](https://github.com/getodk/web-forms/pull/259/commits/6f107bc32897c9819ab64390ed2349b875120f04#diff-c3ea0a2c335137f1e0198eb38dad0bccc5b9685875f0e3fb0ee8edd6711d5362R18). That's a prop on the `OdkWebForm` component (currently the main host <> UI entrypoint). The intent is that host integrations would conditionally specify `'BLANK'` where appropriate. Here is an [example](https://github.com/getodk/web-forms/pull/259/commits/6f107bc32897c9819ab64390ed2349b875120f04#diff-0a6c4aed34584585d8e0d7791c32412d79a771bb028f384467e880bc582fff59R62) where it's in use (effective for both Web Forms Preview and dev mode). I'd expect Central will likely use it for preview mode.

I suppose it'd make sense to update the `@getodk/web-forms` README to mention that prop. I can do that if that would help make this clearer. It didn't really occur to me that we're using that as API documentation. I tend to think of the type definitions as the source of truth for stuff like this at package boundaries, absent a more dedicated documentation site.

On that note: insofar as the `OdkWebForm` Vue component and its props constitute our official host <> UI interface, we might want to consider using a documentation generator instead of the README. That way we can be sure it's up to date, and probably benefit from other details commonly solved by such tooling. I'm not sure whether the tool used for the engine API is suitable, but I'd be happy to file an issue about it if there's interest.

> I am surprised by `MissingResourceBehavior` in general. I expected to see a boolean related only to external secondary instances. Is this intended to apply to other media attachments? Any other kinds of resources? I think we've talked about how external secondary instances are typically more critical than image/audio/video attachments. Someone may want to require external secondary instances to be there but accept that some other media could be missing. Is that maybe why you have the flexibility of a non-boolean type? But then `BLANK` feels like it doesn't leave room for the variations I've mentioned.

Yeah, this gets at the intent of making it an enum. The `fetchFormAttachment` signature is designed to be general enough for loading all form attachments, but the branchy logic isn't likely to be binary when we add support for media. If we made it a boolean flag now, we'd proably end up with at least one additional boolean flag (perhaps more?). As it stands, I hope the enum values are reasonably self-explanatory (and if they're not that, I'd like to revisit!), but it's definitely intended to support additional values.

I also hope that `BLANK` _as a concept_ is general enough that we can expand on the existing application of that concept to produce valid-but-void representations for any other missing resource. That way the engine can provide some meaningful/usable non-`null` value anywhere a client would expect to find and present media. I haven't thought too far ahead on what that `BLANK` representation might be for a given media type—it may just be stub `blob:`/`data:` URLs appropriate for the expected MIME type, maybe something more structured depending on where that design goes. But I think it'll be valuable to have something non-`null` in those cases (at the engine/client boundary) once we get there.

lognaturel:
> It might be interesting to track these specific divergences as they come up

Agreed. I try to at least have them in the differences doc we have and added some details at 
https://docs.google.com/document/d/1hmHrHJYsbovzli2aLHLvdt6Tq9reibnC4PylunKI58I/edit?tab=t.0#heading=h.dima7n868glp

> Unless I'm misunderstanding, I think that's covered by [this prop](https://github.com/getodk/web-forms/pull/259/commits/6f107bc32897c9819ab64390ed2349b875120f04#diff-c3ea0a2c335137f1e0198eb38dad0bccc5b9685875f0e3fb0ee8edd6711d5362R18).

I wasn't clear enough! I mean instead of having that at the top level, introduce a prop with name `config` or `options` that would be an object with a defined type for all of the different optional configurations that we'll eventually have. I think there will be many and that it will be nice from a user standpoint to see a difference between things that are critical to think about and others that are less so.

> Yeah, this gets at the intent of making it an enum.

So instead of a couple of booleans we'd end up with the corresponding cross product, basically. Something like `BLANK`, `ERROR`, `EXTERNAL_SECONDARY_ERROR__OTHERS_BLANK`, `EXTERNAL_SECONDARY_BLANK__OTHERS_ERROR` (maybe not needed) And then we'd have room for other nuances we may eventually learn about.

Did you consider making this part of what the host returns on a per-resource basis instead of a blanket config ("here's how you get the resource and what you should do if you find you can't")? I imagine Central would just use the resource type to decide so what you have now is sufficient but other hosts could do things like state that a specific video is critical to the form but others aren't.

packages/xforms-engine/src/client/EngineConfig.ts
```diff
@@ -55,29 +63,59 @@ export interface EngineConfig {
...
-	 * from sources other than the network, (or even in a test client to provide
-	 * e.g. resources from test fixtures).
+	 * - For calls to {@link initializeForm} with a URL, if this configuration is
+	 *   not specified it will default to the global {@link fetch} function (if
+	 *   one is defined).
+	 */
+	readonly fetchFormDefinition?: FetchResource;
+
+	/**
+	 * @deprecated
```
lognaturel, line 47: I don't think we need to mark things as deprecated and keep them around yet. I think it's simpler to immediately get rid of them.

packages/scenario/test/secondary-instances.test.ts
```diff
@@ -703,4 +705,86 @@ describe('Secondary instances', () => {
...
+			});
+
+			scenario.answer('/data/first', 'a');
+
+			expect(scenario.answerOf('/data/first')).toEqualAnswer(stringAnswer('a'));
+		});
+
+		// This test is redundant to the one above, but demonstrates how we could
+		// define form attachments inline, like we do form definitions—a pattern
+		// worth considering if we want to expand external secondary instance
```
lognaturel, line 853: The philosophy in JR has been to test the basic functionality that is specific to external data attachments (especially their parsing) and then to favor tests on internal data instances for functionality that applies to all secondary instances.

packages/scenario/test/secondary-instances.test.ts
```diff
@@ -823,4 +825,234 @@ describe('Secondary instances', () => {
...
+				rowDelimiter: '\r\n',
+			},
+			{
+				description: 'unsupported row delimiter: LFLF',
+				rowDelimiter: `\n\n`,
+				expectedFailure: 'parse',
+			},
+
+			{
+				description: 'somewhat surprisingly supported row delimiter: LFCR',
```
lognaturel, line 974: 🤷‍♀️

lognaturel (approved):
Couple of small suggestions but this is ready to merge as far as I'm concerned.

I don't love the full `jr:` URL being passed to clients because I'm quite confident all will strip the `jr://type`. But we can always iterate!

packages/common/src/jr-resources/JRResourceURL.ts
```diff
@@ -0,0 +1,46 @@
+const JR_RESOURCE_URL_PROTOCOL = 'jr:';
```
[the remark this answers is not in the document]
lognaturel: @eyelidlessness let me know that this matches https://developer.mozilla.org/en-US/docs/Web/API/URL/protocol

packages/common/src/fixtures/xform-attachments.ts
```diff
@@ -0,0 +1,137 @@
...
+
+	return fileName;
+};
+
+// prettier-ignore
+const xformAttachmentFileExtensions = [
+	'.csv',
+	'.geojson',
+	'.xml',
+	'.xml.example',
```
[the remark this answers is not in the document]
lognaturel: There's an .xml.example from the JR tests. .xlsx might give more specific test errors if in the future we work towards supporting XLSX secondary instances.

packages/common/src/jr-resources/JRResourceURL.ts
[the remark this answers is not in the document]
eyelidlessness: I also wanted to add this other breadcrumb for future reference: there's also potential for confusion with [this bit](https://github.com/getodk/web-forms/pull/259/files#diff-d921e2a60e252e84fb924e5cbd8f80032af4c0cae4889037376d78cdd2bdaf38R29) of naming ported from JavaRosa, where `schemes` is a reference to a set of _directory paths_ in a `jr:` URL.

packages/xforms-engine/src/parse/model/SecondaryInstance/sources/ExternalSecondaryInstanceResource.ts
```diff
@@ -0,0 +1,222 @@
...
+const assertResponseSuccess = (resourceURL: JRResourceURL, response: FetchResourceResponse) => {
+	const { ok = true, status = 200 } = response;
+
+	if (!ok || status !== 200) {
+		throw new ErrorProductionDesignPendingError(`Failed to load ${resourceURL.href}`);
+	}
+};
+
+const stripContentTypeCharset = (contentType: string): string => {
+	return contentType.replace(/;charset=.*$/, '');
```
sadiqkhoja, line 17: What do you think about this:

```diff
- return contentType.replace(/;charset=.*$/, '');
+ return contentType.replace(/;\s?charset=.*$/, '');
```

Sometimes response contenttype includes space between media type and charset, including examples here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type#examples

eyelidlessness: Can you file an issue? This PR has already been merged, so I can't make changes here.

PR #261, opened 2024-12-10
Numeric controls: decimal, int, appearance=numbers

### LINKED ISSUE #251: Support integer inputs
Collect docs: https://docs.getodk.org/form-question-types/#integer-widget

Enketo has an increment/decrement control which can also be controlled using up and down arrows:
Screenshot 2024-11-13 at 1 53 39 PM

Appearances:
* [thousands-sep](https://docs.getodk.org/form-question-types/#thousands-sep)
* [counter](https://github.com/getodk/docs/pull/1849/files)

### LINKED ISSUE #252: Support decimal inputs
Collect docs: https://docs.getodk.org/form-question-types/#decimal-widget

The decimal marker must be localized for display but the underlying value should always have a period decimal marker.

Enketo has an increment/decrement control which goes up or down to the nearest integer.

Appearances: [`thousands-sep`](https://docs.getodk.org/form-question-types/#thousands-sep)

### LINKED ISSUE #260: Support string type with number appearance
https://docs.getodk.org/form-question-types/#number-text-widget

eyelidlessness:
Closes #251.
Closes #252.
Closes #260.

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS) **See "Notable limitations/quirky behaviors"**
- [x] Safari (iOS) **See "Notable limitations/quirky behaviors"**
- ~~[ ] Not applicable~~

Expanded test coverage, largely around bind types. Lots of manual testing of weird edge cases and nuances in the UI.

I considered handling the UI and engine portions of this separately. I think that would have risked splitting responsibilities for integrity of per-`type` values in ways that could bite us longer term. This also sets us up for much more rapid development of the remaining data types (which that shorter term approach would not have done).

I think the greatest areas of risk are anywhere that new branchy logic should have more nuance than I've accounted for. But I expect if that's the case, those nuances are lurking in other pending feature work.

A new fixture is added, which I used for testing in the UI.
	
### Engine: `<bind type>` parsing

- `<bind type>` is now parsed into a `BindTypeDefinition` with:
    - `resolved`: the effective type determined for the node's binding
    - `source`: the literal value of the `type` attribute, if any

    It may not be valuable to preserve `source` longer term, but it was useful during development, especially around some existing or newly introduced nuances...

- Namespaced types specified in the `http://www.w3.org/2001/XMLSchema` namespace are resolved to their LocalPart. Resolution applies the following logic (applicable for namespace declarations on each respective `<bind>` element, or any of its ancestors):

    - _If the XML Schema namespace URI is associated with a namespace declaration: normal XML namespace resolution applies. I believe this is completely consistent with all applicable specs. By convention, the namespace is typically declared for the `xsd` prefix.

    - _Iff the XML Schema namespace URI is **NOT** associated with any namespace declaration_: The `xsd` prefix is used as a default. This exceeds spec. I went with this because there are some form fixtures from JavaRosa which would be affected, and at least at a glance this appears to be the intent/expected behavior for those fixtures.

    - _If any other URI is declared for the `xsd` prefix_: the `xsd` default is not applied.

- `integer` is resolved as an alias to `int`. We may decide to revert that, depending on [this corresponding JavaRosa change](https://github.com/getodk/javarosa/pull/810). I left it in since it was in flight, and would allow us to discuss both in tandem.

    **My take:** If we are confident that we're the only ones using the alias, removing it in both projects makes sense! If we believe users may be using it, I'd like to make sure we have a plan for communicating the breaking change to them.

- Other found aliases are _implicitly resolved_ to `string` (implicitly because it is the default `type`). Those I've found in existing fixtures:

    - `select1`
    - `odk:rank`
    - `rank` (unprefixed)

### Engine/client interface: `ModelValueNode` is generic over additive `valueType` property

**Note:** This change isn't strictly required for the features in this PR. I chose to do this as a testbed for the approach, which I then expanded to input controls. I used this testbed to evaluate both the notion of generic value types overall, and the impact on **client usability**.

 This introduces the following (technically breaking) changes to `ModelValueNode`:

- A `valueType` property is added, associated with a new client-exported union type `ValueType` (effectively the same union of `<bind type>`s).

- The `ModelValueNode` interface has a `ValueType` _type parameter_ ("generic"), i.e. `ModelValueNode<V extends ValueType>`

- Each `ValueType` is associated with a representation of its _runtime value_, and the type for that runtime value is then associated with `ModelValueNode<V>.currentState.value`. In this PR, those representations are as follows:

    - `ModelValueNode<'string'>.currentState.value: string`
    - `ModelValueNode<'int'>.currentState.value: bigint | null`
    - `ModelValueNode<'decimal'>.currentState.value: number | null`
    - `ModelValueNode<AnyOtherValueType>.currentState.value`: `string` (as a default, until we immplement those value types)

- The union of node types we currently call `GeneralChildNode` (which a client iterates over for the children of a corresponding `GeneralParentNode`) is expanded to include the union of all `ModelValueNode<V>`s. An alias to _that_ sub-union is also exported as `AnyModelValueNode` (which is generally how clients will consume these now).

- Each `ModelValueNode` also exposes its current value as a string, regardless of its `valueType`:

    - `AnyModelValueNode.currentState.instanceValue: string`

    This is mostly an affordance for `scenario`, which has many tests exercising the values of model-only nodes. The same state property will be exposed for any other "value node" as we expand the approach. In terms of spec semantics, this is effectively the node's "XPath value", and its _pre-serialized_ "submission value" (i.e. it is **NOT** escaped for XML).

### Engine/client interface: ~~`StringNode`~~ -> `InputNode<V extends ValueType>`

The `StringNode` client interface has now been renamed `InputNode` to reflect the fact that it is no longer a string-only node, and to better reflect the fact that it is strictly associated with `<input>` controls in the form definition.

All of the notes for `ModelValueNode` apply. Additionally, the `setValue` method's value parameter is now also generic over `valueType`:

- `InputNode<'string'>.setValue: (value: string) => RootNode`
- `InputNode<'int'>.setValue: (value: bigint | number | string | null) => RootNode`
- `InputNode<'decimal'>.setValue: (value: bigint | number | string | null) => RootNode`

**Note:** both of the numeric types accept a superset of their runtime value type for writes. This is entirely an accommodation for ergonomics in client integrations (which paid off a bit!).

As with `ModelValueNode`, the engine exports all of the `InputNode<V>` sub-union as part of the `GeneralChildNode` union. And the engine exports a catch all sub-union as `AnyInputNode`.

Unlike `ModelValueNode` (where clients are generally not doing more than accessing the nodes values, if that), it also made sense to provide named representations of the newly introduced input nodes and their value types. So the engine also now exports:

- `StringInputNode`
- `IntInputNode`
- `DecimalInputNode`

Lastly, the engine exports another catch all for the remaining value types which _fall back to a string value representation_ pending implementation of those value types. This is exported as `TemporaryStringValueInputNode`, hopefully providing some clarity that this is not a stable type we expect to keep around forever.

### Engine-internal: `ValueNode<V extends ValueType>` and its subclasses

Internally, the engine now has a new abstract `ValueNode` base class, which handles _nearly all_ of the responsibilities of each affected value node type. The following have been updated to be subclasses of it:

- `ModelValue` (implementation of `ModelValueNode`)
- ~~`StringField`~~ -> `InputControl` (implementation of `InputNode`)

I anticipate moving nearly all of the other value node implementations to be subclasses of `ValueNode`, with the possible exception of selects, and the pending implementation of `odk:rank`. (Currently the generalization does not account for select/rank options.)

This base class largely takes over responsibilities for existing implementation details _common to value nodes_:

- Initializing the node's value state _as a primary instance value_:

    - Creation of a reactive signal to store the value
    - Deriving the initial value (currently from the form definition; from edited submission still anticipated/pending)
    - Ensuring non-relevant values are treated as blank
    - Initializing reactive `calculate` computations

- Mapping that primary instance value to/from a _runtime representation_ (more on this in the next section on `ValueCodec`)

- Integration with XPath (most pertinent now being reactive access to its primary instance state)

- Initialization of node validation (`constraint`, `required`)

### Engine-internal value representation: `ValueCodec<V extends ValueType>`

Each `ValueNode` expects a new engine-internal implementation of a `ValueCodec` which is:

- Implemented for that node's `valueType`
- Implementing `decodeValue`, to decode the node's primary instance value state into its runtime value representation
- Implementing `encodeValue`, to encode the node's runtime _input_ value representation to its primary instance value

This basically breaks apart the same responsibilities from an earlier engine-internal interface (`ValueContext`). It's what allows value nodes to be generic over their value type _independent of_ other value node responsibilities.

This abstraction is also designed so that different **node types** may have different runtime representations of the same **value types**. The intent here is that we can use the same `ValueNode` and `ValueCodec` to serve node type-specific nuances. A couple examples which are top of mind:

- `Note` (implementation of `NoteNode`): it will probably best serve clients if a note's value representation is _always_ `string | null` regardless of its `valueType`. (And at least for some `valueType`s, that `string` representation will probably differ from its `instanceValue`!)

- `TriggerControl` (implementation of `TriggerNode`): its runtime value representation will likely continue to be `boolean`, but its **encoded primary instance value** will be serialized differently from a `'boolean'` model/input node: `'' | 'OK'` (versus `'0' | '1'`).

### Numeric input UI

The existing `InputText` component has now been broken up into:

- `InputControl`: a wrapper for typed inputs, handling responsibilities common to all of them (e.g. label, hint, required status, validation)

- `InputNumeric`: a composable base component for each numeric input type, handling various nuances of integration with PrimeVue's `InputNumber`

- `InputDecimal` composing `InputNumeric` with props specific to the `decimal` value type

- `InputInt` composing `InputNumeric` with props specific to the `int` value type. This includes `min`/`max` bounds which _differ from Collect_ (but match what I understand to be the **intent** of collect's bounds)

- `InputNumbersAppearance` composing `InputNumeric` (for `type="string"` if specified + `<input appearance="numbers">). The behavior is roughly a superset of `InputDecimal` and `InputInt`: decimal values are allowed, and the `min`/`max` bounds are relaxed.

- `InputText`: just the control portion, as extracted from the component of the same name on `main`

For all numeric inputs, the `thousands-sep` appearance determines whether to use "grouping" on the PrimeVue base component. I assume there will be some localization-specific nuances to address as well. For now I deferred to PrimeVue's defaults for that.

#### Notable limitations/quirky behaviors

I've made a reasonable effort toward _reasonable behavior_ for these inputs. There are some caveats worth noting.

First, some which are browser-specific:

- **Safari, iOS:** cannot enter negative ~~or decimal (fixed)~~ values, apparently this is [optional](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode). ~~I believe this can be addressed by setting `inputmode="text" type="number"` and then adding type-specific `pattern` attributes. The downside is that a bunch of non-numeric keys are present on this keyboard variant (for reasons I still don't understand). I'd like to make this change, but would be happy to put it off to discuss nuances/look for better alternatives.~~ Just kidding! This breaks the increment/decrement buttons (causing a runtime error on click/tap), also for reasons I don't understand! As far as I can tell, the best we can do with PrimeVue's `InputNumber` component is `inputmode="text"`, which will use the default software keyboard. Otherwise, negative number input on iOS is not possible.

- **Safari, macOS:** numeric inputs swallow **all keyboard commands** (at least anything with cmd key that I tested)

- **All tested desktop browsers:** numeric inputs swallow **cut keyboard commands** (i.e. cmd x); cut by other means (e.g. menu, touch popup) does still work

The PrimeVue component also has a variety of _general_ quirks, which I've tried to minimize as much as possible. Worth noting here:

- By default, the DOM and reactive state can easily get out of sync. This is true for any keyboard input that doesn't trigger increment/decrement behavior (i.e. typing a value!). Unlike PrimeVue's base `InputText` (which **is** the base component for its `InputNumber`!), state changes are only applied on `change` (which is only dispatched once focus leaves the control). This is exacerbated by enforcing bounds (either with the `min`/`max` props, or the engine's slightly more permissive bounds).

    This is the reason for the very weird (IMO) use of `customRef`. By using state and focusing entirely on numeric values to address these issues, we avoid having to deal with parsing and re-formatting the input's DOM text value, which could add quite a bit of complexity! One important downside to this worth noting is that every time we override the value state it causes the user's cursor to jump to the end of the input. In my testing this is _usually close enough to what I'd expect_, but it might be jarring in some cases.

- The base presentation in the theme we currently use is terrible. Here I've opted for something more subtle/closer to a common native control.

### Setting the stage for pending feature work

The design of engine internals in this PR is positioned to expedite implementation of the remaining data types, and to allow most of that work to focus on those types' UI/UX concerns. In particular, the `ValueNode` and `ValueCodec` abstractions are designed so that we can rapidly add support for encoding/decoding each data type, without making further substantive changes to any other aspect of the affected nodes (apart from assigning them named exports as appropriate).


packages/scenario/test/bind-types.test.ts
```diff
@@ -0,0 +1,647 @@
...
+						body(
+							repeat('/data/repeat-count', '/data/count-calc')))
+				);
+
+				expect(scenario.answerOf('/data/count-calc')).toEqualAnswer(intAnswer(5));
+				expect(scenario.countRepeatInstancesOf('/data/repeat-count')).toBe(5);
+			});
+		});
+		it('casts a calculated value to an int', async () => {
+			// const scenario = await Scenario.init(
```
lognaturel, line 627: Why is this one commented out?

packages/web-forms/src/components/controls/Input/InputNumeric.vue
```diff
@@ -0,0 +1,211 @@
...
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode}
+ */
+type NumericInputMode = 'decimal' | 'numeric';
+
+let inputMode: NumericInputMode;
+
+if (props.isDecimal) {
+	fractionalDigits = {
+		min: 0,
+		max: 18,
```
lognaturel, line 50: I see this is minimum conformance defined by XML. Collect caps at 15 total characters (including the decimal). Let's track this difference for possible follow-up work.

lognaturel (approved):
In Firefox 132.0.2 (aarch64), macOS, the string with numbers appearance turns all characters after the 17th digit into 0s.

I made an inline node about the bounds for decimal. For ints, Collect limits the entry to 9 characters, including the -.

I think all of these could be addressed in follow-up work.

packages/scenario/test/bind-types.test.ts
eyelidlessness: I think it should've been deleted! 🙃 It was one test which became the basis for the three above.

eyelidlessness:
> In Firefox 132.0.2 (aarch64), macOS, the string with numbers appearance turns all characters after the 17th digit into 0s.

I'm pretty sure this is because the component's value is represented as a number. A 17-digit integer exceeds `Number.MAX_SAFE_INTEGER`.

PR #265, opened 2024-12-16
Fix: Web Forms Preview demo form links 

eyelidlessness:

- [x] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [x] Not applicable

Added an e2e test which loads all of the Web Forms Preview form links, from the build product. That should be much more reliable than any manual testing I might do.

It would be preferable if specific fixtures could be referenced by something stable, and not rely on anything dynamic at build time.

It's possible some other nuance of glob loading is affected by this. But I would expect to see some sign of that either in `scenario` or dev mode, both of which still seem to be working as expected.

The links on the preview page should suffice.

## What's changed

- Restored use of build-time asset URLs in links to XForm fixtures
- Added an e2e test to ensure that doesn't break again, along with a couple of basic Vue template supports for (hopefully) clearer intent in e2e tests going forward

### DIFF
--- a/packages/common/src/fixtures/import-glob-helper.ts
+++ b/packages/common/src/fixtures/import-glob-helper.ts
@@ -34,11 +34,16 @@ if (IS_NODE_RUNTIME) {
 
 type ImportMetaGlobURLRecord = Readonly<Record<string, string>>;
 
-export type ImportMetaGlobLoader = (this: void) => Promise<string>;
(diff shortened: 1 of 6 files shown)


eyelidlessness:
I don't think this needs a changeset? It should hopefully only affect the Web Forms Preview secondary build product.

lognaturel (approved):

PR #266, opened 2024-12-17
Post-release (@getodk/web-forms@0.6.0) dependency updates & related chores

eyelidlessness:
I think the pertinent details are captured in individual commit notes. Top-level, the most notable details:

- Removes the `@getodk/ui-solid` package, as we had planned
- The updates to `tree-sitter` etc allowed a vast simplification of the `@getodk/tree-sitter-xpath` package structure, build, etc (as well as nicely paring down top-level git files)

One thing **not mentioned** in any of the commit notes: we are not currently running the `@getodk/web-forms` E2E tests in CI! I finally spent a few minutes checking on their status locally, and they're... well, they're broken. I have (and can include in this PR):

- A complete fix for `style.test.ts` (it just needs to look for a different text snippet to proceed)
- A partial fix for `vue.test.ts`, but it times out. I'm not sure what the best thing is for this test, but it would probably be best if we can break it up into per-fixture tests (~~probably requiring codegen~~ possibly solvable with Playwright's [concept of fixtures](https://playwright.dev/docs/test-fixtures)).


eyelidlessness:
Looks like I have some new lint issues to address 🙃

lognaturel (approved):
I think it's ok to get the E2E test running in CI in its own PR but don't feel strongly either way.

PR #275, opened 2025-01-15
Add reference to Volta

lognaturel:
Otherwise someone has to open the yarn file to see that Volta is used


latin-panda (approved):
Thank you!

PR #277, opened 2025-01-16
Revert select value type support

eyelidlessness:

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [x] Safari (iOS)

Since this rolls back functionality in #273, the tests added for that functionality have been removed. I think everything else is covered by existing tests and/or type checks.

This leaves most of the other substantive improvements from #273 intact. Any considerations there apply here.

This is technically a regression! As of #273, Web Forms behavior for `<select1>` with a non-string `<bind type>` was consistent with Collect/JavaRosa (for the current set of supported bind/value types). But we've decided that is an unsupported use case for now, as it isn't possible to produce an affected form with XLSForm/pyxform.

The fixture added in #273 now produces an error on load (which is now expected). The fixture added in this commit demonstrates correct support for spaces in `<select1>` values (which could have been added in that PR as well).

## What's changed

The primary purpose of this change is to revert support for non-`string` `<bind type>`s associated with a `<select>` or `<select1>` control (introduced in #273). It also establishes the same precedent for `<odk:rank>`. The bulk of this aspect of the change involves:

- Eliminating any `ValueType` type parameterization associated with `SelectNode` and its implementation
- Eliminating the vast majority of implementation code supporting that type parameterization

Some of the breadcrumbs for that functionality are left intact, in `ValueArrayCodec`. There are other use cases where we'll benefit from being able to derive encoding/decoding of array items (e.g. "select from map", which was part of the motivation for typed select support in the first place).

I also took this PR as an opportunity to address a gap in the preparation for `<odk:rank>`, which I had neglected to address in #273. Specifically, [the PR's first commit](https://github.com/getodk/web-forms/pull/277/commits/b38cb2c1a8fd58fd3f5399d0cda9619c52218126) makes some structural changes to prepare for parsing `<odk:rank><item>` or `<odk:rank><itemset>`.

Lastly, since I was already looking in that area, I went ahead and took care of these maintenance chores:

- The `SelectDefinition` class associated with the _body element_ is now named `SelectControlDefinition`, consistent with other parsed representations of similar controls. This eliminates a naming conflict where we also have an _interface_ named `SelectDefinition` representing the `NodeDefinition` associated with that control. (We really should take a more holistic look at naming, especially in these earlier areas of the engine package. For now, this is a piecemeal change aiming for consistency with current thinking.)
- `ItemsetNodesetContext`, which was unused, is removed.


packages/xforms-engine/src/instance/hierarchy.ts
latin-panda, line 1: This is not my review yet, but I just noticed that we need to remove range from [note-types.ts](https://github.com/getodk/web-forms/blob/e6d01b01a40ab1aca26b561996658509b78be071/packages/xforms-engine/src/client/node-types.ts#L3). Could you include that little fix here? :)

packages/xforms-engine/src/parse/body/BodyDefinition.ts
```diff
@@ -48,7 +48,7 @@ const BodyElementDefinitionConstructors = [
 	PresentationGroupDefinition,
 	StructuralGroupDefinition,
 	InputControlDefinition,
-	SelectDefinition,
+	SelectControlDefinition,
```
lognaturel, line 51: Maybe this will expand to something like `ItemBasedControlDefinition` or something with the addition of `rank`?

lognaturel (commented):
Looks like what I expect from a pretty quick pass through!

packages/xforms-engine/src/instance/hierarchy.ts
eyelidlessness: Phew, good catch! And fixed.

It kind of bugs me that nothing automatic caught it. I guess it's probably not a huge deal since we'll support these soon enough.

packages/xforms-engine/src/parse/body/BodyDefinition.ts
eyelidlessness: I can see that becoming an abstract base class for select/select1/rank. Both concrete classes would still be enumerated here though, as it's used for instantiation during parsing. And the base class would make even more sense when we split select/select1.

packages/web-forms/src/lib/format/selectOptionId.ts
latin-panda, line 1: Suggestion:
I understand the intent behind prioritizing reusability, but I'm considering ways to reduce the spread of what defines "select" to make it more self-contained as a component. Additionally, I suggest inverting the responsibility for the ID format.

Assuming the ID format is tied to the select functionality in the engine, what about moving the `selectOptionId(...)` logic into the `Select1Component`? Then, you could pass the formatter as an Input to the `RadioButton` and `CheckboxWidget` components.

This approach shifts the responsibility for defining the ID to the component that uses `RadioButton` and `CheckboxWidget`, which would provide greater flexibility. For example, if these components are later used in contexts requiring a different ID format, the adjustment would be simpler and more modular.

packages/xforms-engine/src/client/hierarchy.ts
latin-panda, line 1: Question:
Is the hierarchy defined based on the XML output?

For instance, are `group` and `repeat` considered higher-order components, while `select` functions as a child or leaf in the tree? Additionally, should all new components be declared in this file first to ensure they are correctly mapped in the typing?

It seems like a good place to start for a new contributor to understanding the engine.

packages/xforms-engine/src/error/SelectValueTypeError.ts
```diff
@@ -0,0 +1,22 @@
...
+ * {@link https://vuejs.org/api/sfc-script-setup#defineoptions | Vue component option}).
+ *
+ * @see {@link https://github.com/getodk/web-forms/issues/276}
+ */
+export class SelectValueTypeError extends XFormsSpecViolationError {
+	constructor(definition: SelectDefinition<UnsupportedSelectValueType>) {
+		const { valueType } = definition;
+
+		super(
+			`Selects of type ${valueType} are not currently supported. If this functionality would be useful for your form, your feedback is welcome!`
```
latin-panda, line 19: Suggestion:

This seems like a common type of error for other components. What do you think about using generics in the declaration? Then, in the select definition (SelectNode? or SelectControl?), we could define the error type, extend it, and pass the types through the generic.

By centralizing the possible errors for `select` within select definition (SelectNode? or SelectControl?), it would make them easier to discover and help provide a clearer picture of the entire select functionality.

packages/xforms-engine/src/error/XFormsSpecViolationError.ts
```diff
@@ -0,0 +1 @@
+export abstract class XFormsSpecViolationError extends Error {}
```
latin-panda, line 1: Question:

I like reusable errors in this folder! 
What is this class for? What will it do?

packages/xforms-engine/src/instance/SelectControl.ts
latin-panda, line 1: Suggestion: <out of scope of this PR>

I'm not suggesting this as a significant change, but rather as an improvement to help new contributors and streamline the process. Grouping by entity, like this, could make it easier for newcomers to learn, discover, and quickly add new components.

/xforms-engine
 |---/src
 |---index.ts
 |---hierarchy.ts

 |---|---/errors
 |---|---|---< common-thing >Error.ts
 
|---|---/controls 
 |---|---|---/select
 |---|---|---|---SelectNode.ts
 |---|---|---|---SelectState.ts
 |---|---|---|---SelectControl.ts
 |---|---|---|---SelectCodec.ts
 |---|---|---|---Select< thing >.ts

packages/xforms-engine/src/instance/SelectControl.ts
```diff
@@ -35,62 +34,72 @@ export type AnySelectDefinition = {
 	[V in ValueType]: SelectDefinition<V>;
 }[ValueType];
 
-interface SelectControlStateSpec<V extends ValueType> extends ValueNodeStateSpec<SelectValues<V>> {
+type AssertSupportedSelectValueType = (
```
latin-panda, line 37: Question:

Is SelectNode the type definition and SelectControl the code to implement the functionality?

packages/xforms-engine/src/lib/codecs/select/MultipleValueSelectCodec.ts
latin-panda, line 1: Question:

1. When codec run? I'm looking to know how to verify the output is what we expect
2. I wonder if the Rank codec will be the same as this one and if we can reuse or duplicate it. 
    2.1 To reduce dependency between components, I would make 1 independent codec for Rank, but I want to know based on your experience with xforms

packages/xforms-engine/src/lib/dom/query.ts
```diff
@@ -3,7 +3,7 @@ import type {
 	KnownAttributeLocalNamedElement,
 	LocalNamedElement,
 } from '@getodk/common/types/dom.ts';
-import type { SelectElement } from '../../parse/body/control/select/SelectDefinition';
+import type { SelectElement } from '../../parse/body/control/SelectControlDefinition.ts';
```
latin-panda, line 6: Question:

Why do we have one for Select, while not for Range, Note, Inputs? 
Looking to understand the reason so I can see if I need it for Rank or Geopoint

latin-panda (changes requested):
Thanks for sending this so fast!

I made a couple of suggestions, but it's mainly questions.

Let me know what you think.

packages/scenario/src/jr/select/SelectChoice.ts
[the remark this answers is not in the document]
eyelidlessness: > 1. When it says ... implements a client..., is it a client to perform tests based on JavaRosa?

Yep! We decided to make this boundary explicit when I started porting JavaRosa tests, as an additional exercise of the engine/client boundary (alongside the `@getodk/web-forms` UI client).

> 1. [...] there is already a [client directory](https://github.com/getodk/web-forms/tree/main/packages/xforms-engine/src/client), and I was wondering about the difference.

Yeah, that directory represents the engine's "client interface". Effectively the full interface of the package boundary. It's not quite as isolated as it could be, since in some cases it references types from other parts of the package. We could improve that. (On the other hand, we may [reconsider that structure](https://github.com/getodk/web-forms/pull/277#discussion_r1920533235) entirely!)

> 2. ... and benchmarks against that engine., is it referring to performance tests?

I think we should probably remove that text. I wasn't sure whether we were going to port benchmarks from JavaRosa, and at least in that pass we ultimately decided not to. There's currently just a placeholder that can be run with the benchmark command, but it serves no purpose.

> 2. [...] And, is there a baseline performance for what "good" means?

That's a great question! I don't think we have a consistent answer to this. In general, all of these feel like good candidates for _performance goals_ (not all of which we currently meet, at least not across the board):

- Form loads in 100ms or less on a representative device[^1][^2]
    - Computations from answering any form question complete in 100ms or less[^1][^2]
- At least as fast as Collect/JavaRosa[^3]
- Faster than Enketo[^4]

> 2. [...] If I recall correctly, there was a delay when having groups and repeats simultaneously, which we'd like not to make it slower in Web Forms.

Can you clarify where you've seen that delay? Do you mean in Enketo (as in, "we'd like not to make it slower in Web Forms [than in Enketo]")? Or is this already a performance issue in Web Forms (as in, "we'd like to make it slower [than it already is]")?

In either case, if you know of any specific forms that perform poorly I would love to have access to them (and bring them into the repo, with appropriate permission from form authors)! I frequently check performance of forms with known performance issues (either in WF or Enketo), at least to be sure we're not introducing significant performance regressions.

And if you want to do the same, these are the ones I check most frequently these days:

- `bench9-alt.xml`
- `nigeria_wards_external.xml`

[^1]: There are a few slow forms that I test against which don't meet this goal, but it's generally what I have in mind when I'm focused on performance.

[^2]: At least with our current feature set, there is one very specific area of poor performance which stands out above all else: itemsets, particularly with an expression like `instance('some-id')/root/item[some = /data/predicate]`.

[^3]: I haven't done explicit comparisons here. I suspect (hunch based on areas of perf concern in JR, some of which are fast in WF) that we do pretty well with specific exceptions[^2].

[^4]: WF is faster than Enketo on almost all of the perf-focused fixtures I test against, with specific exceptions[^2].

packages/xforms-engine/src/client/hierarchy.ts
eyelidlessness: The hierarchy is defined based on:

- The kinds of nodes which exist
- The kinds of nodes which _semantically_ can be parents and/or children of one another

It's also worth noting that there are two `hierarchy.ts` modules (I'm sorry!): one for the client node interfaces, and one for the internal implementation of those. They're similar, but slightly different for ... reasons. Open to reconsidering naming on this, but for now we do need to maintain both in tandem.

These are the (client interface) hierarchy semantics:

- `RootNode`, which is the topmost node, with no parent
    - As such: all nodes **except** `RootNode` are children of another node
- `RepeatRangeNode` is a parent, but it **only** parents `RepeatInstanceNode`s
    - Correspondingly: `RepeatInstanceNode`s are **only** ever children of a `RepeatRangeNode`
- `RootNode`, `GroupNode`, `SubtreeNode` and `RepeatInstanceNode` are "general parents": they can be parents of any node **except** `RootNode` and `RepeatInstanceNode`
    - Correspondingly: all nodes **except** `RootNode` and `RepeatInstanceNode` are "general children", i.e. they can have any "general parent"

These reflect the combined semantics of ODK XForms (per spec) and the engine's client interface representing those spec concepts.

Note: the concept of a "repeat range"—a parent node containing a contiguous series of "repeat instances"—is an invention of the WF engine. As with any abstraction, it's a little leaky, but it's worked out pretty well for reasoning about repeats (which gets kind of bonkers without some notion of set/items).

packages/xforms-engine/src/error/SelectValueTypeError.ts
eyelidlessness: This is a great suggestion. How would you feel about deferring it for now? We can pair on design for it as part of the work on `<odk:rank>`, since it'll have the same type limitations associated with a different control/node type?

packages/xforms-engine/src/error/XFormsSpecViolationError.ts
eyelidlessness: It's currently just being used as a base class for `SelectValueTypeError`.

So... some background is warranted here. One of the big weak points in the current design (throughout the Web Forms project) is errors. How they're defined and organized, how they're invoked, how they're documented, and how they're produced to call sites. We want to improve on all of this. We originally planned on a big push in #202, but decided to defer that because the _feature_ already had a huge scope.

In general, I've been trying to have a light touch on errors since I know we'll be focusing on them more holistically. My intent here was to capture these ideas:

1. there are many XForms spec violations we might want to report in a consistent way
2. we don't know what they all are, or what common things we want to report about them

So this base class is basically just a placeholder for that with a name.

packages/xforms-engine/src/instance/SelectControl.ts
eyelidlessness: I **really** like this idea, and I've filed #282 to track it. In reality, I think we'll find it's actually a pretty big task, in terms of modeling, _naming_, and careful consideration of how we apply the concept to internals. I'm not sure whether it _could work_ given the domain concepts and how we model them. But it's a technique for code organization I really like where I've seen it applied (and one we definitely should discuss for the UI package!).

packages/xforms-engine/src/instance/SelectControl.ts
eyelidlessness: Pretty close! `SelectNode` is the _client interface_, and `SelectControl` is the engine-internal implementation of it. Both are _also treated as types_, but only the former is exposed outside of the engine package.

Basically we have a "client interface" for each type of node. They intentionally hide a lot of details that are necessary for implementing the ODK XForms spec semantics (and underlying XPath spec semantics along with that). Those details are neither necessary nor appropriate for clients to access, and basically give us a clear boundary for engine internals which are fair game to change without needing to make any corresponding client changes.

packages/xforms-engine/src/lib/codecs/select/MultipleValueSelectCodec.ts
eyelidlessness: I'll answer (2) first, because it may make the rest academic (for how it relates to `<odk:rank>`):

> 2. I wonder if the Rank codec will be the same as this one and if we can reuse or duplicate it.

Great catch! Yes, that's exactly what I expect. In fact, here's what I wrote in [the commit](https://github.com/getodk/web-forms/commit/9f8aead8f7a38c4cfdad334c9a4200ec9d7b0484):

> In fact, I wouldn’t be surprised if the codec logic for a hypothetical `RankControl` implementation would be identical [to `MultipleValueSelectCodec`], and we may want to discuss a more general naming scheme.

I think it's pretty likely naming is the only real work for a rank codec.

- - -

> 1. When codec run? I'm looking to know how to verify the output is what we expect

This is from memory, so it may not be exhaustive:

- The codec's `decodeValue` method is run whenever a serialized XML string is used as a (runtime/client facing) node value. Currently that includes:
    - Form init (reading default values from the form definition model); I also anticipate doing the same with values from submissions for edit
    - Any time a `calculate` expression is evaluated (i.e. when the node is relevant, on form init, or whenever anything referenced in the expression is updated)
- The codec's `encodeValue` method is run whenever a (runtime/client facing) node value is:
    - accessed as a node value in any XPath expression
    - serialized as XML (e.g. for submission)

For node types that support non-string value types, I think the best frame of reference for how they're validated is the scenario test module `bind-types.test.ts`.

But those won't be pertinent for `<odk:rank>`, which like selects will only support a string type. I'm not sure offhand how much `@getodk/scenario` coverage already exists for the rank control. If it isn't much (which might not surprise me), existing select tests are probably the best frame of reference.

I think for the most part, testing the rank functionality will implicitly test the codec, as it's effectively just splitting (exercised e.g. by testing default and/or calculated values) and joining (exercised e.g. by testing submission values).

packages/xforms-engine/src/lib/reactivity/createSelectItems.ts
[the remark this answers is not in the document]
eyelidlessness: This functionality is (currently) select-specific. The range control doesn't (currently) have a concept of "items" (or `valueOptions`, as named in the client interface). We may actually add support for that in the future, but I doubt it'll be here because the set of values a `RangeNode` might have will be static based on the bounds defined in the form.

I'll go into (much) more depth about the engine's use of reactivity in the docs I'm writing up. Briefly here for now so I don't leave it completely unanswered: the engine internally uses (SolidJS, at least for now) reactivity to implement the ODK XForms computation model. The computation model is a DAG, which is effectively what the reactivity model implements as well. So we get a lot of the recomputation logic "for free".

packages/xforms-engine/src/parse/body/control/ItemDefinition.ts
```diff
@@ -13,7 +14,7 @@ export class ItemDefinition extends BodyElementDefinition<'item'> {
 
 	constructor(
 		form: XFormDefinition,
-		override readonly parent: AnySelectDefinition,
+		override readonly parent: AnySelectControlDefinition,
```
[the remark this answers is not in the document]
eyelidlessness: Another good catch! Yeah, this module is moving up out of the `select` directory because I expect it to be used for rank too. And yeah, at a glance that annotation looks right. I'm not sure if we want to introduce some kind of common base class between them.

For what it's worth, I'm not sure the `parent` property really gets much use. So this type is really only guarding which controls can construct it (which is still valuable IMO).

packages/xforms-engine/src/parse/body/control/SelectControlDefinition.ts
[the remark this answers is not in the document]
eyelidlessness: Hopefully you've had a chance to read the WIP architecture docs I pushed up EOW, which I think addresses this much better than I can here.

Briefly, the current parsing order is roughly:

1. Parse XML (currently into a browser DOM tree, subject to change)
2. Parse `jr:` URLs to e.g. external secondary instances
3. Fetch and parse external secondary instances
4. Parse body
5. Parse model (binds, primary instance tree into node definitions, translations)

A client calls the engine's `initializeForm` to initiate this. Once all of that parsing is complete, it's used to build the primary instance state which is then returned to the client (well, resolved in a `Promise`, because step 3 does IO). We don't (currently) **expose** any parsing interfaces or logic to a client, it's all encapsulated in that `initializeForm` call.

packages/web-forms/src/lib/format/selectOptionId.ts
latin-panda: I like the idea of computing it upfront because it will eliminate the direct dependency between `RadioButton`, `CheckboxWidget` on `SelectXControl`. It also addresses the concern of prop drilling by centralizing the logic, simplifying the structure of the webform package, and preventing the need for passing it through multiple layers as the component hierarchy evolves 🤩 Do you think we should adopt this approach in this PR?

packages/xforms-engine/src/error/SelectValueTypeError.ts
latin-panda: Yes, that is a good idea!

packages/xforms-engine/src/instance/SelectControl.ts
latin-panda: Awesome! Thank you!

latin-panda (approved):
Thank you so much for answering all my question with great detail 🤩

I'm approving now to unblock, if you implement [this approach](https://github.com/getodk/web-forms/pull/277#discussion_r1922761841) here, I can review again 🙂

packages/web-forms/src/lib/format/selectOptionId.ts
eyelidlessness: (Capturing here for the record: we agreed to defer this so we can consider approach in context of `<odk:rank>`, where it'll also apply.)

packages/scenario/src/jr/select/SelectChoice.ts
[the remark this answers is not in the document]
latin-panda: >Can you clarify where you've seen that delay? Do you mean in Enketo (as in, "we'd like not to make it slower in Web Forms [than in Enketo]")? Or is this already a performance issue in Web Forms (as in, "we'd like to make it slower [than it already is]")?

Absolutely! Much of this comes from my previous experience with Enketo. We received reports from technical partners building forms, such as family registration forms. These forms often included a combination of groups and repeats. For example, in remote communities in Africa, families typically have many members, often 5 to 8 children. When users answered a question like "How many children?" with a number such as 7, the form would dynamically repeat 7 times a section, which can have groups and many questions. We received feedback about performance issues, such as the form hanging while filling it out for several minutes.

At the time, we recommended making the forms "flatter" by reducing the use of groups and repeats. Where possible, we also suggested splitting larger forms into multiple smaller forms to improve performance.

It's very common for health workers to use low-end devices with just 2GB of RAM. 
I don't think I can access those specific forms anymore due to the technical partner's policies, but we can come up with some complex lengthy ones, if we dont have it already. 

I'm going to use these for testing too, thanks!
- bench9-alt.xml
- nigeria_wards_external.xml

PR #279, opened 2025-01-17
chore: adding icon component to easily use material icons

latin-panda:
Closes https://github.com/getodk/web-forms/issues/280

Adds material icon font
Adds icon component with two inputs:
- alias: specifies the name of icon to render
- size: icon size -- it can be expanded to small (sm) and large (lg)

- [X] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [ ] Safari (iOS)
- [x] Android
- [ ] Not applicable

Sample of usage:
```js
<Icon :alias="'print'" :size="'md-18'"/>
<Icon :alias="'drag_indicator'" :size="'md-24'"/>
<Icon :alias="'language'" :size="'md-36'"/>
<Icon :alias="'warning'" :size="'md-48'"/>
```
What it renders:
Screenshot 2025-01-17 at 5 14 18 PM

[There are five steps to add an icon with IcoMoon](https://github.com/getodk/web-forms/tree/main/packages/web-forms), and there's a risk of accidentally deleting icons during this process. 
This solution imports the material icons library, which is 264KB. Maybe it can be compressed even further using Gzip and other assets, but that's not the scope of this PR. 

Because this project supports the latest version of browsers, I have [included only woff2](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_fonts/WOFF#css.at-rules.font-face.woff_2) (actually the Google's CDN for this font only provides woff2) 

- None

## What's changed
- No changes, just additions


latin-panda:
@eyelidlessness I kept getting issues with adding new icons to icoMoon, and I know we talked about having the svg inline. But it seemed like a low effort, so I went ahead and did this small PR. It will also make the Rank HTML cleaner :) 

It's more of a proposal, and if you agree, then I will finish this PR, and I can also replace icoMoon.

lognaturel:
I'm ready to review this when you are, @latin-panda! Maybe you can expand it a bit to replace the existing svg+icomoon icon and then take out icomoon?

> imports the material icons library, which is 264KB

And another thought occurs to me -- did you consider using a library like https://www.npmjs.com/package/@mdi/js that supports tree-shaking?

latin-panda:
Closing. It got too far behind `main`, we have a new PR: https://github.com/getodk/web-forms/pull/384

PR #281, opened 2025-01-18
fix illegal invocation issue when fetching form xml

DanielJDufour:

- [X] Chrome
- [X] Firefox
- [X] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Not applicable

I built a minimal test html file of the web-component build of web-forms and manually tested to make sure it works.

I had tried catching the error, but thought it best to prevent the error from happening at all.  The suggested change also preserve isomorphism and avoids dependency on any browser specific APIs.

This change shouldn't affect any users.

Will work with any form.

## What's changed
Basically just unpacked the fetchResource function (usually built-in fetch I think) before calling it.  When you try to call it on the object as options.fetchResource it throws an illegal invocation error.

## Note
I have only minimal experience with the web-forms code base, so I'd recommend double-checking with someone who has more experience.  I'm honestly a bit confused how this issue didn't come up earlier.  I wonder if it has to do with browser versions or maybe something Vite does.  I'm also wondering if it is specific to web component implementations (but I'm not sure why that is).

Open to your thoughts and feedback!

### DIFF
--- a/.changeset/ten-experts-type.md
+++ b/.changeset/ten-experts-type.md
@@ -1,0 +1,5 @@
+---
+"@getodk/xforms-engine": patch
+---
+
+Fix: illegal invocation when fetching form xml
--- a/packages/xforms-engine/src/instance/resource.ts
+++ b/packages/xforms-engine/src/instance/resource.ts
@@ -10,7 +10,10 @@ export interface ResourceOptions {
 }
 
 const fetchTextFromURL = async (resource: URL, options: ResourceOptions): Promise<string> => {
-	const response = await options.fetchResource(resource);
+	// destructuring prevents "Failed to execute 'fetch' on 'Window': Illegal invocation"
+	const { fetchResource } = options;
+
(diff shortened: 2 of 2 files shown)


DanielJDufour:
@spwoodcock, fyi

eyelidlessness (approved):
Thank you! I've added a changeset and approved. I'll merge once CI completes.

I was curious what was going on. It turns out this is a quirk of how `this` is bound when the global `fetch` function is assigned/accessed as a property of any object besides `globalThis`. If I run the following code in dev tools on a random page, I see the same "Illegal invocation" error:

```js
const notGlobal = { fetch };
await notGlobal.fetch(window.location.href);
```

I'll file an issue to add a test covering this case. It also theoretically affects the other config that defaults to global `fetch` (`fetchFormAttachment`). I don't think that's as urgent to address, because the fallback will generally fail anyway.

DanielJDufour:
Thanks!

PR #284, opened 2025-01-20
Web Component with Shadow Dom

DanielJDufour:
⚠️  This is a draft pull request.  I mostly created this to share knowledge and unblock the FMTM Web Forms integration work.  I'm not sure this exact implementation is best for everyone.

Builds a Web Component version of `<OdkWebForm/>` without the Shadow DOM

screenshot of web component demo
Screenshot 2025-01-20 at 4 28 04 PM

- [X] Chrome
- [X] Firefox
- [X] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Not applicable

Viewed minimal demo (link coming soon)

I'm not sure it's the best.  I've tried with using the Shadow DOM but that was more complicated.  I think the benefit of this solution is that it's minimal changes and can serve as a temporary implementation as a Shadow DOM - based approach is built.

By not using the shadow DOM, this approach is currently leaking styles to the global DOM.  It should be noted as experimental.

## What's changed
Basically built a web component version of OdkWebForm.  This built on top of the previous work on web components, but passes `shadowRoot: false` to createCustomElement.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "turbo": "^2.3.3",
     "typescript": "~5.7.2",
     "typescript-eslint": "^8.18.1",
-    "vue": "3.3.4",
+    "vue": "3.5.13",
     "vue-tsc": "^2.1.10"
   },
   "resolutions": {
--- a/packages/web-forms/package.json
+++ b/packages/web-forms/package.json
@@ -31,6 +31,7 @@
     "build:clean": "rimraf dist/ dist-demo/",
     "build:demo": "vite build --mode demo --outDir dist-demo",
     "build:js": "vite build",
+    "build:wc": "vite build --mode web-component --outDir dist-web-component",
     "dev": "vite",
     "dist-demo": "yarn build && yarn vite serve dist-demo --port 5174",
     "test": "npm-run-all -nl test:*",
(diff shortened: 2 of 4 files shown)


DanielJDufour:
Here's the demo file I used for manual testing:

```
<html>
	<head>
		<script src="../dist-web-component/odk-web-form.js" type="module"></script>
	</head>
	<body>
		<odk-web-form form-xml="http://localhost:8080/test-data/whova_form-BvHYggiJ.xml" />
	</body>
</html>

```

spwoodcock:
> I think the main question before us is: should the web component version of OdkWebForms without a Shadow DOM be merged and maintained? Or should this work live in a forked version until a web component with Shadow DOM is built. I'm open to either way.
> 
> Would like to hear everyone's thoughts :-)

I think for now we can work on a [fork](https://github.com/hotosm/web-forms) and embed the component from that (for us to get moving for now)

In the long term, we can discuss the approach with the web-forms team & decide on what type of PR would be preferable👍

lognaturel:
Thanks again for sharing this @DanielJDufour and @spwoodcock, I should have closed this once we decided y'all would maintain in your fork for now! We haven't so far received other interest in a Web Component so we're not sure when this will be prioritized for mainline but we will make sure to update as we learn about others wanting this.

PR #287, opened 2025-01-21
Include primary instance root attributes, namespace declarations in submission XML

### LINKED ISSUE #286: Pass through static id and version attributes on primary instance
The primary instance in a form definition has well-known `id` and `version` attributes that have static values and need to be passed through to the submission instance. These would be handled as part of broader attribute support (#285) but for now we will prioritize getting those two passed through.

The `version` attribute is optionally in the `http://openrosa.org/xforms` namespace (typically `orx`). The expectation is that clients like Web Forms would pass through what it gets on the form definition directly to the submission primary instance. All known consumers of forms recognize `version` without an explicit namespace so Web Forms could possibly strip the `orx` namespace to normalize the representation. That could possibly throw off a schema-aware consumer that uses the form definition in a sophisticated way but we are not aware of any system like that.

eyelidlessness:
Closes #286.

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Not applicable

- Added tests for attributes
- Updated tests for XML namespace declarations

It'd be better to have a more general solution for both attributes and namespace declarations.

Both would involve significant work on both parsing (to capture attributes and namespace prefixes/declarations throughout the primary instance subtree), as well as the primary instance state data model (as we don't have a concept of attributes, and presently ignore non-default-namespace element names).

I'd be remiss if I failed to mention that there would have been an even hacker way to deal with namespace declarations: abuse built-in DOM serialization (as it functionally does the same thing to derive namespaces as this change does more explicitly). I didn't do this because:

- It'd be **a lot less clear** what's happening and why
- I don't think it's standardized in any way, and I wouldn't be surprised if `jsdom`/any other DOM-compat library would make that technique unreliable

It's possible that even the simplified solution has ignored some edge case, which would be most likely around namespaces. I think I was fairly thorough, but the parse-stage collection of namespaces in a temporary map would be the most likely suspect.

It's also possible some logical error might have introduced subtle fallibility. I'm hopeful that letting the type system guide the change avoided most if not all of that risk.

I think any form will do?

## What's changed

- **Static** attributes on the primary instance root element are parsed, and re-serialized on submission
- Namespace declarations are parsed (or derived) on the primary instance root element, also re-serialized on submission

### DIFF
--- a/packages/scenario/test/submission.test.ts
+++ b/packages/scenario/test/submission.test.ts
@@ -141,7 +141,7 @@ describe('Form submission', () => {
 
 			expect(scenario).toHaveSerializedSubmissionXML(
 				// prettier-ignore
-				t('data',
+				t('data id="xml-serialization-basic-default-values"',
 					t('grp',
 						t('inp', defaults.inp ?? ''),
 						t('sel1', defaults.sel1 ?? ''),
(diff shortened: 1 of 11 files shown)


packages/scenario/test/submission.test.ts
```diff
@@ -156,6 +156,116 @@ describe('Form submission', () => {
...
+				expect(scenario).toHaveSerializedSubmissionXML(
+					// prettier-ignore
+					t(`data id="${formId}" version="${version}"`,
+						t('inp', 'val'),
+						t('meta',
+							t('instanceID', DEFAULT_INSTANCE_ID))).asXml()
+				);
+			});
+
+			it('preserves instance root attribute namespace prefixes, as they appear in the form definition', async () => {
```
lognaturel, line 196: Is the namespace declaration also included in the serialization? That could be layered on, I think, but is important too.

lognaturel (approved):
I should have checked what Collect and Enketo do with namespaces in submissions before commenting! Neither includes the default namespace declaration. I don't think it will break any existing systems to include but it does add more noise.

I'm inclined to go with this for now and come back to it if we see any issues. Alternately, we could special case the default namespace, maybe?

eyelidlessness:
I was also under the impression that Enketo omits the default namespace. Now I wish I'd mentioned that!

> Neither includes the default namespace declaration. I don't think it will break any existing systems to include but it does add more noise.

I'm actually concerned that it could. I know a lot of namespace stuff is pretty loosey goosey in a lot of places, but inclusion where it's not usually expected can cause subtle differences:

- ... in anything downstream that uses and builds on existing namespace-aware XML tooling. When omitted, the default namespace is `null` unless there's some explicit logic to override that (and off the top of my head, mechanisms to override are surprisingly inconsistent even in familiar browser APIs).

- ... in any round trip involving more than one XForms implementation. Not sure how high this risk is, but it does come to mind.

My instinct is to omit it as a special case, unless you feel strongly about a wait-and-see approach.

packages/xforms-engine/src/parse/model/RootNamespaceDeclaration.ts
```diff
@@ -0,0 +1,28 @@
...
+
+	constructor(
+		readonly prefix: string | null,
+		readonly namespaceURI: string | null
+	) {
+		if (prefix == null) {
+			// We intentionally omit the default namespace declaration, which is
+			// consistent with both Collect and Enketo. Including it may technically
+			// be more "correct", but consistency ensures we don't introduce subtle
+			// problems in any namespace-aware usage downstream.
```
eyelidlessness, line 15: Hmmm. So, I just added this comment because I think it could be confusing without. But I'm also a little bit cautious about this being a blanket omission without any logic _or_ explanation about the expected default. I am assuming here that the default namespace will always be XForms (`http://www.w3.org/2002/xforms`). But I could see making any of these changes:

- Explicitly check for that assumption (don't omit if some other default namespace)
- Explicitly assert that assumption (fail form load if some other default namespace)
- Expand this comment to mention that assumption, and let weird forms do weird things
- Test in Collect and Enketo to be sure

lognaturel (approved):

packages/xforms-engine/src/parse/model/RootNamespaceDeclaration.ts
eyelidlessness: Agree the comment covers everything important then! Thank you.

PR #288, opened 2025-01-22
feat(#243): Adds the rank question type

### LINKED ISSUE #243: Add the rank question type

latin-panda:
Closes #243

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [ ] Safari (iOS) <I don't have device>
- [x] Android

- I've added demo forms with `random`, `choice-filter`, `relevant` and `required` 
- It's tested in Android, Chrome for macOS, Safari and Firefox
- Demo videos has been shared with the team regularly to gather feedback

This implementation takes `select` as reference and reuses some of its code.

- N/A

Please assist in testing in iOS because I don't have an iPhone or iPad. Other than that, I did the following tests:

<details>
<summary> Test scenario video: Readonly and translation </summary>

<br>
The rank question has a condition where it becomes readonly when there are less than 5 options selected.
It also gets translated into French. 

[2-rank-with-choice-filter-read-only.xlsx](https://github.com/user-attachments/files/18608967/2-rank-with-choice-filter-read-only.xlsx)

https://github.com/user-attachments/assets/5d98b8e5-8148-4430-a7dc-88fd039298a4

</details>

<details>
<summary> Test scenario video: It should work in all browsers we support. Also, rank as required question. </summary>
<br>

It works in Chrome, Firefox, Safari and Chrome for Android. 
It shows rank working as required questions. 

https://github.com/user-attachments/assets/ecb64052-b92c-4f70-9779-aa1a840788d2

https://github.com/user-attachments/assets/c5a21b97-97b4-45f7-a987-4a1eb74aea1d

</details>

<details>
<summary> Test scenario video: Responsive design and becoming relevant again has value assigned </summary>
<br>

It shows rank in responsive design, where the buttons are hidden for phone devices. 
It also tests when the user ordered a rank's options, then it became no-relevant, and relevant again. At this point, the rank has a value assigned, and the user can submit the form without ordering the options again.  

https://github.com/user-attachments/assets/f0d04bf4-0491-47ec-8c54-27236b888906

</details>

## What's changed
- Added the rank question type in the xform-engine and web forms
- Added 2 demo rank forms
- Extracted common code between `select` and `rank` 
- New scenario tests were added.


yarn.lock
latin-panda, line 1: When using npm to add a new dependency, npm adds a few lines to package-lock.json that are relevant to the new dependency.

I was surprised to see so many changes after running `yarn add <the dependency>` in `packages/web-forms/package.json`. I researched a bit, and it seems that's how to add deps. Maybe this is yarn just making the whole tree more efficient (which npm is known to be less efficient)... 

I would appreciate a second pair of eyes, just in case.

packages/xforms-engine/src/lib/reactivity/createBaseItemset.ts
latin-panda, line 1: Extracted code that `select` and `rank` need from `packages/xforms-engine/src/lib/reactivity/createSelectItems.ts`

packages/xforms-engine/src/instance/children.ts
```diff
@@ -68,6 +65,10 @@ const isSelectDefinition = (definition: ControlNodeDefinition): definition is Se
 	return definition.bodyElement.type === 'select' || definition.bodyElement.type === 'select1';
 };
 
+const isRankDefinition = (definition: ControlNodeDefinition): definition is RankDefinition => {
+	return definition.bodyElement.type === 'rank';
```
latin-panda, line 69: In another PR, we can define all these strings (`rank`, `select`, etc) in one place in the engine. 
```
enum NodeType {
  Rank = "rank",
  Select = "select",
  Note = "note",
}
```

packages/xforms-engine/src/error/RankValueTypeError.ts
latin-panda, line 1: This is a suggestion about keeping all error definitions together. Let me know what you think

packages/web-forms/src/components/controls/RankControl.vue
```diff
@@ -0,0 +1,206 @@
...
+		<div
+			v-for="(option, index) in options"
+			:key="option.value"
+			class="rank-option"
+			:class="{ 'moving': highlight.index.value === index }"
+			tabindex="0"
+			@keydown.up.prevent="moveUp(index)"
+			@keydown.down.prevent="moveDown(index)">
+			<div class="rank-label">
+				<svg xmlns="http://www.w3.org/2000/svg" width="25" height="25" viewBox="0 0 768 768">
```
latin-panda, line 105: These SVGs are temporary, I have [another PR](https://github.com/getodk/web-forms/pull/279) that will make this nicer

packages/xforms-engine/src/error/RankError.ts
latin-panda, line 1: This suggestion is to keep Rank's error definitions together instead of individual files since each class is straightforward.

packages/xforms-engine/src/lib/reactivity/createBaseItemset.ts
```diff
@@ -0,0 +1,104 @@
+import { UpsertableMap } from '@getodk/common/lib/collections/UpsertableMap.ts';
```
latin-panda, line 1: Extracted code that `select` and `rank` need from `packages/xforms-engine/src/lib/reactivity/createSelectItems.ts`

yarn.lock
```diff
@@ -2,18 +2,24 @@
 # yarn lockfile v1
```
latin-panda, line 2: When using npm to add a new dependency, npm adds a few lines to package-lock.json that are relevant to the new dependency.

I was surprised to see so many changes after running yarn add <the dependency> in packages/web-forms/package.json. I researched a bit, and it seems that's how to add deps. Maybe this is yarn just making the whole tree more efficient (which npm is known to be less efficient)...

I would appreciate a second pair of eyes, just in case.

packages/xforms-engine/src/error/RankError.ts
```diff
@@ -0,0 +1,20 @@
+import { XFormsSpecViolationError } from './XFormsSpecViolationError.ts';
```
latin-panda, line 1: This suggestion is to keep Rank's error definitions together instead of individual files since each class is straightforward.

latin-panda:
Hi @eyelidlessness, if you have time, I would really appreciate an early review while I'm working on the test coverage. I'm especially interested in your feedback on the xform-engine implementation.

I manually tested `rank` and `select` in Chrome, Firefox, and Safari.

packages/xforms-engine/src/client/RankNode.ts
```diff
@@ -0,0 +1,50 @@
...
+	get valueOptions(): RankValueOptions;
+
+	/**
+	 * An ordered collection of values from {@link RankItem}.
+	 * The order of the items is important and must be preserved during processing.
+	 */
+	get value(): readonly string[];
+}
+
+export interface RankDefinition extends LeafNodeDefinition<V> {
```
eyelidlessness, line 26:
```suggestion
export interface RankDefinition<V extends ValueType = ValueType> extends LeafNodeDefinition<V> {
```

This probably got jumbled up in the mixup around select value types, but `RankDefinition` should have a `ValueType` type parameter, like `SelectDefinition` does in `SelectNode`. This is somewhat important so the types match up with the runtime check (in `RankControl`) to enforce only supporting the `'string'` value type. (I'll leave a corresponding suggestion there.)

packages/xforms-engine/src/client/RankNode.ts
```diff
@@ -0,0 +1,50 @@
...
+	readonly bodyElement: RankControlDefinition;
+}
+
+export interface RankNode extends BaseValueNode<'string', readonly string[]> {
+	readonly nodeType: RankType;
+	readonly valueType: 'string';
+	readonly definition: RankDefinition<'string'>;
```
eyelidlessness, line 33:
```suggestion
	readonly definition: RankDefinition<'string'>;
	readonly appearances: UnknownAppearanceDefinition;
```

(See explanatory comment at top of `RankControlDefinition.ts`)

packages/xforms-engine/src/client/RankNode.ts
```diff
@@ -0,0 +1,50 @@
...
+	readonly root: RootNode;
+	readonly parent: GeneralParentNode;
+	readonly currentState: RankNodeState;
+	readonly validationState: LeafNodeValidationState;
+
+	/**
+	 * Convenience API to get the {@link RankItem}'s label.
+	 */
+	getValueLabel(value: string): string;
```
eyelidlessness, line 42:
```suggestion
	getValueLabel(value: string): TextRange<'item-label'> | null;
```

You wouldn't know this without background knowledge of the design intent, but `TextRange` is a stub type we'll use to provide support for ODK XForms' subset of Markdown/styling. While the support is incomplete, I think we should continue to use the same interface everywhere formatting will be applied when support is added. For now, clients reference the `asString` property to work around this incompleteness.

(I'll leave a corresponding suggestion in `RankControl.ts`.)

packages/xforms-engine/src/client/hierarchy.ts
```diff
@@ -11,19 +11,18 @@ import type { RootNode } from './RootNode.ts';
...
 import type { TriggerNode } from './TriggerNode.ts';
-import type { RankNode } from './unsupported/RankNode.ts';
+import type { RankNode } from './RankNode.ts';
 import type { UploadNode } from './unsupported/UploadNode.ts';
 
 // prettier-ignore
-export type AnyUnsupportedControlNode =
-	| RankNode
-	| UploadNode;
+export type AnyUnsupportedControlNode = UploadNode;
```
eyelidlessness, line 18: It looks so lonely! But it'll be in good company soon enough.

packages/xforms-engine/src/error/RankError.ts
eyelidlessness: Ah I see you addressed this before me. We've generally settled on a convention of 1 _exported class_ to 1 same-named module. I think this is the most common idiom in most TS/JS projects (probably inherited from Java). I don't think we've applied the convention as consistently as maybe we ought to.

But I think **especially** for things like errors it would be valuable, because the _eventual intent_ is that they'll be used to produce documentation of which errors can occur. (The value will probably be most visible in generated documentation, where each class/module is a unit.)

packages/xforms-engine/src/error/RankError.ts
```diff
@@ -0,0 +1,20 @@
...
+	constructor(definition: RankDefinition<UnsupportedBaseItemValueType>) {
+		const { valueType } = definition;
+
+		super(
+			`Ranks of type ${valueType} are not currently supported. If this functionality would be useful for your form, your feedback is welcome!`
+		);
+	}
+}
+
+export class RankFunctionalityError extends XPathFunctionalityError {
```
eyelidlessness, line 16: Aside from moving to a same-named module... I don't think this should subclass `XPathFunctionalityError`. Not sure I can guess the intent... was it meant to subclass `XFormsSpecViolationError`? I don't know if that's right or not, but it doesn't feel right either.

packages/xforms-engine/src/instance/RankControl.ts
```diff
@@ -0,0 +1,144 @@
...
+import { sharedValueCodecs } from '../lib/codecs/getSharedValueCodec.ts';
+import type { AnyNodeDefinition } from '../parse/model/NodeDefinition.ts';
+import { createItemCollection } from '../lib/reactivity/createItemCollection.ts';
+
+type AssertRangeNodeDefinition = (definition: RankDefinition) => asserts definition is RankDefinition<'string'>;
+const assertRangeNodeDefinition: AssertRangeNodeDefinition = (definition) => {
+	if (definition.valueType !== 'string') {
+		throw new RankValueTypeError(definition);
+	}
+};
```
eyelidlessness, line 37:
```suggestion
export type AnyRankDefinition = {
	[V in ValueType]: RankDefinition<V>;
}[ValueType];

type AssertRankNodeDefinition = (
	definition: AnyRankDefinition
) => asserts definition is RankDefinition<'string'>;

const assertRankNodeDefinition: AssertRankNodeDefinition = (definition) => {
	if (definition.valueType !== 'string') {
		throw new RankValueTypeError(definition);
	}
};
```

1. Note the name changes (range -> rank). (I make similar-word typos like this _all the time_ and I know it's maddening! I can't count the number of times I've typed `NoteNote` or `NodeNode` 🙃)
2. Let me know if the use of mapped type for `AnyRankDefinition` isn't clear. The intent is to derive the type so you don't need to define a huge union just to discard most of it.

I'll add a bit more explanation at the `from` signature, and hopefully tie the whole definition type parameter -> `'string'` thing together.

packages/xforms-engine/src/instance/RankControl.ts
```diff
@@ -0,0 +1,144 @@
...
+	extends ValueNode<'string', RankDefinition<'string'>, readonly string[], readonly string[]>
+	implements
+		RankNode,
+		XFormsXPathElement,
+		EvaluationContext,
+		ValidationContext,
+		ClientReactiveSubmittableValueNode
+{
+	static from(parent: GeneralParentNode, definition: RankDefinition): RankControl {
+		assertRangeNodeDefinition(definition);
```
eyelidlessness, line 57:
```suggestion
	static from(parent: GeneralParentNode, definition: RankDefinition): RankControl;
	static from(parent: GeneralParentNode, definition: AnyRankDefinition): RankControl {
		assertRankNodeDefinition(definition);
```

So... this change, combined with making `RankDefinition` generic over `ValueType`, and combined with `AnyRankDefinition` as input to the assertion. All together, they ensure:

1. We know that we _might_ parse a `RankDefinition` of any value type. At parse time, this would represent a combination of `<odk:rank ref="/data/some-rank-ref">` + `<bind nodeset="/data/some-rank-ref" type="$anyValueType"/>`.
2. We know that any value type besides `'string'` is unsupported for the control.

Without making much more disruptive changes to the parse logic, this is the first point in the form init process where we have all of that information in one place, and where we know we're just dealing with rank. So... it's not exactly the ideal place to enforce the string-only restriction (it feels more like a parse responsibility!), but it's what we currently have to work with in a reasonably scoped change.

The combination of changes to types basically represent the same thing we're doing at runtime:

- Parse form definition into something overly broad (rank with any value type)
- Validate that we're only dealing with the supported case (rank with string value type)

The `assertRangeNodeDefinition` signature basically represents _at the type level_ that that's what happens at runtime, and ensures that we ultimately only return `RankDefinition<'string'>` to clients (as is the contract on `RankNode`).

packages/xforms-engine/src/instance/RankControl.ts
```diff
@@ -0,0 +1,144 @@
...
+
+	// XFormsXPathElement
+	override readonly [XPathNodeKindKey] = 'element';
+
+	// InstanceNode
+	protected readonly state: SharedNodeState<RankControlStateSpec>;
+	protected readonly engineState: EngineState<RankControlStateSpec>;
+
+	// RankNode
+	readonly nodeType = 'rank';
```
eyelidlessness, line 73: Here we'll also want to add `appearances`, see comment at the top of `RankControlDefinition.ts`.

packages/xforms-engine/src/instance/RankControl.ts
```diff
@@ -0,0 +1,144 @@
...
+
+		const state = createSharedNodeState(
+			this.scope,
+			{
+				reference: this.contextReference,
+				readonly: this.isReadonly,
+				relevant: this.isRelevant,
+				required: this.isRequired,
+				label: createNodeLabel(this, definition as AnyNodeDefinition),
+				hint: createFieldHint(this, definition as AnyNodeDefinition),
```
eyelidlessness, line 115:
```suggestion
				label: createNodeLabel(this, definition),
				hint: createFieldHint(this, definition),
```

Only one of these is a type error, but both are unnecessary and overly broad.

Sometimes it really is necessary or just the least-bad option to type cast, but oftentimes it masks other problems. It's usually worth exploring other ways to work with the type system before overriding it.

(If I had to guess, these casts were probably added to work around other type errors, possibly around the mismatched `ValueType` generic?)

packages/xforms-engine/src/instance/RankControl.ts
```diff
@@ -0,0 +1,144 @@
...
+
+		this.state = state;
+		this.engineState = state.engineState;
+		this.currentState = state.currentState;
+	}
+
+	getValueLabel(value: string): string {
+		const valueOption = this.currentState.valueOptions.find(item => item.value === value);
+		return valueOption.label.asString;
+	}
```
eyelidlessness, line 132:
```suggestion
	getValueLabel(value: string): TextRange<'item-label'> | null {
		const valueOption = this.currentState.valueOptions.find((item) => item.value === value);

		return valueOption?.label ?? null;
	}
```

(I left an explanation on `RankNode.getValueLabel`.)

packages/xforms-engine/src/instance/RankControl.ts
```diff
@@ -0,0 +1,144 @@
...
+
+	setValues(valuesInOrder: readonly string[]): Root {
+		const sourceValues: string[] = Array.from(this.mapOptionsByValue().keys());
+		const hasAllValues = !sourceValues.some((sourceValue) => valuesInOrder.includes(sourceValue));
+		if (hasAllValues) {
```
eyelidlessness, line 137:
```suggestion
		const hasAllValues = sourceValues.every((sourceValue) => valuesInOrder.includes(sourceValue));

		if (!hasAllValues) {
```

There's a name/logic mismatch here. I _prefer_ the affirmative name, and I think the negative conditional is clear enough. So suggestion is to make the logic match the name.

packages/xforms-engine/src/instance/RankControl.ts
```diff
@@ -0,0 +1,144 @@
...
+			throw new RankFunctionalityError('There are missing options. Rank should have all options.', 'Rank Control');
```
eyelidlessness, line 138: As mentioned on inline where the class is defined, I'm not sure about this error's class hierarchy. I think addressing that will help sort out the type error here too.

packages/xforms-engine/src/instance/children.ts
eyelidlessness: So... let's talk more about enums! I personally like them (well, string enums like this here). But they're generally frowned upon in TS, and likely to be either deprecated or unsupported in future TS updates and/or other target environments.

Currently, we only use them in two places and I was kind of fighting myself over that when I added them.

eyelidlessness (commented):
I think I've looked at _almost_ all of the engine changes, but I'm going to have to come back to the codecs.

I know there's already a LOT of commentary. I'm not sure how best to provide all the pertinent context without it being overwhelming, so let me know if you'd find it more helpful to jump on a call or something!

Either way, I wanted to share my thoughts so far before my brain runs out of brain juice for the day.

packages/xforms-engine/src/lib/reactivity/createItemCollection.ts
```diff
@@ -38,19 +42,19 @@ const createSelectItemLabel = (
 	return createTextRange(context, 'item-label', label);
 };
 
-interface SourceValueSelectItem {
+interface SourceValueItem {
```
eyelidlessness, line 45: Not urgent to change unless another good naming option comes to mind, but I do want to make note that this is carrying on a naming mistake I left behind—introduced in #273 (where introducing select types was the underlying mistake), probably should have been renamed in #277.

It's worth noting that whatever we _would_ name this, it'll have the same effective type as the currently-named `ItemType`, which I also called out for rethinking its name. So maybe there's one thing to address rather than two! (In other words, maybe whatever this should/would be named, we reference it in lieu of `ItemType` instead of trying to come up with two better names for effectively the same interface.)

packages/xforms-engine/test/lib/reactivity/createSelectItems.test.ts
```diff
@@ -15,16 +15,16 @@ import {
...
 
 /**
  * @todo Consider these alternative testing strategies:
  *
- * - Reducing tests of reactive internals like {@link createSelectItems} to more
+ * - Reducing tests of reactive internals like {@link createItemCollection} to more
  *   conventional unit tests: If there's a reasonable way to do that, it would
  *   probably begin (especially in this case) with relaxing the
- *   {@link createSelectItems} signature to accept something more minimal than a
+ *   {@link createItemCollection} signature to accept something more minimal than a
```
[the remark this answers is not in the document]
eyelidlessness: Hm, but piggybacking on that:

- I think this test module name should be renamed to match
- The `describe` description should be updated too, at least to match the rename
- I'm iffy on whether there's value in adding rank-specific tests (it'd be more thorough, but redundant)... and similarly iffy about whether `<odk:rank>` should also be mentioned in the `describe` description

packages/web-forms/src/components/controls/RankControl.vue
```diff
@@ -0,0 +1,209 @@
...
+};
+</script>
+
+<template>
+	<ControlText :question="question" />
+
+	<VueDraggable
+		:id="question.nodeId"
+		v-model="options"
+		:delay="HOLD_DELAY"
```
eyelidlessness, line 91:
```suggestion
		:delay="HOLD_DELAY"
		:delay-on-touch-only="true"
```

It took me several tries in the UI to get an item to drag, before I realized there was a delay. I think there are other ways to make that more obvious without eliminating the delay, but I think intuitively I just don't expect a delay with a mouse. I was pleased to see this option already existed (and of course we're at the mercy of the third party library getting it right, but I think that's a reasonable starting point).

This one change would address _by far_ the biggest concern I have from playing around with the change locally.

packages/web-forms/src/components/controls/RankControl.vue
```diff
@@ -0,0 +1,209 @@
...
+			</div>
+
+			<div class="rank-buttons">
+				<button @click="moveUp(index)" @mousedown="setHighlight(index)">
+					<svg xmlns="http://www.w3.org/2000/svg" width="15" height="15" viewBox="0 0 768 768">
+						<path d="M384 256.5l192 192-45 45-147-147-147 147-45-45z" />
+					</svg>
+				</button>
+
+				<button @click="moveDown(index)" @mousedown="setHighlight(index)">
```
eyelidlessness, line 121: I'd make a suggestion in the styles, but each of these should get distinct classes to target for those styles. Anyway, the suggestion I want to make is to either hide, dim, or otherwise de-emphasize:

- The up button, for the first item
- The down button, for the last item

I don't know if it's just a bug in how my brain filters out noise, but I barely noticed the buttons even existed. For some reason, I think making the ones which don't do anything less visible would help with that. (And I think they're definitely useful to have, and to make noticeable! I could see myself using them instead of drag/drop a lot of the time.)

packages/web-forms/src/components/controls/RankControl.vue
```diff
@@ -0,0 +1,209 @@
...
+		:show-message="touched || submitPressed"
+	/>
+</template>
+
+<style scoped lang="scss">
+@import 'primeflex/core/_variables.scss';
+
+$rankSpacing: 7px;
+$rankBorder: 1px solid var(--surface-200);
+$rankBorderRadius: 10px;
```
eyelidlessness, line 141: Does `vue-draggable-plus` reference these variable names too? Or do they only affect the local styles? If they're just local, I'd prefer to use CSS custom properties ("variables") over Sass syntax. In general I'd like to try to keep Sass-specific syntax light, if we use it at all. CSS has come a long way, and Sass comes with a lot of surprises and a few headaches. In the long run, it would be nice to see if we can get away with eliminating (at least direct) Sass usage entirely.

eyelidlessness (commented):
I know there are some changes incoming, particularly on the UI side, so I took a light touch there for now.

And I think I've covered everything that I could on the engine side.

One thing that we'll still need to address is `@getodk/scenario`. At minimum, it'll need to pass type checks. Mostly that will involve creating and referencing rank-specific equivalents of various things (a good starting point would be to reference select-specific things in `scenario` package). I think I also mentioned a few specific test cases that would make sense to add. I'm happy to help, I know there's a whole slew of other new concepts to discover there!

@latin-panda All in all, this is an **excellent start**! I am honestly more than a little mind boggled by how much context you managed to gain on your own with very little information or guidance. I hope the feedback so far is helpful getting this closer to the finish line, and I'm looking forward to taking another pass when you're ready!

packages/web-forms/src/components/controls/RankControl.vue
eyelidlessness: Oh... I didn't know where to put this, and then forgot to post it in the last set of comments.

I'd like to discuss (and probably bring in @alyblenkin):

- Does it make sense for the whole item to be draggable, or just (a touch-friendly area around) the handle icon?
- Should that differ on touch vs pointer input?
- If we reduce draggable area for pointer input, I'd want to make sure the drag cursor is only shown where dragging works. (This was my original feedback, before I realized that the whole item is draggable, hence the other questions.)

packages/xforms-engine/src/parse/body/control/SelectControlDefinition.ts
```diff
@@ -52,7 +52,7 @@ export class SelectControlDefinition<Type extends SelectType> extends ControlDef
 		const itemsetElement = getItemsetElement(element);
 		const itemElements = getItemElements(element);
 
-		if (itemsetElement == null) {
+		if (itemsetElement === null) {
```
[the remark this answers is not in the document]
latin-panda: How do you feel about this? 
```js
if (!itemsetElement) {
```
We don't expect `itemsetElement` to be zero or false, and it would prevent another developer from making the same suggestion

packages/xforms-engine/src/instance/children.ts
latin-panda: In that case, we could use regular objects or maps.

```
// Option 1
const NodeType {
  Rank: "rank",
  Select: "select",
  Note: "note",
}
Object.freeze(NodeType); 

// Option 2
const NodeType = new Map([
  ["Rank", "rank"],
  ["Select", "select"],
  ["Note", "note"]
]);
```

packages/xforms-engine/src/parse/body/control/RankControlDefinition.ts
```diff
@@ -1,27 +1,54 @@
-import { ODK_NAMESPACE_URI } from '@getodk/common/constants/xmlns.ts';
+import type { LocalNamedElement } from '@getodk/common/types/dom.ts';
```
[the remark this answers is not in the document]
latin-panda: It makes sense! Fixed

packages/xforms-engine/src/parse/body/control/RankControlDefinition.ts
```diff
@@ -1,27 +1,54 @@
...
-export class RankControlDefinition extends ControlDefinition<'rank'> {
 	static override isCompatible(localName: string, element: Element): boolean {
-		return localName === 'rank' && element.namespaceURI === ODK_NAMESPACE_URI;
+		return RankControlDefinition.isRankElement(element);
 	}
 
-	readonly type = 'rank';
-	readonly appearances: UnknownAppearanceDefinition;
+	readonly type: RankType;
+	readonly element: RankElement;
```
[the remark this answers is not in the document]
latin-panda: Done

packages/xforms-engine/src/parse/body/control/RankControlDefinition.ts
```diff
@@ -1,27 +1,54 @@
...
+
+export class RankControlDefinition extends ControlDefinition<RankType> {
+	private static isRankElement(element: Element): element is RankElement {
+		return element.localName === 'rank';
+	}
 
-export class RankControlDefinition extends ControlDefinition<'rank'> {
 	static override isCompatible(localName: string, element: Element): boolean {
-		return localName === 'rank' && element.namespaceURI === ODK_NAMESPACE_URI;
+		return RankControlDefinition.isRankElement(element);
```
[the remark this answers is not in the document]
latin-panda: Fixed

packages/xforms-engine/src/parse/body/control/RankControlDefinition.ts
```diff
@@ -1,27 +1,54 @@
...
+			throw new Error(`Invalid rank element: <${element.nodeName}>`);
+		}
+
 		super(form, parent, element);
 
-		this.appearances = unknownAppearanceParser.parseFrom(element, 'appearance');
+		this.type = element.localName as RankType;
+		this.element = element;
+		const itemsetElement = getItemsetElement(element);
+		const itemElements = getItemElements(element);
```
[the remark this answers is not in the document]
latin-panda: Good idea! Fixed

packages/xforms-engine/test/lib/reactivity/createSelectItems.test.ts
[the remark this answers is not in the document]
latin-panda: I fixed the comment and mentioned `<odk:rank>` in the `describe` since it's related. But I didn't add rank-specific tests since they are almost the same as `select` and feel redundant.

packages/xforms-engine/src/lib/reactivity/createItemCollection.ts
```diff
@@ -17,6 +19,8 @@ import { createComputedExpression } from './createComputedExpression.ts';
 import type { ReactiveScope } from './scope.ts';
 import { createTextRange } from './text/createTextRange.ts';
 
+type ItemCollectionControl = SelectControl | RankControl;
+type ItemType = SelectItem | RankItem
```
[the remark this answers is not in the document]
latin-panda: Fixed! I used `Item`

packages/xforms-engine/src/client/RankNode.ts
latin-panda: Fixed!

packages/xforms-engine/src/client/RankNode.ts
latin-panda: Oh thanks for explaining, I didn't catch it like that before! Fixed.

packages/xforms-engine/src/error/RankError.ts
latin-panda: I intend to throw an error when not all `valueOptions` are in `value` 

```js
setValues(valuesInOrder: readonly string[]): Root {
		const sourceValues: string[] = Array.from(this.mapOptionsByValue().keys());
		const hasAllValues = sourceValues.some((sourceValue) => valuesInOrder.includes(sourceValue));
		if (!hasAllValues) {
			throw new RankFunctionalityError('There are missing options. Rank should have all options.', 'Rank Control');
		}

		this.setValueState(valuesInOrder);
		return this.root;
	}
```

So, I was looking for an existing error that fits, so I thought that this qualifies as "a functionality" error, that's how I interpreted `XPathFunctionalityError`. 
I can make this one a stand a long class:

```js
export class RankMissingValuesError extends Error {
	constructor(message string) {
		super(message);
	}
}
```

What do you think?

packages/xforms-engine/src/instance/RankControl.ts
latin-panda: > Note the name changes (range -> rank)

🤦🏽‍♀️ Thanks!

packages/xforms-engine/src/instance/RankControl.ts
latin-panda: Agree! Fixed

packages/xforms-engine/src/instance/RankControl.ts
latin-panda: Fixed! Thank you

packages/xforms-engine/src/lib/codecs/BaseItemCollectionCodec.ts
```diff
@@ -1,18 +1,18 @@
...
  * This generalizes the application of a {@link SharedValueCodec} implementation
- * over individual select values, where those values are serialized as a
+ * over individual select and rank values, where those values are serialized as a
  * whitespace-separated list. All other encoding and decoding logic is deferred
- * to the provided {@link baseCodec}, ensuring that select value types are
+ * to the provided {@link baseCodec}, ensuring that select and rank value types are
  * treated consistently with the same underlying data types for other controls.
  */
-export class MultipleValueSelectCodec extends BaseSelectCodec<readonly string[]> {
+export class BaseItemCollectionCodec extends BaseItemCodec<readonly string[]> {
```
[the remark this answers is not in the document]
latin-panda: A lot better, thanks! Fixed

packages/xforms-engine/src/lib/reactivity/createItemCollection.ts
latin-panda: I have removed this interface because it's the same as `ItemType` (now called `Item`). Let me know if it looks good :)

packages/web-forms/src/components/controls/RankControl.vue
latin-panda: Good catch! I have hidden them: 

Screenshot 2025-01-27 at 1 12 55 PM

packages/web-forms/src/components/controls/RankControl.vue
latin-panda: > Does vue-draggable-plus reference these variable names, too? 

No, it doesn't. These variables are for this specific component. 

> If they're just local, I'd prefer to use CSS custom properties ("variables") 

Sure! I have fixed it, and it is now using CSS variables. 
I have left the variable's definition in a separate block at the top. IMO, It's more organized that way. Let me know what you think.

packages/xforms-engine/src/parse/body/control/RankControlDefinition.ts
```diff
@@ -1,27 +1,54 @@
...
+		if (itemsetElement === null) {
+			this.itemset = null;
+			this.items = itemElements.map((itemElement) => new ItemDefinition(form, this, itemElement));
+		} else {
+			if (itemElements.length > 0) {
+				throw new Error(`<${element.nodeName}> has both <itemset> and <item> children`);
+			}
+
+			this.items = [];
+			this.itemset = new ItemsetDefinition(form, this, itemsetElement);
```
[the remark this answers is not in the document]
latin-panda: I'm a bit confused, just to double-check, is it `AnyRankDefinition` or `RankControlDefinition`?:
```js
override readonly parent: AnySelectControlDefinition |  AnyRankDefinition
```

Or 

```js
override readonly parent: AnySelectControlDefinition |  RankControlDefinition
```

packages/web-forms/src/components/controls/RankControl.vue
latin-panda: > Does it make sense for the whole item to be draggable, or just (a touch-friendly area around) the handle icon?

Last time I met @alyblenkin, it's the whole item, so to make it more accessible, I tried to capture it in [this comment](https://github.com/getodk/web-forms/issues/243#issuecomment-2594192464).

> Hold and drag are anywhere in the option (not just the icon), making them easier to use.

latin-panda: > It took me several tries in the UI to get an item to drag, before I realized there was a delay. I think there are other ways to make that more obvious without eliminating the delay, but I think intuitively I just don't expect a delay with a mouse.

I'm going to experiment and see if there's a way to detect touch instead of mouse.

packages/xforms-engine/src/parse/body/control/SelectControlDefinition.ts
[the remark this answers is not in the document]
eyelidlessness: In general my preference is to use operators that most clearly express the intended condition or operation, most closely for the type of the operand, unless there's a compelling reason to deviate from that[^1]. The `== null`/`!= null` is a special case specifically because it's the most common idiom in the language. And I believe that's because the language is unusual in having more than one representation of `null`.

I'm especially cautious about conflating truthy/falsy semantics with nullable semantics, because it's a common source of bugs, and a common source of confusion about code intent after the fact. While you're right that we don't expect a falsy value here besides `null` or `undefined`, and while I don't expect it to change in this case, I feel a lot more comfortable with a foundation of having one answer to "how do I check if a binding has a value?" The alternative is that it's subjective and case-by-case, where we have to consider not just whether something might have a falsy-but-non-nullish value now but _whether it may evolve to have one in the future_.

If this were introduced in new code or alongside other related changes, I think I'd probably either have missed it entirely or just made a mental note to adjust it to the more common idiom when I'm making other changes in the same area. It mostly jumped out at me because it's the only change in the module, and neither change is necessary. I don't feel **super strongly** about sending it back for another change, but if nothing else I think it'd be good long term not to add noise to the commit history for it.

[^1]: I think the most common justification is performance, e.g. various optimization techniques with bitwise operators. In the past, another common justification was reducing bytes on the wire, but that's much less a concern with a build step.

packages/xforms-engine/src/error/RankError.ts
eyelidlessness: Yeah I think that's probably best class hierarchy in the short term. We have outstanding work around errors pretty much project wide, and I expect we'll get a much clearer picture of how to model them when we consider them together.

packages/xforms-engine/src/instance/RankControl.ts
eyelidlessness: Right.

So, **usually** the intent in the engine—_loosely_ inspired by the concept of ["parse, don't validate"](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)—is to use the parse phase of form initialization to:

- Build a coherent model of the form, in a way that instance state can make sense of it.
- Produce error cases when parsing something that's either invalid or unsupported, which effectively ensures that _if it's parsed, it **is** valid_.

Handling this here is basically a **really stinky smell** that the current design around parsing value types doesn't quite meet that goal. When we have some time to tighten things up, I think we'll probably end up breaking out subclasses of `LeafNodeDefinition` for each node type. That way we'll be able to enforce the value type semantics of a given node type while parsing its node definition, instead of enforcing it here (which definitely feels wrong!).

eyelidlessness: Er, maybe it'll be more clear if I address this head on:

> I struggled to understand why adding support for any value type instead of just string from the start.

The intent isn't so much to **add support** (and then take it away), but to make sure the types reflect that the parsed `LeafNodeDefinition` may **already have** any `valueType`, and nothing has been done (yet) to say "that's not allowed for a rank control".

packages/xforms-engine/src/error/RankError.ts
latin-panda: I've separated the errors to different files

packages/xforms-engine/src/instance/RankControl.ts
latin-panda: The last 2 comments make it a lot clearer now, thank you!

packages/web-forms/src/components/controls/RankControl.vue
eyelidlessness: > I'm going to experiment and see if there's a way to detect touch instead of mouse.

I was wondering if it was a good idea to mix code suggestions and text the way I have in several comments. In this case I don't think it helped! So just in case you didn't see it in the suggestion, adding `:delay-on-touch-only="true"` should work. It did for me (although I guess I didn't try it on mobile, I'm just assuming the library works!).

It's listed as `delayOnTouchOnly`[^1] in the [`vue-draggable-plus` API docs](https://vue-draggable-plus.pages.dev/en/api/#api-1)—but the docs are wrong about the type, it is definitely a `boolean` in the prop types.

[^1]: It's still taking me some getting used to that Vue uses camelCase and kebab-case interchangeably for props. Just mentioning that here in case it also helps you to keep in mind when referencing Vue-related docs.

packages/xforms-engine/src/instance/RankControl.ts
latin-panda: Resolved

packages/web-forms/src/components/controls/RankControl.vue
eyelidlessness: Paging @alyblenkin for opinions!

My gut seeing it now is that the last item up arrow being in the down arrow column feels really weird. Any of these might be better:

- Reduce opacity
- Set `visibility: none` (so they still take the same space but aren't shown)
- Some other disabled/not clickable style
- Just show them after all (e.g. I was wrong, sorry for the mixup!)

packages/web-forms/src/components/controls/RankControl.vue
eyelidlessness: I think that's a great pattern! I've often ended up with something similar. Maybe even feels worth seeing if we can find a lint rule that helps us make it a convention.

packages/xforms-engine/src/parse/body/control/SelectControlDefinition.ts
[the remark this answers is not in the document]
latin-panda: I see your point, and I’ve made adjustments accordingly:
1. I’ve reverted the change in SelectControlDefinition, so this file has no modifications.
2. In RankControlDefinition, I’ve opted for an explicit check to address both concerns: the falsy check and the strict equality comparison. This ensures the code’s intent is clear and minimizes ambiguity: `if (itemsetElement === null || itemElements === undefined) {`

I believe this solution achieves the best of both worlds 🙂

packages/web-forms/src/components/controls/RankControl.vue
latin-panda: Oh yeah, I missed it! It's working good in my Android phone as well

packages/web-forms/src/components/controls/RankControl.vue
latin-panda: This is how it looks when keeping the button's place
Screenshot 2025-01-27 at 4 27 49 PM

packages/xforms-engine/src/client/RankNode.ts
```diff
@@ -0,0 +1,50 @@
...
+export type RankValueOptions = readonly RankItem[];
+
+export interface RankNodeState extends BaseValueNodeState<readonly string[]> {
+	get valueOptions(): RankValueOptions;
+
+	/**
+	 * An ordered collection of values from {@link RankItem}.
+	 * The order of the items is important and must be preserved during processing.
+	 */
+	get value(): readonly string[];
```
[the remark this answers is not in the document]
latin-panda: I’d prefer to address this update to `select` and `rank` in a separate PR to keep this one focused and manageable.

packages/web-forms/src/components/controls/RankControl.vue
alyblenkin: So sorry for the slow response! I think it is very strange not to see the other button, and I haven't seen something like this before. @latin-panda I don't think we spoke about this in our quick chat, but I was originally thinking we show both but one is disabled so everything stays in the same place. So the user can try clicking but it won't do anything. Now that I'm seeing it, perhaps making it disabled and lighter might be a good combination, so the behaviour is very clear.

packages/xforms-engine/src/lib/codecs/getSelectCodec.ts
latin-panda, line 1: I moved this file out from /select directory since it was the only file there

yarn.lock
eyelidlessness: I'm sorry! I missed this in earlier passes.

When I add (or update) a dependency, I make the change in the appropriate `package.json`, and then run `yarn install` at the monorepo root. I'm not saying you _shouldn't_ trust the `yarn` CLI sub-commands to do this, but I use a small subset of those sub-commands to reduce the set of things I have to know about and trust in what the tool's doing.

At that point, I do trust the tool to do the right thing with automatic changes to `yarn.lock` when I add new dependencies (and _almost always_ for updates to existing deps).

<details>
<summary>Why "almost always" for updates?</summary>

**Note:** I'm collapsing this because it's not pertinent here, but it's be useful to know for post-release dependency updates.

There are some very specific exceptions where I pay close attention around updates to direct dependencies on:

- `tree-sitter` and anything related to that package, which are often an ordeal (and it's different every time)
- `vite` and `vitest`: for reasons I don't quite understand, the way Yarn resolves `vitest > vite` almost always ends up conflicting with our direct updates to `vite`'s version. I always resolve this in a second step, by:
    1. Delete the redundant entries for `vite` versions at the top level of the lockfile
    2. Re-run `yarn install`
    3. Double check the updated diff to make sure there's only one entry for `vite` again

    I'm not sure how much that conflict matters, but it's almost always the culprit when we get type errors in any given package's `vite.config.ts`/`vitest.config.ts`.
- `primevue`: we haven't updated this to the latest semver-major, and we're tracking it separately
- `vue` (and anything else blocked by updates to it): we currently hard-code this to the version used in @getodk/central-frontend. As mentioned in the meeting today, I'd like to stop doing this when we can prioritize it.
</details>

I think the amount of detail in `yarn.lock` isn't so much about efficiency. I believe it's mostly concerned with reproducibility and security:

- versions of all dependencies (direct and transitive) are pinned at install time, to ensure the exact same packages get installed in any environment, fresh or long-running alike
- all dependencies (direct and transitive) can be checked against their `integrity` hash to ensure the contents that get installed are what they claimed to be (also at install time)

I know the `package-lock.json` format has been revised a couple of times to address similar concerns, but I'm not sure how the details compare.

packages/web-forms/src/components/controls/RankControl.vue
latin-panda: Resolving as it was discussed in today's team meeting, we are leaving the whole row draggable and observing user feedback

packages/scenario/test/rank.test.ts
```diff
@@ -0,0 +1,117 @@
...
+
+				rank(
+					'/data/rankQuestion',
+					"instance('options')/root/item[selected(/data/selectQuestion, value)]"
+				)
+			)
+		);
+	};
+
+	it('should update value when rank has <items>', async () => {
```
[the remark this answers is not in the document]
eyelidlessness: I think it would be helpful for _each_ of these tests to mention in the description that the **order** of values is preserved when set. It's one of the two semantic distinctions from `<select>`. I'd also be happy to see _either_ (XOR):

- Use "update" for a test _specifically exercising reordering_, and avoid that use for tests performing only one write
- Keep "update" in each description, and further emphasize preservation of order by performing multiple writes and asserting the value's order changed to match

packages/scenario/test/rank.test.ts
```diff
@@ -0,0 +1,117 @@
...
+		const scenario = await Scenario.init('Rank form', getRankForm());
+
+		expect(scenario.answerOf(RANK_QUESTION).getValue()).toBe('');
+
+		scenario.answer(RANK_QUESTION, 'option1', 'option4', 'option2', 'option3');
+
+		expect(scenario.answerOf(RANK_QUESTION).getValue()).toBe('option1 option4 option2 option3');
+	});
+
+	it('should filter values when rank has choice-filter and it should update values when options are ranked', async () => {
```
eyelidlessness, line 89: I usually 👀 👀 when I see "and" in a test description. I'm kind of on the fence about whether this might be better as two tests (split on "and").

packages/scenario/test/rank.test.ts
```diff
@@ -0,0 +1,117 @@
...
+		expect(scenario.answerOf(RANK_QUESTION).getValue()).toBe('');
+
+		// Make rank relevant again
+		scenario.answer(SELECT_QUESTION, 'option1', 'option5');
+		expect(scenario.answerOf(SELECT_QUESTION).getValue()).toBe('option1 option5');
+		expect(scenario.answerOf(RANK_QUESTION).getValue()).toBe('option1 option5');
+
+		scenario.answer(RANK_QUESTION, 'option5', 'option1');
+		expect(scenario.answerOf(RANK_QUESTION).getValue()).toBe('option5 option1');
+	});
```
eyelidlessness, line 116: I think we should have one more test, exercising the requirement to include all values on write. Along with ordering, this is the other important _semantic_ difference from `<select>`, and I think if nothing else it'll be good to have a test as a defense against regressions. The biggest regression risk that comes to mind is someone (probably me) getting too cute about generalizing the node implementations more than they already are.

eyelidlessness (commented):
Submitting another partial round of feedback, before I run out for a break! One small nit on the changeset language, otherwise I think this covers everything I want to in `scenario`. (Almost all nits there too!)

packages/scenario/test/rank.test.ts
latin-panda: Yes, I mostly agree about `and` and having separate tests, as long as they (`it(...)`) are independent of each other. 
My intention for this test is that the options can be ordered as soon as the rank question becomes relevant, but maybe I didn't explain it well in the description. 

How about if I change the description to: `should order options as soon as the rank question becomes relevant`

packages/scenario/test/rank.test.ts
```diff
@@ -0,0 +1,117 @@
...
+	selectDynamic,
+	t,
+	title,
+} from '@getodk/common/test/fixtures/xform-dsl/index.ts';
+import { describe, it, expect } from 'vitest';
+import { Scenario } from '../src/jr/Scenario.ts';
+import { r } from '../src/jr/resource/ResourcePathHelper.ts';
+
+describe('Rank', () => {
+	const getRankForm = () => {
```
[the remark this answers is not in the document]
latin-panda: 1. The suggestion is about moving the function content to inside/inline the test case (`it(..)`) and making the form DSL, did I understand correctly? 

2. What is the meaning of DSL acronym?

packages/scenario/src/jr/Scenario.ts
```diff
@@ -379,21 +382,30 @@ export class Scenario {
 		return this.setNonTerminalEventPosition(() => index, reference);
 	}
 
-	private answerSelect(reference: string, ...selectionValues: string[]): ValueNodeAnswer {
+	private answerItemCollectionQuestion(
```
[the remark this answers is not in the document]
latin-panda: Yeah, I'm not super happy with the naming, but it seemed like common ground for questions with a list of `<item>` and an `<itemset>`. Also, it aligns with the terms we use in the engine for this work since the scenario is almost like a client of the engine, and it ends up checking that the reactivity is working.

This comes from someone with little background in JR 😅 What is a better name?

.changeset/proud-beans-march.md
```diff
@@ -0,0 +1,8 @@
+---
+'@getodk/xforms-engine': minor
+'@getodk/web-forms': minor
+'@getodk/scenario': minor
+'@getodk/common': patch
+---
+
+Support for `<rank>`
```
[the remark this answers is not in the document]
latin-panda: Done, I went with a combination of both.

packages/common/src/test/fixtures/xform-dsl/index.ts
```diff
@@ -218,6 +218,23 @@ export const proposed_selectDynamic: Proposed_selectDynamic = (
 
 export { proposed_selectDynamic as selectDynamic };
 
+type RankParameters = readonly [ref: string, nodesetRef: string];
+type Rank = (...args: RankParameters) => XFormsElement;
+
+export const rank: Rank = (...[ref, nodesetRef]: RankParameters): XFormsElement => {
```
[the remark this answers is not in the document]
latin-panda: > 1. I'm not sure how realistic it is to backport additive changes from Web Forms in JavaRosa. But when they're introduced we've usually used a pattern of a proposed_ prefix on additive APIs.

Fixed!

> 2. It looks like this is derived from select1Dynamic and/or proposed_selectDynamic. I think it would be best to preserve that naming convention, as a distinction from the static format with inline items (even if we don't add/use it now).

I am not sure if I got the suggestion accurately, but I renamed the function to have `Dynamic` on them.

> 3.  I'm torn on whether we should mimic the signatures of those as prior art, or keep this simpler signature with only the two required parameters. If we prefer the former, I think it makes sense to keep the separate type definitions for consistency. But I don't know how important it is for this to support arbitrary value/label references. If we don't think that's important, I'd rather the simpler signature go inline like most normal functions.

I'd recommend keeping it minimal, with only the necessary things. In the future, we can add more parameters if necessary. I agree with a simpler function and have inlined the types.

packages/scenario/test/rank.test.ts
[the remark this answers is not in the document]
latin-panda: I have moved the content of the functions, which are DSL forms, to inside the test case (`it(..)`). Please let me know if I got the feedback wrong :)

packages/scenario/test/rank.test.ts
```diff
@@ -0,0 +1,117 @@
...
+						item('option3', 'Option 3'),
+						item('option4', 'Option 4')
+					)
+				)
+			),
+			body(rank('/data/rankQuestion', "instance('options')/root/item"))
+		);
+	};
+
+	const getRankWithChoiceFilterForm = () => {
```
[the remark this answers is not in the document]
latin-panda: Fixed

packages/scenario/test/rank.test.ts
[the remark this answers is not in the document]
latin-panda: I missed that typo, thanks! I've updated the descriptions like this:

'should preserve the order of values when set, if the rank contains <item>s'
'should preserve the order of values when set, if the rank contains an <itemset>'

packages/scenario/test/rank.test.ts
latin-panda: For now, I've updated the description to `'should order options correctly when the rank question becomes relevant'`. Please let me know if this is okay :)

packages/scenario/test/rank.test.ts
latin-panda: I have added a test that checks all values are included, even when ordering just a few. 

`should include all values on write and preserve the order of options when rank is relevant`

packages/scenario/test/rank.test.ts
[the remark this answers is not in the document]
eyelidlessness: (I consider this resolved. I'm just leaving the thread open so the explanation doesn't get lost!)

packages/scenario/test/rank.test.ts
eyelidlessness: That makes the intent of the test much clearer!

packages/scenario/test/rank.test.ts
eyelidlessness: Thanks! The test does _almost_ everything I expected:

1. (act phase) attempt to set incomplete set of values
2. (assert phase) ensure the incomplete values were not set

There's something else I expected to see, and was surprised by its absence. At first I thought I had miscommunicated the intent of the test! (And wasted some time writing that up before I looked more closely at the test.)

What I don't see _in the test_ is anything about the expected error (now named `RankMissingValueError` in the engine 👍). I had mixed feelings[^1] about whether it should be one of the test's assertions, but I now think it **definitely** should be asserted, for the test to be complete. The assertion will fail, catching a bug in the current logic. I'll add a comment inline about that, and link back to this comment!

[^1]: Because we're going to change how errors are produced throughout at least the engine, and probably up the stack.

packages/xforms-engine/src/instance/RankControl.ts
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	setValues(valuesInOrder: readonly string[]): Root {
+		if (isBlankValueState(valuesInOrder)) {
+			this.setValueState(valuesInOrder);
+			return this.root;
+		}
+
+		const sourceValues: string[] = Array.from(this.mapOptionsByValue().keys());
+		const hasAllValues = sourceValues.some((sourceValue) => valuesInOrder.includes(sourceValue));
```
eyelidlessness, line 190:
```suggestion
		const hasAllValues = sourceValues.every((sourceValue) => valuesInOrder.includes(sourceValue));
```

Once the [assertion for this check](https://github.com/getodk/web-forms/pull/288#discussion_r1937726138) is added, it'll fail before this change and pass after.

packages/scenario/src/jr/Scenario.ts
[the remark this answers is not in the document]
eyelidlessness: I'm not sure I have a better option in mind, and since Hélène hasn't jumped in with an opinion either, I'm inclined to just not bikeshed this one. It's `private`, good enough for me! If something more JR/`Scenario`-themed strikes me in the future while working on something related, maybe I'll suggest it then!

packages/common/src/test/fixtures/xform-dsl/index.ts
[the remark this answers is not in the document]
eyelidlessness: Yeah, I also prefer the simpler signature unless/until we need to test the behavior of `value`/`label` ref parameters specifically for rank (which I don't expect to do, we know it's handled one way for all `<itemset>`s, and it would be silly to change that after the fact).

I just wanted to make sure we're considering the trade-off of consistency (with JR) vs sensible additive changes!

packages/xforms-engine/src/instance/children.ts
eyelidlessness: Tracked in #296!

packages/web-forms/src/components/controls/RankControl.vue
eyelidlessness: Gonna resolve this now since that's a place to discuss in more detail!

packages/xforms-engine/src/parse/body/control/RankControlDefinition.ts
```diff
@@ -1,27 +1,54 @@
...
+		}
 	}
 
-	override toJSON(): object {
+	override toJSON() {
```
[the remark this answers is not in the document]
eyelidlessness: Gonna resolve this with reservations: _dependencies_ may also call `toJSON`, usually implicitly, e.g. a `JSON.stringify`/`JSON.parse` round trip. I spotted this while exploring some of the changes that landed in [4742935](https://github.com/getodk/web-forms/pull/288/commits/47429356bf0663a0c1fca157eb39e4e70dac0227). Maybe it's good to just leave some of these around as a reminder that we can even just make JSON serialization a no-op when it causes problems in code we don't control. 🙃

packages/web-forms/src/components/controls/RankControl.vue
```diff
@@ -0,0 +1,246 @@
...
+import { computed, inject, ref } from 'vue';
+import { VueDraggable } from 'vue-draggable-plus';
+
+interface RankControlProps {
+	readonly question: RankNode;
+}
+
+interface HighlightOption {
+	index: Ref<number | null>;
+	timeoutID: NodeJS.Timeout | null;
```
eyelidlessness, line 15: **Definitely** something to solve separately, but this is a really good example of how the TypeScript config is broken for Vue. The type for a web environment should be `number | null`, and we shouldn't have Node types in scope here at all. Just making note because I've also run into issues with this.

packages/web-forms/src/components/controls/RankControl.vue
```diff
@@ -0,0 +1,246 @@
...
+}
+
+const props = defineProps<RankControlProps>();
+const HOLD_DELAY = 200; // Delay in ms to hold an item before dragging, avoids accidental reordering on swipe.
```
eyelidlessness, line 19: No need for change now unless you feel like it, but I'd usually make this a JSDoc comment so it's also available for references inline.

packages/web-forms/src/components/controls/RankControl.vue
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	setHighlight(index);
+
+	const swappedValues = values.value.slice();
+	const movedValue = swappedValues[index];
+
+	swappedValues[index] = swappedValues[newPosition];
+	swappedValues[newPosition] = movedValue;
+	values.value = swappedValues;
```
eyelidlessness, line 85: This came out of some suggestions I made in an exploratory commit. Noting it because, if we find another use case, I'll want to extract a `@getodk/common` function for moving an array item to a different index. It would likely look more like the existing `insertAtIndex`, and probably call into it.

packages/web-forms/src/components/controls/RankControl.vue
```diff
@@ -0,0 +1,246 @@
...
+.disabled .rank-option,
+.disabled button {
+	cursor: not-allowed;
+}
+
+@media screen and (max-width: #{$sm}) {
+	.rank-buttons {
+		display: none;
+	}
+}
```
eyelidlessness, line 245: Not sure about this! The most likely way a user would experience it is on a phone. I think the buttons are especially convenient for a touch interface (where drag/drop can be really awkward, even with finesse like an appropriate delay). I am happy to defer this for now, since it'll probably need more UI/UX design attention. But wanted to note that it feels like a gap to me.

packages/scenario/test/rank.test.ts
```diff
@@ -0,0 +1,165 @@
...
+		);
+		expect(scenario.answerOf(SELECT_QUESTION).getValue()).toBe(
+			'option1 option2 option3 option4 option5 option6'
+		);
+		expect(scenario.answerOf(RANK_QUESTION).getValue()).toBe('');
+
+		scenario.answer(RANK_QUESTION, 'option6 option2 option5');
+		expect(scenario.answerOf(RANK_QUESTION).getValue()).toBe(
+			'option6 option2 option5 option1 option3 option4'
+		);
```
eyelidlessness, line 163:
```suggestion
		// (Precondition) Ensure rank question has a value, to be checked
		// before/after attempting to set incomplete value.
		scenario.answer(
			RANK_QUESTION,
			'option1',
			'option4',
			'option3',
			'option6',
			'option5',
			'option2'
		);

		// Expected value of precondition
		const EXPECTED_VALUE = 'option1 option4 option3 option6 option5 option2';

		// (Sanity check) assert setting precondition value was effective
		expect(scenario.answerOf(RANK_QUESTION).getValue()).toBe(EXPECTED_VALUE);

		// Set an incomplete value is not allowed for rank. This is wrapped as a
		// function so we can assert that this is enforced by the engine.
		const answerWithMissingValues = () => {
			scenario.answer(RANK_QUESTION, 'option6', 'option2', 'option5');
		};

		// Assert: engine enforces requirement to set all available rank values.
		expect(answerWithMissingValues).toThrowError();

		// Assert: attempting failed incomplete answer doesn't change value state.
		expect(scenario.answerOf(RANK_QUESTION).getValue()).toBe(EXPECTED_VALUE);
```

Feel free to remove/rewrite comments as you see fit!

packages/scenario/test/rank.test.ts
eyelidlessness: I'm gonna mark this resolved because it is now in the wrong place for an inline comment, and I have a clearer explanation/suggestion incoming.

eyelidlessness (commented):
I think this wraps up my feedback. Great work @latin-panda! 🎉 

I think the only things outstanding are the fix for `hasAllValues`, and ensuring the test exercising it gets updated to account for the error it'll throw. Otherwise LGTM! I'm excited to get this in.

packages/web-forms/src/components/controls/RankControl.vue
latin-panda: The reason @alyblenkin and I removed the buttons on mobile devices was to create more space for text and reduce excessive wrapping, which makes reading harder. Buttons take up too much screen space.

packages/web-forms/src/components/controls/RankControl.vue
latin-panda: I've converted it to JS Doc, and it shows in the IDE as info for that constant 
:) thanks for that!
Screenshot 2025-01-31 at 2 35 05 PM

packages/web-forms/src/components/controls/RankControl.vue
latin-panda: Yes! Agree 100%. I tried other types, but I was still getting the errors

eyelidlessness:
Just did a quick manual test on my iPhone. Nothing concerning on Safari (iOS)!

packages/web-forms/src/components/controls/RankControl.vue
eyelidlessness: Yeah. The config and tooling around Vue is kind of a mess. I've come back to it a few times, but always end up frustrated. Last time I looked, there were a lot of improvements in flight on the tooling front, so maybe I'll take another look at it again soon.

packages/web-forms/src/components/controls/RankControl.vue
```diff
@@ -0,0 +1,246 @@
...
+const setHighlight = (index: number | null) => {
+	highlight.index.value = index;
+
+	if (highlight.timeoutID) {
+		clearTimeout(highlight.timeoutID);
+		highlight.timeoutID = null;
+	}
+
+	if (highlight.index.value !== null) {
+		highlight.timeoutID = setTimeout(() => setHighlight(null), 1000);
```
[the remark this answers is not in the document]
latin-panda: > This delay feels longer than I'd expect. And I think tweaking it might be aided by making it a SHOUTY_CONSTANT like HOLD_DELAY is.

I've added constant with JS Docs, and made it 500ms

> There's a pixel shift while the highlight is applied, because the border width changes. I expect we'll make other presentation changes, so I'm not too concerned about it for now.

With the new Figma style, this is not a problem. It's temporary. For now, it looks like the highlight from select, which is a border of 2px

packages/web-forms/src/components/controls/RankControl.vue
latin-panda: Oh! I need to explore those libs. I'll keep it in mind next time, for now, leaving here.

packages/scenario/test/rank.test.ts
latin-panda: I kept some of the comments

packages/xforms-engine/src/instance/RankControl.ts
latin-panda: Done

eyelidlessness (approved):
🎉 Awesome awesome work @latin-panda! Feel free to merge once CI passes 🚀

latin-panda:
Thank you for reviewing @eyelidlessness

PR #291, opened 2025-01-24
Emit submission payload(s) to host app

### LINKED ISSUE #23: Provide filled instance to host application for submission
## User Stories
As an application developer
I want to get access to filled forms from web forms
So that I can submit them to a server or otherwise process them

As a form end user
If my submission fails, I want to know what happened (e.g. I'm offline, the server isn't responding, etc)
So that I can take appropriate action

As a form end user
When my auth to the host application has expired when I try to submit
I want to have an opportunity to log in again
So that I can complete submission

### LINKED ISSUE #290: Make instance (+attachments) available to a host application
Towards #23

For a host application, the instance (+ attachments) produced by Webforms needs to be available, somehow.
This is a stub issue for PR #289 .

[@eyelidlessness wants to make sure that the event payloads have useful type information.](https://getodk.slack.com/archives/C051L6AKG4F/p1737574043113399?thread_ts=1737573679.891789&cid=C051L6AKG4F)

eyelidlessness:
Supersedes and closes #289. Closes #23. Closes #290.

- [x] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Not applicable

It would shock me if there's anything browser-specific here.

Types do much of the work. Added console logs in `FormPreview.vue` demonstrate both handlers working.

I mentioned this in Slack, and will reiterate here: in hindsight, since there are two distinct `SubmissionResult` payload formats, events may not be the best API for this. On the other hand, it may be possible to improve the API by:

- Making `submissionMaxSize` generic at the component/package boundary
- Restoring the single `submit` event emit, implicitly branching the payload type so it's chunked if `submissionMaxSize` is non-nullish, monolithic otherwise

This change already felt big enough (especially compared to the #289 1-liner), so I didn't want to hold it up to explore that. Especially because I'm already kind of uncomfortable with the implicit coupling between a prop and emit, and I'm not sure branching on the prop would ameliorate that!

I can't think of anything beyond... this exposes existing functionality to a host application, along with any known or unknown bugs that may already be present.

N/A

## What's changed

The main `@getodk/web-forms` component now:

- Emits a `MonolithicSubmissionResult` if a host application attaches a `submit` event handler
- Adds an optional `submissionMaxSize` prop
- Emits a `ChunkedSubmissionResult` if a host application specifies `submissionMaxSize` and attaches a `submitChunked` event handler

### DIFF
--- a/packages/web-forms/src/demo/FormPreview.vue
+++ b/packages/web-forms/src/demo/FormPreview.vue
@@ -1,6 +1,11 @@
 <script setup lang="ts">
 import { xformFixturesByCategory, XFormResource } from '@getodk/common/fixtures/xforms.ts';
-import type { FetchFormAttachment, MissingResourceBehavior } from '@getodk/xforms-engine';
+import type {
+	ChunkedSubmissionResult,
+	FetchFormAttachment,
+	MissingResourceBehavior,
+	MonolithicSubmissionResult,
+} from '@getodk/xforms-engine';
 import { constants as ENGINE_CONSTANTS } from '@getodk/xforms-engine';
 import { ref } from 'vue';
 import { useRoute } from 'vue-router';
@@ -50,8 +55,16 @@ xformResource
 		alert('Failed to load the Form XML');
 	});
 
-const handleSubmit = () => {
+const handleSubmit = (payload: MonolithicSubmissionResult) => {
+	// eslint-disable-next-line no-console
+	console.log('submission payload:', payload);
+
 	alert(`Submit button was pressed`);
+};
+
+const handleSubmitChunked = (payload: ChunkedSubmissionResult) => {
+	// eslint-disable-next-line no-console
+	console.log('CHUNKED submission payload:', payload);
 };
 </script>
 <template>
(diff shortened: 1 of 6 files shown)


packages/web-forms/src/demo/FormPreview.vue
```diff
@@ -50,17 +55,27 @@ xformResource
...
+	console.log('CHUNKED submission payload:', payload);
+};
 </script>
 <template>
 	<template v-if="formPreviewState">
 		<OdkWebForm
 			:form-xml="formPreviewState.formXML"
 			:fetch-form-attachment="formPreviewState.fetchFormAttachment"
 			:missing-resource-behavior="formPreviewState.missingResourceBehavior"
+			:submission-max-size="Infinity"
```
latin-panda, line 76: Question:

Is there any limit in some other place, like Central dropping the connection or rate limit, after a while of long or many chucks? Asking in case this is a configuration setting.

latin-panda (approved):
It looks good to me! Just one question below but other than that, it's approved.

packages/web-forms/src/demo/FormPreview.vue
latin-panda: > but I'd also be happy to put that off until we have support for submission attachments, where we'll actually get chunked submissions working 

Sounds good! Thank you

PR #298, opened 2025-02-04
Variadic `distance`, minor related XPath geo functionality improvements

### LINKED ISSUE #118: Add multiple-argument distance variant

eyelidlessness:
Closes #118. Supersedes and closes #244.

**Please don't pay too much attention to the commit history of this branch!** It's... well, it's bananas.

I expect we'll want to squash this on merge.

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [x] Not applicable

I don't **think** there should be anything browser specific that wouldn't be caught by CI running tests across targeted browsers. Happy to manually verify if we feel otherwise in review!

- Incorporated test coverage from getodk/javarosa#761
- Added a test to clarify multi-argument type semantics
- Updated `@getodk/xpath` tests where appropriate
- Updated previously failing scenario tests to reflect related improvements

This expands on the _logical changes_ in #244 with:

- a minor semantic correction, for the distinction between single-argument and variadic calls
- an incremental effort to isolate decoding logic for geographic data types, which will be pertinent for pending work to support those data types in the engine and UI
- fixes a couple of convenient related issues while working in the area

It's possible that the existing test coverage didn't exercise some previous semantic nuance that's been lost, especially in extracting the decoding logic.

Attached.

### Feature: variadic `distance`

The primary purpose of this change is to support the spec expansion of the [`distance`](https://getodk.github.io/xforms-spec/#fn:distance) function, matching getodk/javarosa#761.

### Geo* groundwork

After a brief discussion, we decided it would be worthwhile to expand the scope of this change to provide some early support for upcoming feature work to support the `geopoint` data type (with `geotrace` and `geoshape` presumably soon to follow).

**Why:** The existing implementation of `area` and `distance` include logic to **decode** serialized `geopoint` values, and semicolon-separated lists thereof. This logic is quite likely going to be _very similar if not identical_ to the logic we'd use in incoming engine implementations of the geographic `ValueCodec`s (e.g. `GeopointCodec`, `GeotraceCodec`, `GeoshapeCodec`).

The change largely leaves the `xpath`-internal representations of these data types unchanged (other than making them a bit more formal, as classes).

**Note:** I don't necessarily expect these interfaces (as representations of each data/value type) to be stable or complete! In part, that's because we have corresponding **encode** behavior in the engine which could also probably be incorporated into said value codecs, in the parse-stage serialization of GeoJSON values, from [GeoJSON external secondary instances](https://github.com/getodk/web-forms/blob/81a57c3d4f5c1b511ccb6b2e0333a3767dbe7019/packages/xforms-engine/src/parse/model/SecondaryInstance/sources/GeoJSONExternalSecondaryInstance.ts).

This change is intended to call out where existing decode logic exists since I was already working on it, and to set up some WIP structures that look roughly how they might in a proper engine `ValueCodec`. But that leaves a big open question: how do we share logic between `@getodk/xpath` and `@getodk/xforms-engine`? To be determined! That's definitely out of scope for this PR.

### Related fixes

This change also addresses a couple of fixes/improvements in consistency with JavaRosa's `distance` _and `area`_ implementations:

- Trailing semicolons are now supported in `geotrace`/`geoshape` values (both of which `@getodk/xpath` really only knows about as a serialized list of `geopoint` values)
- Errors are produced for invalid arguments to `distance` (I'm reasonably certain the conditions are now consistent with JavaRosa)
- `area` now returns `0` instead of `NaN` for invalid input (definitely consistent with JavaRosa; breaking previous consistency with Enketo, as demonstrated by changes to tests inherited from `openrosa-xpath-evaluator`)

### DIFF
--- a/packages/xpath/src/functions/xforms/geo.ts
+++ b/packages/xpath/src/functions/xforms/geo.ts
@@ -1,6 +1,7 @@
 import { pairwise } from 'itertools-ts/lib/single';
 import type { XPathNode } from '../../adapter/interface/XPathNode.ts';
 import { EvaluationContext } from '../../context/EvaluationContext.ts';
+import { JRCompatibleError } from '../../error/JRCompatibleError.ts';
 import type { EvaluableArgument } from '../../evaluator/functions/FunctionImplementation.ts';
 import { NumberFunction } from '../../evaluator/functions/NumberFunction.ts';
 
@@ -213,7 +214,9 @@ export const distance = new NumberFunction(
 		const lines = evaluateLines(context, args);
 
 		if (lines.some(isInvalidLine)) {
-			return NaN;
+			throw new JRCompatibleError(
+				"The function 'distance' received a value that does not represent GPS coordinates"
+			);
 		}
 
 		const distances = lines.map(geodesicDistance);
--- a/packages/xpath/src/functions/xforms/geo.ts
+++ b/packages/xpath/src/functions/xforms/geo.ts
@@ -84,10 +84,9 @@ const INVALID_LINE: Line = {
 
 const evaluateLines = <T extends XPathNode>(
 	context: EvaluationContext<T>,
-	expression: EvaluableArgument
+	expression: readonly EvaluableArgument[]
 ): Line[] => {
-	const points = evaluatePoints(context, expression);
-
+	const points = expression.flatMap((el) => evaluatePoints(context, el));
 	if (points.length < 2) {
 		return [INVALID_LINE];
 	}
@@ -172,7 +171,7 @@ export const area = new NumberFunction(
 	'area',
 	[{ arityType: 'required' }],
 	(context, [expression]) => {
-		const lines = evaluateLines(context, expression!);
+		const lines = evaluateLines(context, [expression!]);
 
 		if (lines.some(isInvalidLine)) {
 			return NaN;
@@ -209,9 +208,9 @@ const sum = (values: readonly number[]) => {
 
 export const distance = new NumberFunction(
 	'distance',
-	[{ arityType: 'required' }],
-	(context, [expression]) => {
-		const lines = evaluateLines(context, expression!);
+	[{ arityType: 'variadic' }],
+	(context, args) => {
+		const lines = evaluateLines(context, args);
 
 		if (lines.some(isInvalidLine)) {
 			return NaN;
--- a/packages/xpath/src/functions/xforms/geo.ts
+++ b/packages/xpath/src/functions/xforms/geo.ts
@@ -1,53 +1,9 @@
-import { pairwise } from 'itertools-ts/lib/single';
 import type { XPathNode } from '../../adapter/interface/XPathNode.ts';
 import { EvaluationContext } from '../../context/EvaluationContext.ts';
-import { JRCompatibleGeoValueError } from '../../error/JRCompatibleGeoValueError.ts';
 import type { EvaluableArgument } from '../../evaluator/functions/FunctionImplementation.ts';
 import { NumberFunction } from '../../evaluator/functions/NumberFunction.ts';
-import type { GeopointCoordinates } from '../../lib/geo/Geopoint.ts';
-import { geopointCodec } from '../../lib/geo/geopointCodec.ts';
-
-const evaluatePoints = <T extends XPathNode>(
-	context: EvaluationContext<T>,
-	expression: EvaluableArgument
-): GeopointCoordinates[] => {
-	const results = expression.evaluate(context);
-
-	const stringResults = [...results].map((result) => result.toString());
-	const geopointStrings = stringResults.flatMap((result) => {
-		const string = result.toString().trim();
-
-		return string.split(/\s*;\s*/);
-	});
-
-	return geopointStrings.map((string) => {
-		return geopointCodec.decodeValue(string);
-	});
-};
-
-interface Line {
-	readonly start: GeopointCoordinates;
-	readonly end: GeopointCoordinates;
-}
-
-const evaluateLines = <T extends XPathNode>(
-	context: EvaluationContext<T>,
-	expression: readonly EvaluableArgument[]
-): Line[] => {
-	const points = expression.flatMap((el) => evaluatePoints(context, el));
-	if (points.length < 2) {
-		throw new JRCompatibleGeoValueError();
-	}
-
-	return Array.from(pairwise(points)).map((line) => {
-		const [start, end] = line;
-
-		return {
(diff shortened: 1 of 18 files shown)


packages/xpath/src/functions/xforms/geo.ts
```diff
@@ -209,12 +100,13 @@ const sum = (values: readonly number[]) => {
...
-		const lines = evaluateLines(context, expression!);
+	[{ arityType: 'required' }, { arityType: 'variadic' }],
+	(context, args) => {
+		const values = evaluateArgumentValues(context, args);
+		const lines = Geotrace.fromEncodedValues(values)?.lines;
 
-		if (lines.some(isInvalidLine)) {
-			return NaN;
+		if (lines == null) {
+			throw new JRCompatibleGeoValueError('distance');
```
eyelidlessness, line 109: **Sigh**. Apparently it's more nuanced than this in JavaRosa! From what I can tell based on my understanding of outstanding `scenario` test failures, and playing around a bit with those same tests in JR:

- This error is expected _when an individual point is malformed_
- The function is expected to return `0` when there are not enough points to compute a distance

This is a mistake in logic in `Geotrace`, and it strongly suggests that `Geotrace` is doing too much work here. Which in turn implicates sharing some of the same logic with `area`.

packages/xpath/src/lib/geo/Geopoint.ts
```diff
@@ -0,0 +1,105 @@
...
+			return null;
+		}
+
+		return new this(coordinates);
+	}
+
+	readonly latitude: number;
+	readonly longitude: number;
+
+	private constructor(coordinates: GeopointCoordinates) {
```
latin-panda, line 101: Question:

Will this decode [the altitude and accuracy](https://getodk.github.io/xforms-spec/#data-types) in this file too?

packages/xpath/src/lib/geo/Geotrace.ts
```diff
@@ -0,0 +1,82 @@
...
+		);
+
+		return acc;
+	}, Array<GeotraceLine>());
+};
+
+export class Geotrace {
+	static fromEncodedGeotrace(encoded: string): Geotrace | null {
+		const geopoints = encoded
+			.replace(/\s*;\s*$/, '')
```
eyelidlessness, line 42: This is _still_ inconsistent with JavaRosa! After playing around with it a bit more, it turns out we should ignore any number of trailing semicolons, not just one!

I did a quick search for the actual code handling this detail, but I don't see it. I strongly suspect there's some **Java nuance** at play, but I couldn't say for sure at my current familiarity with the language.

packages/xpath/src/lib/geo/Geotrace.ts
```diff
@@ -0,0 +1,82 @@
...
+				return geopointCodec.decodeValue(value);
+			});
+
+		return this.fromGeopoints(geopoints);
+	}
+
+	static fromEncodedValues(values: readonly string[]): Geotrace | null {
+		const [head, ...tail] = values;
+
+		if (head == null) {
```
latin-panda, line 54: Question:

[In the docs](https://getodk.github.io/xforms-spec/#data-types), it says, `where the last geopoint’s latitude and longitude is not equal to the first`

Is that a validation that can be done here or better at the engine level? I'm thinking at engine level so it throws an exception, but I thought of double-checking

latin-panda (approved):
Thank you for sending this! I was easy to review, just some questions below but approving to unblock

packages/xpath/src/lib/geo/Geotrace.ts
latin-panda: Maybe something like `.replace(/^[\s;]+|[\s;]+$/g, '')`

But it'll remove spaces and semicolon at the beginning and end of the string, it might need a different one for the empty middle sections

packages/xpath/src/lib/geo/Geopoint.ts
eyelidlessness: Currently, it acknowledges that they exist (see [here](https://github.com/getodk/web-forms/pull/298/files/48cb3c9f7947d253b8f3ad9e4953403b051cf94e#diff-33969c65f71b89c8518951cb767745b11df1b91e2bfc3249108e84f9946b469fR5-R6) and [here](https://github.com/getodk/web-forms/pull/298/files/48cb3c9f7947d253b8f3ad9e4953403b051cf94e#diff-33969c65f71b89c8518951cb767745b11df1b91e2bfc3249108e84f9946b469fR16)), but ignores them ([here](https://github.com/getodk/web-forms/pull/298/files/48cb3c9f7947d253b8f3ad9e4953403b051cf94e#diff-33969c65f71b89c8518951cb767745b11df1b91e2bfc3249108e84f9946b469fR54)). The intent in breaking it out now is to lay groundwork for expanding the full decoding logic, but defers those aspects since the XPath logic doesn't use them.

packages/xpath/src/lib/geo/Geotrace.ts
latin-panda: OK. Even if the last and first are equal and there are no more points in between (so the array length is 2), it may be a valid value of something that didn't move 🤔

packages/xpath/src/functions/xforms/geo.ts
eyelidlessness: I think we should pay close attention to this remaining WF/JR discrepancy. It raises questions about whether and how we should share decoding logic between the XPath/engine packages. With the logic directly in the XPath package, we can tailor that logic to address the nuance however we see fit. If it's shared (especially with a `ValueCodec` API), we'll need to special case it one way or another, e.g.:

- Introduce some intermediate value representation which is almost a geotrace but not quite
- Expose the intermediate decoding step in some other way, so the XPath implementation can reason about failure to decode points separately from success/failure to decode a list

I stopped short of fixing this now specifically because I want to consider how it intersects with immediate downstream work.

packages/xpath/src/lib/geo/Geotrace.ts
eyelidlessness: Yeah. And as for how it affects XPath behavior, both `area` and `distance` will return `0` for two equal points.

packages/xpath/src/functions/xforms/geo.ts
latin-panda: It's okay to make improvements in another iteration if it's big or needs more information from future usages. We can leave the comment open for easier reference later.

Issue #296, opened 2025-01-31
How should we represent enumerated types?


eyelidlessness:
Stub issue, will fill out more detail as we prioritize.

Quick list of alternatives (probably not exhaustive):

1. TypeScript `enum`. Risky for a variety of reasons, which are [only growing](https://devblogs.microsoft.com/typescript/announcing-typescript-5-8-beta/#the---erasablesyntaxonly-option).
2. Bare union type (e.g. `'foo' | 'bar' | ...`)
3. Something `enum`-like, derived from plain JavaScript values ([some examples](https://github.com/getodk/web-forms/pull/288#discussion_r1931023931))
4. ?

PR #300, opened 2025-02-06
feat(#249): Adds default geopoint question type

### LINKED ISSUE #249: Form filling: Geopoint with no appearance
As a data collector or someone doing self report 
I want to **get my location (without a map)**
so that I can collaborate with my PM to ensure I have high accuracy 

Given I've allowed Web Forms to have my location
When I've select the button "get location"
Then I see a dialog that says "Finding your location" with an accuracy message below 

**Happy path**
Given I want to save the location
When I see I have "Good accuracy" message in blue 
Then I will select the button "save location" 

Given I've saved the location
When I'm looking at my form
Then I see the question with an accuracy message 
And I can select the button "remove" it if I want to try again and get higher accuracy 

**Sad path**
Given I want to save the location despite having poor accuracy 
When I see the "Poor accuracy" warning message in red
Then I will select the button "save location" 

Given I've saved the location with poor accuracy
When I'm looking at my form
Then I see the question with an accuracy message in red
And I can select the button "remove" it if I want to try again and get higher accuracy 

## notes
Need to decide on the button language to remove saved location

**Options:**

- Clear location
- Delete location 
- Remove or delete 
- Change location 
- Refresh

latin-panda:
Closes #249

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [x] Safari (iOS) (@lognaturel confirmed that it's working)
- [x] Chrome for Android (Need to serve using HTTPS to test, I used ngrok)
- [ ] Not applicable

[The test plan is here](https://github.com/getodk/web-forms/issues/249#issuecomment-2635178304)
[The e2e tests are here](https://github.com/getodk/web-forms/pull/313)

Test videos: 

<details><summary> Test case: Responsive design and capturing good accuracy</summary>

https://github.com/user-attachments/assets/00225d98-f6da-4bc7-9465-af6882a5701a
</details>

<details><summary> Test case: Responsive design and capturing poor accuracy </summary>

https://github.com/user-attachments/assets/dbade800-f9a9-496e-9f48-22927e810de8

</details>

<details><summary> Test case: Having multiple fields and retry capturing location </summary>

https://github.com/user-attachments/assets/501dab57-76e3-4eb1-8756-f716cc409226

</details>

<details><summary> Test case: Autosave and no autosave </summary>

https://github.com/user-attachments/assets/0031e1b5-7330-4833-8237-331c59e12385

</details>

<details><summary> Test case: Permission denied and required field</summary>

https://github.com/user-attachments/assets/fdbb6229-fa3b-4da7-aea9-35a9bec02159

</details>

<details><summary> Test case: Field becoming relevant and disabled </summary>

https://github.com/user-attachments/assets/c6e66c37-d964-4f1e-8a8d-164a938b1e1a

</details>

<details><summary> Test case: reacts to form translation and Chrome, Firefox and Safari </summary>

Geopoint's text translation isn't supported yet.

https://github.com/user-attachments/assets/e5e8b3ea-adb3-46a8-8e65-23c4a13c20c5

</details>

<details><summary> Test case: Geopoint note type </summary>

https://github.com/user-attachments/assets/4fabc796-834d-4e69-8ed5-66f1fb9ebed0

</details>

<details><summary> Test case: Android phone </summary>

https://github.com/user-attachments/assets/6a26bdf4-6ccc-4933-ad46-c9fb89c4ab36

Edit [27 Feb]: We changed to 3 decimals maximum and made a space between the number and the meter unit. Screenshot shows update:

</details>

- A separate component in the Web Forms client now handles the readonly representation of Geopoint data. This component is reused in both Geopoint Note and Geopoint Input, ensuring consistent display of the same data.
- The Geopoint Codec now provides more robust support for Geopoint data tuples with type checking, that clarifies the expected representation of this data.

This work should support the default behavior of Geopoint input and no regression are expected. 

The test forms are at the end [of this comment](https://github.com/getodk/web-forms/issues/249#issuecomment-2635178304)

## What's changed

- Adds a demo form for geopoint: 1-geopoint.xml
- Adds a geopoint note in the note's demo form: 2-all-possible-notes.xml
- Adds new scenario tests for geopoint input (bind-types.test.ts), there are already some scenario smoke tests working with geopoint.
- Adds a geopoint input and geopoint readonly components in web-forms (vue client) 
- Registers geopoint in the input node and in note node 
- Adds a geopoint codec 
- Adds node options concept for question attributes - see input control


packages/xforms-engine/src/parse/body/control/InputControlDefinition.ts
```diff
@@ -11,10 +11,16 @@ export class InputControlDefinition extends ControlDefinition<'input'> {
...
+	readonly accuracyThreshold: number | null;
+	readonly unacceptableAccuracyThreshold: number | null;
 
 	constructor(form: XFormDefinition, parent: BodyElementParentContext, element: Element) {
 		super(form, parent, element);
 
 		this.appearances = inputAppearanceParser.parseFrom(element, 'appearance');
+		this.rows = Number(element.getAttribute('rows')); // ToDo: Make parser function, validate NaN
+		this.accuracyThreshold = Number(element.getAttribute('accuracyThreshold')); // ToDo: Make parser function, validate NaN
+		this.unacceptableAccuracyThreshold = Number(element.getAttribute('unacceptableAccuracyThreshold')); // ToDo: Make parser function, validate NaN
```
eyelidlessness, line 24: Not suggesting a change this early on. Just adding a minor "yes and" on these TODO comments for additional context. There are two gaps here, and they apply in a few other places too:

- `accuracyThreshold` and `unacceptableAccuracyThreshold` are specified as decimals. We'll want to make sure that whatever parsing we do accounts for that (including potential float rounding errors).
- `rows` doesn't specify its type explicitly, but I think it's safe to assume it should be an integer. Same thing about parsing the type, different type to account for.

I wish we'd been tracking these cases sooner, because like I said there are others!

packages/xforms-engine/src/parse/body/control/InputControlDefinition.ts
```diff
@@ -11,10 +12,18 @@ export class InputControlDefinition extends ControlDefinition<'input'> {
 
 	readonly type = 'input';
 	readonly appearances: InputAppearanceDefinition;
+	readonly rows: number | null;
```
latin-panda, line 15: The `rows` implementation will be completed in this issue: https://github.com/getodk/web-forms/issues/301

packages/web-forms/src/components/controls/Input/InputGeopoint.vue
```diff
@@ -0,0 +1,392 @@
...
+	stop();
+
+	if (!coords.value) {
+		return;
+	}
+
+	const { latitude, longitude, altitude, accuracy } = coords.value;
+	// ToDo: if one is missing, it still need to know the position of the others.
+	// ToDo: Change the value type to object? or default to negative / zero number?
+	props.node.setValue([latitude ?? 0, longitude ?? 0, altitude ?? 0, accuracy ?? 0].join(' '));
```
latin-panda, line 158: @eyelidlessness I'd like to know your opinion.

**First question:**
The order of these values in the string [is essential for locating them correctly](https://github.com/getodk/web-forms/pull/300/files#diff-ef3a1c07c25fb7b323a6f67771bcd1646917b62fd662fbb1ec8e354b259a4302R27), avoiding confusion between `altitude` and `accuracy`.
I noticed that `altitude` isn't always available (e.g., when testing with WiFi on a desktop). To preserve the position, what do you think about setting a default value of zero or null? 
The other option is to assume that if only three values come in the string, the third one is `accuracy` and assume that `altitude` wasn't provided 🙁

**Second question:**
We need to create a new codec that builds on the one in XPath you just merged and call it in [getSharedValueCodec](https://github.com/getodk/web-forms/blob/d55f7ee715d295946d62d72aceba85b57f7e5406/packages/xforms-engine/src/lib/codecs/getSharedValueCodec.ts#L67), correct? It looks like ValueTypePlaceholderCodec doesn’t handle this automatically and is just a placeholder.

latin-panda: > First question.....

All the JR imported tests in the scenario suite use a string with the four values (latitude, longitude, altitude, accuracy), [and zero is used. ](https://github.com/getodk/web-forms/blob/d55f7ee715d295946d62d72aceba85b57f7e5406/packages/scenario/test/data-types/geo.test.ts#L75)

Maybe zero is the way to go 🤔; I haven't found anything explicit about this in the documentation of XForms

eyelidlessness: I think it might be worth taking a step back. The intent of the engine/client interface—in this case `InputNode`—is to provide a representation of an `<input>` (in ODK XForms terms) which is most useful for clients, e.g. to provide a UI. The intent of `ValueCodec` is to decode readable values in that representation (i.e. `InputNode.currentState.value`), and to encode that representation of a value back to a "node value" (for submission, as serialized XML; for evaluation in XPath expressions).

It's generally up to us what that representation will be. So far, we've usually determined that by choosing a platform type with the closest semantics to the ODK XForms type: e.g. `int` <-> `bigint`, `decimal` <-> `number`.

There's an added wrinkle in this case, because there are currently two representations of the same type (`geopoint`) of value:

- In `@getodk/xpath`, decoding a `geopoint` node value as an object which is pretty much a subset of the web standard [`GeolocationCoordinates`](https://developer.mozilla.org/en-US/docs/Web/API/GeolocationCoordinates)
- In `@getodk/xforms-engine`, encoding a `geopoint` node value from a [GeoJSON Point](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.2) (see also the semi-official [type definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c20bcfd9a030e8f2329de6bed554ddc05e6ca995/types/geojson/index.d.ts#L113))

If we're going to share the encode/decode logic for these existing cases, it seems pretty likely (unless we have a good reason otherwise) that we're going to want a representation of the value which is _either_:

1. One of the two we already use
2. A combined interface which is a superset of both
3. Some other representation entirely???

Option 1 pretty much designs itself:

```ts
import { Point as GeoJSONPoint } from 'geojson';

interface GeopointRuntimeValue {
  readonly latitude: number;
  readonly longitude: number;
  readonly altitude: number; // Or `number | null`?
  readonly accuracy: number; // Or `number | null`?
}

type GeopointInputValue =
  | GepointRuntimeValue
  | GeoJSONPoint;

class GeopointValueCodec extends ValueCodec<GepointRuntimeValue, GeopointInputValue> {
  /**
   * @todo Implementation will:
   *
   * 1. Decode `GeopointRuntimeValue`, pretty much reusing the logic in `@getodk/xpath`
   * 2. Encode `GeopointInputValue` by detecting whether input is a `GeoJSON` point or a `GeopointRuntimeValue`, and then:
   *     - if the former, reuse logic for GeoJSON external secondary instances
   *     - if the latter, invert the decode logic
   */
}
```

Coming back to the question of what to do when `altitude` isn't available, we can make it nullable (like it is in the web standard `GeolocationCoordinates`), or default to `0` ... _in the runtime representation_ (`GeopointRuntimeValue`). There's trade-offs either way: nullable is closer to the platform interface, but we can't distinguish a node value's `0` as missing or sea level. In either case, we have to _encode_ a missing value as `0` (as far as I know, the `altitude` substring isn't optional).

Option 2 (combined superset interface) would make decode a little more complex (assigning the same sub-values each in two places), and simplify the encode logic (only need to reuse logic from GeoJSON external secondary instances).

Option 3 (a completely new representation)... I'd be open to it, but I would want to justify it by referencing something that's successful in projects with strong geo background. But my gut tells me that most representations of a `geopoint` are going to look very similar to one of the representations we already use.

- - -

In any case, I'd like to see if we can settle this aspect of the design before going too much further with the _client integration_. Whatever the representation we end up with, I don't think that we want the client to be responsible for decoding or encoding node values, since that's an engine responsibility everywhere else.

latin-panda: Thank you for the chat easier! This is a quick summary of the plan:

First round:
Build a normal geopoint's engine codec, in which runtime value representation is the shape of  [GeolocationCoordinates](https://developer.mozilla.org/en-US/docs/Web/API/GeolocationCoordinates)

Second round:
Evaluate the codec and see how to fit GeoJSON; maybe external secondary serialization to use the same encoding logic.

Third round:
Evaluate sharing between xpath and xform-engine packages.

After evaluating the second or third, we might decide that these are out of scope and create GH issues for those.

packages/web-forms/tests/components/controls/Input/InputGeopoint.ts
latin-panda, line 1: Work in progress

packages/xforms-engine/src/instance/InputControl.ts
```diff
@@ -61,6 +98,7 @@ export class InputControl<V extends ValueType = ValueType>
 	// InputNode
 	readonly nodeType = 'input';
 	readonly appearances: InputNodeAppearances;
+	readonly nodeOptions: InputNodeOptions<V>;
```
latin-panda, line 101: I haven't decided yet if I'll add this to BaseNode in this PR on the following one. If it's >20 file changes, I'll open a separate PR

packages/scenario/test/bind-types.test.ts
```diff
@@ -480,6 +512,68 @@ describe('Data (<bind type>) type support', () => {
...
+				'-15.2936673 120.7260063 ABC $$',
+				'-2.33373 36.7260063 ABC 15',
+				'20.2936673 -16.7260063 1200 ABCD',
+				'99 179.99999 1200 0',
+				'89.999 180.1111 1300 0',
+			])('has an null when incorrect value is passed', (expression) => {
+				scenario.answer('/root/geopoint-value', expression);
+				answer = getTypedInputNodeAnswer('/root/geopoint-value', 'geopoint');
+				expect(answer.value).toBeNull();
+			});
```
eyelidlessness, line 575: I think this suite is probably the ideal place to also exercise:

- `value` after a write setting:
    - `accuracy: null`
    - `altitude: null`
    - `accuracy: null, altitude: null`
- `value` when a form-defined default and/or `calculate` produces those same cases
- Serialization/encoding of those same cases. I think for this, you can reference other tests with assertions of `answer.stringValue`.

packages/web-forms/src/components/controls/Input/InputControl.vue
```diff
@@ -35,13 +36,19 @@ provide('isInvalid', isInvalid);
...
 		</template>
+		<template v-else-if="node.valueType === 'geopoint'">
+			<InputGeopoint :question="node" />
+		</template>
 		<template v-else>
 			<InputText :node="node" />
 		</template>
 
-		<i v-show="isInvalid && (doneAnswering || submitPressed)" class="icon-error" />
+		<i v-show="isInvalid && (doneAnswering || submitPressed) && node.valueType !== 'geopoint'" class="icon-error" />
```
eyelidlessness, line 46: Maybe add a comment for this exception? It took me a while to understand the intent of the change.

Longer term, I'm not really sure we want this structure: it seems like maybe each control (and maybe even specific appearances of some controls?) should determine whether to include this. I also wouldn't mind if a comment included a TODO note noting this!

packages/xforms-engine/src/client/InputNode.ts
```diff
@@ -56,13 +84,15 @@ export interface InputNode<V extends ValueType = ValueType>
 export type StringInputNode = InputNode<'string'>;
 export type IntInputNode = InputNode<'int'>;
 export type DecimalInputNode = InputNode<'decimal'>;
+export type GeopointInputNode = InputNode<'geopoint'>;
```
eyelidlessness, line 87: Something worth considering: these specific parameterizations of `InputNode` are given explicit names and exported because (at least I believe) they add clarity and convenience for client usage. We _could_ have clients reference the same thing as `InputNode<'geopoint'>` etc., but I think it's useful to be able to reference these _by name_.

Seeing parameterized `InputValue` types in clients has me wondering if we should do the same for those value representations. E.g.

- `StringInputValue`
- `IntInputValue`
- `DecimalInputValue`
- `GeopointInputValue`

Or perhaps switching the order of the naming convention, e.g. `InputStringValue`, ..., `InputGeopointValue`? There's something nice about that, and it might also help the next point feel less intractable...

It _might_ also be useful to do the same for their corresponding input representations (`InputNodeInputValue<V>` -> ?). But that would open an awkward naming can of worms. I don't see any reference to `InputNodeInputValue` outside of the engine, so the utility in this case is more theoretical and I'd happily defer this aspect.

packages/web-forms/src/components/controls/Input/InputGeopointReadonly.vue
```diff
@@ -0,0 +1,48 @@
+<script setup lang="ts">
```
eyelidlessness, line 1: I'm not sure about the name of this component. As I understand it, the intent is to have a common presentation of _geopoint value as text_. This lines up with what we discussed, and makes sense! I think that intent might be a better basis for naming the component itself.

**Future consideration:** while it'd probably be overkill (now), it's also worth thinking about whether _every_ `NoteNode<V>` / `InputNode<V>` pair might want a similar shared _value-as-text_ component? That would be good for consistency, and it would also be good for when we might want to display the value of an `InputNode` rather than the control (e.g. to collapse or summarize some/all of the form state).

packages/web-forms/src/components/controls/Input/InputGeopoint.vue
```diff
@@ -0,0 +1,499 @@
...
+<template>
+	<div ref="controlElement" class="geopoint-control">
+		<Button
+			v-if="value == null"
+			rounded
+			class="get-location-button"
+			:disabled="disabled"
+			@click="start()"
+		>
+			<svg
```
eyelidlessness, line 187: Something that might make these feel less awkward in this inbetween time until we have a better icon solution: could we give each of these names, create components for them, and import/reference them by name?

packages/web-forms/src/components/controls/Input/InputGeopoint.vue
```diff
@@ -0,0 +1,499 @@
...
+let intervalID: NodeJS.Timeout | undefined;
+
+const value = computed((): InputValue<'geopoint'> => {
+	return props.question.currentState.value;
+});
+
+/**
+ * Default accuracy in meters that can usually be reached by modern devices given enough time.
+ */
+const ACCURACY_THRESHOLD_DEFAULT = 5;
```
eyelidlessness, line 33: Not super concerned about this _yet_, but I think we should consider that some (maybe all) of these constants and accuracy-threshold logic are going to be shared across multiple presentations/components. I'd also like to think about a more systematic and consistent solution to constants overall. In some cases it really does make the most sense for them to be module-local (e.g. giving a name to, and documenting, a value that we'd otherwise call a "magic value"). But every time I define a constant that's basically representing business logic, design decisions, shared configuration, etc... I get kinda nervous that each new one-off is a little bit different, harder to find and update, harder to document, harder to answer questions like "Is this the same as Collect? Should it be? Is it different for a specific reason?" etc.

packages/web-forms/src/components/controls/Input/InputGeopoint.vue
```diff
@@ -0,0 +1,499 @@
...
+const accuracyThreshold = computed<number>(() => {
+	return props.question.nodeOptions.accuracyThreshold ?? ACCURACY_THRESHOLD_DEFAULT;
+});
+
+/**
+ * Default unacceptable accuracy in meters, which is about the length of a city block.
+ */
```
eyelidlessness, line 40: This is the kind of thing where I _really_ wish the language and type system had a way of expressing units. 🙃

packages/web-forms/src/components/controls/Input/InputGeopoint.vue
```diff
@@ -0,0 +1,499 @@
...
+const unacceptableAccuracyThreshold = computed<number>(() => {
+	const threshold = props.question.nodeOptions.unacceptableAccuracyThreshold;
+	return threshold ?? UNACCEPTABLE_ACCURACY_THRESHOLD_DEFAULT;
+});
+
+/**
+ * Defines geolocation accuracy levels:
+ *  - 'good' if accuracy is <= unacceptableAccuracyThreshold
+ *  - 'poor' if accuracy > unacceptableAccuracyThreshold
+ *  - 'unknown' if accuracy is null or undefined
```
eyelidlessness, line 51: I feel a little icky that we use CAPS/lower inconsistently for enumerated values throughout the project, but I think we can address that when we settle #296. In the meantime, I think it's best if we at least self-reference them consistently 🙃.

```suggestion
 *  - 'GOOD' if accuracy is <= unacceptableAccuracyThreshold
 *  - 'POOR' if accuracy > unacceptableAccuracyThreshold
 *  - 'UNKNOWN' if accuracy is null or undefined
```

packages/web-forms/src/components/controls/Input/InputGeopoint.vue
```diff
@@ -0,0 +1,499 @@
...
+
+	if (accuracy > unacceptableAccuracyThreshold.value) {
+		return ACCURACY_QUALITY.POOR;
+	}
+
+	return ACCURACY_QUALITY.GOOD;
+};
+
+const getQualityLabel = (quality: AccuracyQualityValue): string => {
+	// TODO: translations
```
eyelidlessness, line 76: Since this applies to each of the return values:

```suggestion
// TODO: translations
const getQualityLabel = (quality: AccuracyQualityValue): string => {
```

eyelidlessness (commented):
As mentioned in Slack, I haven't finished a full pass. But I wanted to get you the actionable feedback I have now before I'm off for a break!

packages/web-forms/src/components/controls/Input/InputGeopoint.vue
latin-panda: Fixed!

packages/common/src/fixtures/geopoint/1-geopoint.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0"?>
```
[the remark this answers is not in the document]
latin-panda: Done, I've removed `<?xml version="1.0"?>`

packages/common/src/fixtures/geopoint/1-geopoint.xml
```diff
@@ -0,0 +1,38 @@
...
+            <instanceID/>
+          </meta>
+        </data>
+      </instance>
+      <bind nodeset="/data/facility_gps" type="geopoint"/>
+      <bind nodeset="/data/meta/instanceID" type="string" readonly="true()" jr:preload="uid"/>
+    </model>
+  </h:head>
+  <h:body>
+    <input accuracyThreshold="5" unacceptableAccuracyThreshold="50" ref="/data/facility_gps">
```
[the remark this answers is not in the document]
latin-panda: Done,  I added these inputs and 1 is a required field, all in this form

packages/web-forms/src/components/controls/Input/InputControl.vue
latin-panda: Done. I added a comment

packages/web-forms/src/components/controls/Input/InputGeopoint.vue
latin-panda: Fixed

packages/web-forms/src/components/controls/Input/InputGeopointReadonly.vue
latin-panda: I'm renaming it to `GeopointFormattedValue.vue` and moving it out of /input folder. 

`packages/web-forms/src/components/controls/GeopointFormattedValue.vue`

packages/web-forms/src/components/controls/Input/InputGeopoint.vue
latin-panda: Yes, I can relate to this feeling. For now, I prefer to keep it close to where it's used and aligned with Collect specs. This is mainly because I’m unsure if other question types or clients might have different default values. It's something we can revisit once Web Forms has an attribute where the default is reused in multiple places or even just to serve as documentation if it's different from Collect at some point

packages/scenario/test/bind-types.test.ts
```diff
@@ -182,6 +184,33 @@ describe('Data (<bind type>) type support', () => {
...
+		describe('type="geopoint"', () => {
+			let answer: ModelValueNodeAnswer<'geopoint'>;
+
+			beforeEach(() => {
+				answer = getTypedModelValueNodeAnswer('/root/geopoint-value', 'geopoint');
+			});
+
+			it('has a GeopointValue | null static type', () => {
+				expectTypeOf(answer.value).toEqualTypeOf<InputValue<'geopoint'>>();
+			});
```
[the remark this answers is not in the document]
latin-panda: I like this one! I've fixed both tests

packages/xforms-engine/src/client/InputNode.ts
latin-panda: It makes sense! I've added those four input types for completeness, even though they are not used yet. 

NoteNode and InputNode respectively 
```
export type StringNoteValue = NoteValue<'string'>;
export type IntNoteValue = NoteValue<'int'>;
export type DecimalNoteValue = NoteValue<'decimal'>;
export type GeopointNoteValue = NoteValue<'geopoint'>;
```

packages/common/src/fixtures/geopoint/1-geopoint.xml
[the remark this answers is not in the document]
eyelidlessness: Oof my bad, I should've been more clear! That is the XML prolog, I don't think it should be removed. I was talking about the numeric prefixes on fixture _file names_. And I wouldn't worry about changing that now either, just wanted to let you know that's not an intentional  convention we expect!

packages/xforms-engine/src/parse/body/control/InputControlDefinition.ts
eyelidlessness: Sorry, I meant to come back to this before. I meant in the project in general. My intent was not to change anything in how the attributes are parsed, just to add context to the TODO comments as they were already written.

packages/web-forms/src/components/controls/Input/InputGeopoint.vue
```diff
@@ -0,0 +1,499 @@
+<script setup lang="ts">
```
eyelidlessness, line 1: Capturing some discussion from Slack: I would like to explore the possibility of reworking this component's state into an explicit state machine. As I started looking at this I was concerned about trying to follow all of the various states and how they interact, and that concern grew as I tried to build a mental model. My main concerns are around maintainability: it might be easier to track all of the implications of each state change now while this is fresh (although I find that challenging already!), but I am more doubtful about that as it fades from recent memory.

As part of trying to build a mental model, I starting taking text notes roughly structured as how I'd approach the various related/interdependent states as a state machine. Below is a cleaned up version of that with a bit of added context.

## Given: `Geopoint` and `GeopointValue` types

```ts
interface Geopoint {
  readonly latitude: number;
  readonly longitude: number;
  readonly altitude: number | null;
  readonly accuracy: number | null;
}

type GeopointValue = Geopoint | null;
```

Naming is not the important detail here. But what is important: defining this type explicitly, in the client, allows treating these as mutually assignable:

* `GeopointInputNode.currentState.value`
* Client-local (in memory) state of the same value, regardless of its source

## Given: `GeopointInputNode.currentState.value: GeopointValue` and `GeopointInputNode.setValue(value: GeopointValue)`

This is just explicitly acknowledging that:

- The node's current state is assignable to the local state (and vice versa)
- The node's value setter is ultimately used to commit local state, and can accept the in-memory representation

## Given: in-form and dialog value displays (may) have different sources

* In-form display value's source is always the node's committed value state (`GepointInputNode.currentState.value`)
* Dialog display value's source is whichever is latest:
  * In-memory _value state_ (latest if non-`null`)
  * Node's committed value state (latest if in-memory value state is `null`)

## `InputGeopoint` state machine

This is written as a tree, and references a variety of related sub-states. I would _strongly encourage_ considering how the various sub-states can be represented as a **complete state**, i.e. essentially a single `Ref<InputGeopointState>` (likely composed!).

* **INITIAL**:
  * Set in-memory _value state_: `null`
* **REQUESTED**:
  * Essential state: request geolocation
    * Watch (`navigator.geolocation.watchPosition(...)`, retain `watchId` reference for cleanup). _SUB-STATES_:
      * _PENDING_
      * _SUCCESS_: set in-memory _value state_: `GeolocationPosition`
        * if _value state_ `accuracy` satisfies `accuracyThreshold` -> **COMMITTED**
        * else continue in **REQUESTED** state, until either:
          * subsequent _SUCCESS_ -> **COMMITTED**
          * User action "Cancel" -> **DISMISSED**
          * User action "Save" -> **COMMITTED**
      * _FAILURE_: set in-memory _error state_: [`GeolocationPositionError`](https://developer.mozilla.org/en-US/docs/Web/API/GeolocationPositionError), and note distinct error conditions (consider splitting these as distinct _sub-states_?):
        * if `PERMISSION_DENIED` (`1`): User denied permissions (message, as currently implemented)
        * if `POSITION_UNAVAILABLE` (`2`): (other message TBD)
        * ~~if `TIMEOUT` (`3`)~~: (not applicable unless/until we configure one)
    * Derived and secondary states (can be handled separately and composed in! But note that enter/exit implies access to current/previous **STATE**)
      * Entering **REQUESTED**:
        * [Note: `watchId` is tracked as essential!]
        * Trigger: show dialog (can be computed from state being **REQUESTED**!)
        * Store `startTime` (`Date.now()`)
        * Compute `duration`: `Date.now()` - `startTime`
          * Compute `'mm:ss'`
        * Start interval timer (retain `timerId` for cleanup): on each tick trigger recompute of `duration` (can be any recompute-triggering write to any reactive dependency referenced there!)
      * Exiting **REQUESTED**:
        * [Cleanup essential state `watchId`]
        * Trigger: hide dialog (can be computed from no longer being **REQUESTED**!)
        * Cleanup `startTime`
        * Cleanup interval timer `timerId`
* **DISMISSED**: effectively a side-effect trigger for "Exiting **REQUESTED**"
  * Retain requested state/sub-state in memory
* **COMMITTED**: effectively a side-effect trigger for:
  1. "Exiting **REQUESTED**"
  2. Write in-memory _value state_ to instance (`GeopointInputNode.setValue`)
  3. Cleanup in-memory _value state_ any other in-memory state associated with **REQUESTED**/_[PENDING | SUCCESS | FAILURE]_
  4. Return to **INITIAL** state

packages/common/src/fixtures/geopoint/1-geopoint.xml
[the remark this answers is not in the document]
latin-panda: Ohh okay, I've put back the `<?xml version="1.0"?>`
And removed the number in the file name

packages/scenario/test/bind-types.test.ts
latin-panda: I've made the test to accept GeopointValue object

packages/web-forms/src/components/controls/Input/InputGeopoint.vue
eyelidlessness: I want to make sure to call out that the above has been edited to address "quality of accuracy" in a "Given" section. I forgot to include those in the first pass on the state machine description, and opted not to include them there after editing because I think:

* it's fully derived state (same logic from any `GeopointValue` regardless of its source), and probably wouldn't have to be incorporated directly into a state machine
* it would make the state machine description unnecessarily deeper and more complicated, introducing a cross-reference for each value source

packages/web-forms/src/components/controls/Input/InputGeopoint.vue
latin-panda: I’m tempted to leave it as is for now—it’s a nudge to swap it out for an icon component next week. Plus, I’m planning to propose a small folder structure change in the Web Forms client, so this keeps fewer components shuffling around.

packages/xforms-engine/src/instance/InputControl.ts
eyelidlessness: I think it'd be good to add in this PR regardless of size, so there's a clear context for the addition. It'll be a little more diff noise. But it'll be easy to spot, eyeball as correct, and move on to the more salient changes. I'm happy to leave this unresolved while we're focused on those more salient changes and address it as a wrapup if you'd prefer!

packages/xforms-engine/src/lib/codecs/Geopoint/Geopoint.ts
```diff
@@ -0,0 +1,117 @@
...
+	getTuple(): GeopointTuple {
+		const { latitude, longitude, altitude, accuracy } = this.internalValue;
+
+		if (altitude.value == null && accuracy.value != null) {
+			return [latitude, longitude, altitude, accuracy];
+		}
+
+		if (altitude.value != null && accuracy.value != null) {
+			return [latitude, longitude, altitude, accuracy];
+		}
```
eyelidlessness, line 84: This is logically equivalent to:

```suggestion
		if (accuracy.value != null) {
			return [latitude, longitude, altitude, accuracy];
		}
```

packages/xforms-engine/src/lib/codecs/Geopoint/Geopoint.ts
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	getRuntimeValue(): GeopointRuntimeValue {
+		const { latitude, longitude, altitude, accuracy } = this.internalValue;
+		const isLatitude = this.isValidDegrees('latitude', latitude.value);
+		const isLongitude = this.isValidDegrees('longitude', longitude.value);
+
+		if (!isLatitude || !isLongitude) {
+			return null;
+		}
```
eyelidlessness, line 100: Just an observation, no action requested!

I hadn't thought about this case when I was pseudocoding out the tuple and semantic value concepts. Normally, I'd want to handle the invalid cases during (error) or prior to (return `null` from a factory) construction.

But I think this is the right call for now, **please don't change anything about this yet!** We'll need to revisit this as we incorporate the other codec use cases, and this leaves a good starting point to think about how to encode error conditions _into the value_ (presumably within its `internalValue`), which will be important as `Geopoint` is composed into `Geotrace` and `Geoshape` (and potentially `Line` distinct from `Geotrace`, for the XPath cases).

packages/xforms-engine/src/lib/codecs/Geopoint/GeopointValueCodec.ts
```diff
@@ -0,0 +1,46 @@
...
+				.getTuple()
+				.map((item) => item.value ?? 0)
+				.join(' ');
+		};
+
+		const decodeValue: CodecDecoder<GeopointRuntimeValue> = (value: string) => {
+			return decodeStringValue(value);
+		};
+
+		super('geopoint', encodeValue, decodeValue);
```
eyelidlessness, line 44: I'm wondering if it might make sense to move more (maybe even all) of the encode/decode logic to `Geopoint`, e.g. adding:

- `static from(value: Geopoint | string | null): Geopoint | null`
- `serializeXMLValue(): string`[^1]

... with `encodeValue` and `decodeValue` just wrapping those here?

I'm expecting that will be valuable when we go to implement the other geo types, as well as integrate the other geo value use cases.

[^1]: I'm open to bikeshedding the name. I originally typed out `toString` but realized that's error prone and could obscure intent.

packages/xforms-engine/src/parse/body/NodeOptionsParser.ts
```diff
@@ -0,0 +1,25 @@
+export const parseToInteger = (value: string | null): number | null => {
```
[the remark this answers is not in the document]
eyelidlessness: Eventually we'll want these in `@getodk/common` or some other place that we can share logic across packages. But let's leave this here for now and revisit when we can gather all of the cases where we need to get number values out of strings, so we can be sure we know all the nuances. Top of mind:

- failure modes
- non-failure parsing flexibility (which may be different depending on usage)
- semantic differences between decimals and floats
- usage (or not) in computations and e.g. whether `number`/float precision errors are acceptable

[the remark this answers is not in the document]
eyelidlessness: Oof. I didn't actually pay much attention to the _module name_ here! I only spotted it when I was surprised by the corresponding [test suite description](https://github.com/getodk/web-forms/pull/300/files#diff-b31eb7edf3bbf52cf8bf0281c26416ec9ff105467d57dd2f45dd02eaf26527d0R4). I think we should reconsider at least the module name (and test suite to go match that), and at that point might as well move to `lib` until we have a more suitable shared home for it.

eyelidlessness (commented):
I think I've been through a full pass _top to bottom_ through the PR, although I have seen some changes and comments coming in as I've progressed. So I know that there's already more for me to look at. But I'm not sure how much more steam I have for the day, so I wanted to make sure I post the feedback I do have before I run out!

packages/xforms-engine/src/parse/body/NodeOptionsParser.ts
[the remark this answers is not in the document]
latin-panda: fixed

packages/web-forms/src/components/controls/Input/InputGeopoint.vue
latin-panda: I've made the refactor to accommodate a state machine. 

> if PERMISSION_DENIED (1): User denied permissions (message, as currently implemented)
if POSITION_UNAVAILABLE (2): (other message TBD)

@alyblenkin  and I chatted about these error messages over Figma (I can't find it anymore; maybe it was deleted when we redesigned the UI). For both permission denied and position unavailable, we're displaying the same troubleshooting message for the user so they check permissions and the location is turned on. 

`Cannot access location Grant location permission in the browser settings and make sure location is turned on.`

packages/xforms-engine/src/lib/codecs/Geopoint/GeopointValueCodec.ts
latin-panda: Done

latin-panda:
@eyelidlessness, I've incorporated the state machine into the Vue component and refactored the Goepoint codec. Can you please have a look?

packages/xforms-engine/src/instance/InputControl.ts
latin-panda: @eyelidlessness Please find the implementation [in this commit ](https://github.com/getodk/web-forms/pull/300/commits/1eca15a697e9f4c1c7b26fc8d2440e8e25312d24). Does it look good?

eyelidlessness: Looks almost like what we discussed! I think we'll want to add `readonly nodeOptions: null` to the client interfaces:

- `GroupNode`
- `ModelValueNode`
- `NoteNode`
- `RangeNode`
- `RankNode`
- `RootNode`
- `SelectNode`
- `SubtreeNode`
- `TriggerNode`

It's likely we'll add non-`null` options for some of these relatively soon (like an interface for form title on `RootNode`). I think it's best for them to reflect their actual types so the client interface and implementations remain in sync as we make any node-specific additions.

latin-panda: Done! [In this commit](https://github.com/getodk/web-forms/pull/300/commits/37efc2f9842faa9377ffe033716f4715b4fc178b)

latin-panda:
@eyelidlessness ready for review

latin-panda:
The final manual test is complete. 
- [Test case scenarios ](https://github.com/getodk/web-forms/issues/249#issuecomment-2635178304)
- Test case evidence: All videos have been uploaded in this PR description

latin-panda:
@eyelidlessness I pushed a small format change that Yaw, Hélène, and Aly requested. 
- Adds a space before meter unit: from `123m` to `123 m` (I didn't know it was the correct way) 
- Truncates only the accuracy to max three decimals

packages/xforms-engine/src/lib/NumberParsers.ts
```diff
@@ -0,0 +1,25 @@
...
+		return null;
+	}
+
+	const parsed = Number(value);
+	if (typeof value !== 'string' || value.trim() === '' || isNaN(parsed)) {
+		throw new Error(`Expected a float, but got: ${value}`);
+	}
+
+	return parsed;
+};
```
eyelidlessness, line 25: I don't want to send this through another round trip, just for future reference: we usually use `PascalCase` modules names as 1:1 with a class and/or interface of the same name. If I were creating this module I'd probably `kebab-case` it.

packages/xforms-engine/src/lib/NumberParsers.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const parseToFloat = (value: string | null): number | null => {
+	if (value === null) {
+		return null;
+	}
+
+	const parsed = Number(value);
+	if (typeof value !== 'string' || value.trim() === '' || isNaN(parsed)) {
```
eyelidlessness, line 20: Up to you if you want to change this to
```suggestion
	if (typeof value !== 'string' || value.trim() === '' || !Number.isFinite(parsed)) {
```

Note: in general I think for _just_ `NaN` checking, `Number.isNaN` is preferred over the global `isNaN`. But if you don't want to do the suggested change I'd just leave as-is. This is really something we should have a lint rule for.

packages/web-forms/src/components/controls/Input/InputGeopoint.vue
eyelidlessness: I think it's fair to assume these followups are deferred for a later scope! And I've volunteered to come back to them. I'll probably want to tackle both followup concerns together, as it's likely the sharing solution will be useful for/applicable to other concerns relatively immediately (ahem #311).

I'm trying not to let my focus wander from this review, I'm nearing the end! Would you mind filing an issue to track following up on this (if someone hasn't already)?

packages/scenario/test/smoketests/child-vaccination.test.ts
```diff
@@ -760,7 +760,7 @@ describe('ChildVaccinationTest.java', () => {
 
 			scenario.next('/data/not_single');
 			scenario.next('/data/not_single/gps');
-			scenario.answer('1.234 5.678');
+			scenario.answer('1.234 5.678 0 2.3');
```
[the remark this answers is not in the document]
eyelidlessness: For context: it's been helpful, as much as possible, to try to have as few source-level differences as we can in `scenario` test bodies between JavaRosa and Web Forms. In some cases it's impossible or really impractical for them to be the same, but trivial differences are a good way for us to get confused when we try to reconcile changes over time. (And this is one of the first tests of that I think! So I think it's a good idea to reinforce that precedent here.)

eyelidlessness (approved):
Thank you! This is 🎉  **AWESOME!** 🎉 

I've added comments about a last few little nitpicky things. I think at least a few are worth addressing, but I'm excited to land this so I'm pre-approving pending those remaining nits.

Feel free to merge when you're satisfied with any changes related to those last few comments!

🚀

packages/web-forms/src/components/controls/Input/Geopoint/GeolocationRequest.ts
```diff
@@ -0,0 +1,42 @@
+import type { GeopointValueObject } from './GeopointValueObject.ts';
+
+export const GEOLOCATION_STATUS = {
+	PENDING: 'PENDING',
+	SUCCESS: 'SUCCESS',
+	FAILURE: 'FAILURE',
+} as const;
```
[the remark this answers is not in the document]
latin-panda: Yes. It'd be nice to find something that can do a good type check but is also accessible with dot notation to not hardcode string, and the IDE can do the autocomplete – small things for the developer's heart😁

packages/xforms-engine/src/lib/NumberParsers.ts
latin-panda: I also see a couple of filenames with `cammelCase`

PR #302, opened 2025-02-10
Add `cities.geojson`, external secondary instance for “all questions" demo form

eyelidlessness:
~~Closes #~~

- [x] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [x] Not applicable

Ran `yarn workspace @getodk/web-forms dev`, no longer see error loading the "all questions" demo form.

This file wasn't included when we introduced the demo form, but we didn't yet support external secondary instances so it didn't matter much.

I haven't added any of the _other media_ resources referenced by the form here. I'm hoping that, with whatever design we end up with for supporting those other form attachments, we'll similarly be alerted that they're absent. If nothing else, I think that will be an opportunity to make sure the form itself is up to date as well.

N/A

This is adding a resource which was missing for the "all questions" form.

## What's changed

See above.

### DIFF
--- a/packages/common/src/fixtures/preview-service/xforms/cities.geojson
+++ b/packages/common/src/fixtures/preview-service/xforms/cities.geojson
@@ -1,0 +1,65 @@
+{
+    "type": "FeatureCollection",
+    "features": [
+        {
+            "type": "Feature",
+            "properties": {
+		"title": "Warsaw",
+                "info": "Capital city of Poland",
+                "id": "1"
+            },
+            "geometry": {
+                "type": "Point",
+                "coordinates": [
+		    21.0122,
+                    52.2297
+                ]
+            }
+        },
+        {
+            "type": "Feature",
+            "properties": {
+		"title": "Berlin",
+                "info": "Capital city of Germany",
+                "id": "2"
+            },
+            "geometry": {
+                "type": "Point",
+                "coordinates": [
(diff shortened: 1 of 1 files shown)


packages/common/src/fixtures/preview-service/xforms/cities.geojson
```diff
@@ -0,0 +1,65 @@
+{
+    "type": "FeatureCollection",
+    "features": [
+        {
+            "type": "Feature",
+            "properties": {
+		"title": "Warsaw",
```
latin-panda, line 7: Observation - no blocking: 

For some reason, the spacing for indentation displays differently in GitHub from properties like this `title` vs `info` Line 8. Are these different characters? 

Another one, L14 vs L15

latin-panda (approved):
LGTM! 🚀

packages/common/src/fixtures/preview-service/xforms/cities.geojson
eyelidlessness: Looks like mixed tabs/spaces. FWIW I literally downloaded this file from Google Drive and committed as-is. As with any fixtures we've inherited from JR, I think it's a good idea to preserve the original formatting so we can clearly see substantive differences as it gets updated from upstream over time.

PR #313, opened 2025-02-25
chore(#330): Adds e2e tests for default geopoint question type

latin-panda:
Closes https://github.com/getodk/web-forms/issues/330
Part of #249

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [X] Not applicable

Running tests locally

Initially, I tried adding tests in Vue, but our Web Form client is primarily a presentation layer, and Vue doesn't support tests requiring browser permissions. Therefore, I chose Playwright for automation (already a project dependency), following a standard folder structure and the use of page objects, as I’ve done with other test frameworks (Protractor, WebDriverIO). This is compatible with most frameworks if we ever switch in the future. Additionally, I included a minimal README file to guide contributors in adding new tests.

N/A

N/A

## What's changed

This is not perfect, and more can be improved, but it is a good start that we can iterate as we add more tests. 

- Adds minimal README file to guide contributors when adding new tests
- Adds a new folder structure 
- Moves the content of `wf-preview.test.ts` to `/e2e/test-cases/build/web-forms-preview-forms.test.ts`. The original name wasn't descriptive. 
- Adds new basic tests for geopoint and note
- Adds new page objects for fill form page, preview page and geopoint control


packages/web-forms/e2e/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+This directory contains end-to-end (E2E) tests for the `@getodk/web-forms` package, which powers form filling and submission editing of ODK forms in a web browser. These tests use [Playwright](https://playwright.dev/) to simulate user interactions and ensure the package works as expected in real-world scenarios.
+
+## Folder Structure
+
+The E2E tests are located in `packages/web-forms/e2e/`. Here's the structure and purpose of each directory:
+
+```
+e2e/
+├── fixtures/         # Sample forms and data for testing.
```
latin-panda, line 11: At the moment, this folder is empty, but next time, it'd be good to take the XML forms from here. I added a ToDo comment for this improvement in the PreviewPage.ts

packages/web-forms/e2e/build/wf-preview.test.ts
latin-panda, line 1: This file name is confusing. I moved the test to `/e2e/specs/build/preview-form-render.test.ts`

We can move the `/e2e/specs/build/preview-form-render.test.ts` to `/e2e/specs/preview-form-render.test.ts`

packages/web-forms/e2e/specs/build/preview-form-render.test.ts
latin-panda, line 1: This file now contains the tests of `vue.test.ts` and `wf-preview.test.ts` since both verify the demo forms. 
I didn't change the tests; I just added the PreviewPage to navigate to the demo page. Everything else is as original

packages/web-forms/e2e/specs/build/style.test.ts
latin-panda, line 1: We could move this file out from the build folder, from `e2e/specs/build/style.test.ts` to `e2e/specs/style.test.ts`

packages/web-forms/e2e/vue.test.ts
latin-panda, line 1: Same, this file name is not descriptive, for now I moved the content to `e2e/specs/build/preview-form-render.test.ts`

latin-panda:
Hi @eyelidlessness, can you please review? Thank you for your time! 

Quick question, we don't need a changeset for this PR, or do we?

packages/web-forms/e2e/page-objects/pages/FormPage.ts
```diff
@@ -0,0 +1,65 @@
+import { expect, Page } from '@playwright/test';
+import { GeopointControl } from '../controls/GeopointControl.js';
+
+export class FormPage {
+	private readonly page: Page;
+
+	public readonly geopoint: GeopointControl;
```
eyelidlessness, line 7: I'm kind of confused about how this abstraction is named and how it interacts with either/both `PreviewPage` and `GeopointControl`. I thiiiiink the intent is something like:

- `PreviewPage` handles general navigational setup concerns
- `FormPage` handles semantic concerns **broadly related** to the UI _of a specific form_(?) and...
  - navigation to that form (by `PreviewPage`) is implied as a precondition(?)
  - `geopoint: GeopointControl` handles semantic concerns **narrowly related** to the UI of _geopoint-specific functionality_(?)

Maybe it would make more sense for `FormPage` to:

- Have a different name, or some other way of distinguishing it from the conceptually very different `PreviewPage`
- Compose (e.g. `extends`, but not necessarily) `PreviewPage`, maybe as a convenience
- Something else, if I've misunderstood important aspects of the intent

packages/web-forms/e2e/specs/geopoint.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+		test('captures good-accuracy location', async ({ context }) => {
+			await context.setGeolocation({
+				latitude: 40.7128,
+				longitude: -74.006,
+				accuracy: 10,
+				altitude: 0,
+			});
+
+			await formPage.expectLabel(`
```
eyelidlessness, line 27: Not a concern now/for these tests, but I have a lurking feeling that the "at index"/unindexed dichotomy is going to be a source of confusion, and a potential risk factor for e.g. flaky tests (things moving around, for reasons unrelated to the test) or false positives (things which are expected at a specific point in form filling progress, but happen to appear somewhere else for some unrelated reason).

Some conceptual prior art we can think about:

- In `@getodk/scenario`: `PositionalEvent`, the notion of a form as a linear sequence of positioned questions, and stateful progression through that (which also adapts to important form concepts like changing `relevant` state, adding/removing repeats, etc).

    **Notably**, this concept is reflected in `Scenario` methods like `next` (which advances positional state through a form, and its `prev` counterpart), `answer` (which sets state _at the current positional state_).
    
    **Also notably**, this whole set of concepts is _very well battle tested_, as it's all inherited from JavaRosa (albeit implemented rather differently!).

- We _may eventually_ (maybe even soon!) want to **generalize** something for this. Probably either in the engine, or at some layer between engine/client, so the same logic (especially stateful aspects of it) can be shared as appropriate by `scenario`, `web-forms` **UI** (e.g. paginated presentation of forms, progress indicator), E2E tests.

packages/web-forms/e2e/README.md
```diff
@@ -0,0 +1,58 @@
...
+├── page-objects/     # Page Object Model structure for e2e tests, organizing UI abstractions into pages and reusable controls.
+    ├── controls/     # Reusable controls such as form fields, UI components, etc.
+        ├── GeopointComponent.ts  # Example of a reusable control for the geopoint question type.
```
eyelidlessness, line 14: At the E2E level, I think it might be best to avoid "component" terminology here entirely. Since it's become a technical term of art, I worry that something like `GeopointComponent` might drift from association with a specific UI component.

packages/web-forms/e2e/README.md
```diff
@@ -0,0 +1,58 @@
...
+    ├── pages/        # Full page representations.
+        ├── FormPage.ts           # Example of a full page representation for a form.
+├── specs/            # Test specification files.
+        ├── geopoint.test.ts      # Example of a test file for the geopoint question type.
+```
+
+### Key Components
```
eyelidlessness, line 21: Same commentary on "component" but different reasoning. Maybe...

```suggestion
### Key concepts
```

packages/web-forms/e2e/README.md
```diff
@@ -0,0 +1,58 @@
...
+   Or run specific tests:
+
+   ```bash
+   yarn workspace @getodk/web-forms test:e2e <filepath, e.g. e2e/specs/geopoint.test.ts>
+   ```
+
+3. **Add new tests**
+   - Create fixtures in `fixtures/` if needed.
+   - Update or add Page Objects in `page-objects/` for new UI elements.
+   - Write tests in `specs/` with descriptive names.
```
eyelidlessness, line 52: Maybe remove this section for now?

1. Building dependent packages is a prerequisite for everything, and should be covered by package and/or project READMEs. Building the `@getodk/web-forms` package is actually handled by Playwright config.
2. Would probably fit better at the package-level README, in a section covering all package-specific testing.
3. Is pretty much redundant to information in other sections.

packages/web-forms/e2e/README.md
```diff
@@ -0,0 +1,58 @@
...
+    ├── pages/        # Full page representations.
+        ├── FormPage.ts           # Example of a full page representation for a form.
+├── specs/            # Test specification files.
+        ├── geopoint.test.ts      # Example of a test file for the geopoint question type.
+```
+
+### Key Components
+
+- **Fixtures**: Reusable test data (e.g., sample XForms) to simulate real-world use cases. Add new forms here as needed.
+- **Page Objects**: Implements the [Page Object Model (POM)](https://playwright.dev/docs/pom) pattern to encapsulate UI interactions, enhancing test readability and maintainability.
```
eyelidlessness, line 24: (Not our fault/problem, but I can't help call this out...)

What an unfortunate [name collision](https://maven.apache.org/guides/introduction/introduction-to-the-pom.html)! Microsoft should have known better.

eyelidlessness (changes requested):
Lots of feedback, I know! But I don't want that to detract from saying...

**THIS IS AWESOME!** Thank you so much for taking this on.

packages/web-forms/e2e/README.md
latin-panda: Thanks for the context!

I'm still unsure if the fixtures in `@getodk/common` will fully meet the E2E testing needs, but if a special case arises, e.g., verify UX based on special appearance that modifies the display, I don’t think there will be many, and we can add it to`@getodk/common`. 

This was intended to keep it contained from other types of testing, but I don't have a strong preference. This approach has existed for a while without producing any issues with other test types. I'm open to keeping it that way. 

Do you agree to remove the fixture from these e2e tests and keep using the one in `@getodk/common`?

packages/web-forms/e2e/specs/build/preview-form-render.test.ts
```diff
@@ -0,0 +1,130 @@
...
+				await langChanger.click();
+				await page.locator('.language-dd-label').last().click();
+			}
+
+			await page.goBack();
+			await expandAllCategories(page);
+		}
+
+		expect(consoleErrors).toBe(0);
+	});
```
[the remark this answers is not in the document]
latin-panda: @eyelidlessness, I would prefer to delete this test, create a new ticket with this detail, and link this PR. So it can be assigned as a proper task. 

Would that be okay?

packages/web-forms/e2e/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+```
+e2e/
+├── fixtures/         # Sample forms and data for testing.
+├── page-objects/     # Page Object Model structure for e2e tests, organizing UI abstractions into pages and reusable controls.
+    ├── controls/     # Reusable controls such as form fields, UI components, etc.
+        ├── GeopointComponent.ts  # Example of a reusable control for the geopoint question type.
+    ├── pages/        # Full page representations.
+        ├── FormPage.ts           # Example of a full page representation for a form.
+├── specs/            # Test specification files.
```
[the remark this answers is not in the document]
latin-panda: I was actually expecting to find the e2e tests inside the `/test` folder, any type of test actually, for example:

|─ tests/
|─|─ e2e/
|─|─ |─ page-objects/
|─|─ |─ specs/
|─|─ unit/
|─|─|─ components/
|─|─ integration etc....

Would it be possible for us to do this? Starting with Web Forms client. What do you think? 

<br>

> But it's potentially a major source of confusion where we as a team already refer to several things as "specs"

Good point! What do you think about these options:
- `cases/` or `test-cases/` -- simple and clear
- `assertions/` -- not super fan, it feels like utilities asserting values, but still clear
- `journeys/` -- like user journeys as e2e tests is simulating user interactions

packages/web-forms/e2e/specs/build/preview-form-render.test.ts
```diff
@@ -0,0 +1,130 @@
...
+		const collapsedDetails = document.querySelectorAll<HTMLDetailsElement>('details:not([open])');
+
+		collapsedDetails.forEach((details) => {
+			details.open = true;
+		});
+	});
+};
+
+test.describe('Preview Form Render', () => {
+	test('demo forms load', async ({ context, page }) => {
```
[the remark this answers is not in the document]
latin-panda: Sure! 
- The file will remain in the build directory of e2e tests: `packages/web-forms/e2e/specs/build/`
- Changing the test description to `test.describe('Web Forms Preview Demo Forms', () => {`
- Changing the test filename to `web-forms-preview-forms.ts` or `preview-demo-forms.ts`

Did I capture correctly?

packages/web-forms/e2e/README.md
```diff
@@ -0,0 +1,58 @@
...
+   ```
+
+3. **Add new tests**
+   - Create fixtures in `fixtures/` if needed.
+   - Update or add Page Objects in `page-objects/` for new UI elements.
+   - Write tests in `specs/` with descriptive names.
+
+## Contributing
+
+- Keep tests focused: one feature per file.
```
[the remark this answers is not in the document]
latin-panda: Yes, that's good

packages/web-forms/e2e/page-objects/pages/PreviewPage.ts
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	async goToPage() {
+		await this.page.goto(BASE_URL);
+	}
+
+	/**
+	 * Opens a preexisting demo form by navigating through the demo forms UI.
+	 *
+	 * @TODO Add support for uploading XML forms from /e2e/fixtures directory for test-specific scenarios
```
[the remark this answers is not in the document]
latin-panda: Agree, I've created a ticket: https://github.com/getodk/web-forms/issues/325

packages/web-forms/e2e/page-objects/controls/GeopointControl.ts
```diff
@@ -0,0 +1,68 @@
+import { expect, Page } from '@playwright/test';
+
+export class GeopointControl {
```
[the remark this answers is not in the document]
latin-panda: Thank you!

packages/web-forms/e2e/page-objects/pages/PreviewPage.ts
```diff
@@ -0,0 +1,47 @@
+import { expect, Page } from '@playwright/test';
+
+const BASE_URL = 'http://localhost:5173';
```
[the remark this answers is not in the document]
latin-panda: I didn't notice that one-number difference. Having something like these classes will add a lot of clarity to the tests. Great catch!

packages/web-forms/e2e/specs/build/style.test.ts
```diff
@@ -1,7 +1,9 @@
 import { expect, test } from '@playwright/test';
+import { PreviewPage } from '../../page-objects/pages/PreviewPage.js';
 
 test('Build includes component-defined styles', async ({ page }) => {
-	await page.goto('http://localhost:5174/');
+	const previewPage = new PreviewPage(page);
+	await previewPage.goToPage();
```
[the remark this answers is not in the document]
latin-panda: Fixed

packages/web-forms/e2e/specs/build/preview-form-render.test.ts
```diff
@@ -0,0 +1,130 @@
...
+		const previewPage = new PreviewPage(page);
+		await previewPage.goToPage();
```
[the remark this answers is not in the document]
latin-panda: Fixed

packages/web-forms/e2e/specs/geopoint.test.ts
```diff
@@ -0,0 +1,122 @@
...
+			const previewPage = new PreviewPage(page);
+			await previewPage.goToPage();
+			await previewPage.openDemoForm('geopoint', 'geopoint.xml', 'Geopoint');
+		});
+
+		test('displays troubleshooting message when permission is denied', async () => {
+			await formPage.expectLabel('Where are you filling out the survey?');
+			await formPage.geopoint.openDialog();
+			await formPage.geopoint.expectGeopointErrorMessage(
+				'Cannot access location Grant location permission in the browser settings and make sure location is turned on.'
```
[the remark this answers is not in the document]
latin-panda: > E2E testing may be a good place to actually test app translations? I'm not sure, that may get excessive pretty quickly!

Yes, the e2e tests are usually the place to test UI translations because they are a visual feature for the user directly. We can even start testing that now, we have some demo forms with french translations. I've created a ticket for that https://github.com/getodk/web-forms/issues/326

 > This will eventually be translated, and testing for English prose is more brittle than we'll want; could be better to check for something in the DOM associated with the targeted prose

I'm unsure if I get this comment 100%, but I will try answering. The intention is to get as close to the actual user experience as possible. E2e tests don't catch some errors (I've seen it happen) when checking by DOM element's IDs or classes; the element turned out hidden, or it's there, but the content is empty. 
It's okay to use a little help from the CSS class selector but then check by content (text) and visibility. 

This image isn't using Playwright selectors but e2e in general:

Screenshot 2025-03-04 at 5 57 50 PM

packages/web-forms/e2e/specs/note.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+	test.beforeEach(async ({ page }) => {
+		formPage = new FormPage(page);
+
+		const previewPage = new PreviewPage(page);
+		await previewPage.goToPage();
+		await previewPage.openDemoForm('notes', '2-all-possible-notes.xml', 'Notes');
+	});
+
+	test('renders all possible note types', async ({ context }) => {
```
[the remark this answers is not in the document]
latin-panda: I also wanted to test the display of these notes: 
Screenshot 2025-03-04 at 6 08 38 PM

What about if we leave them here?

packages/web-forms/e2e/README.md
latin-panda: Fixed

packages/web-forms/e2e/README.md
latin-panda: Agree, removed

packages/web-forms/e2e/README.md
latin-panda: I thought the same thing!

packages/web-forms/e2e/README.md
latin-panda: Sure! I changed it to control for these things: GeopointControl

packages/web-forms/e2e/page-objects/pages/FormPage.ts
latin-panda: This is the idea from user exp point of view:

There are two pages the user is interacting with: Web Forms' "Preview" page and "Forms" page

- The `PreviewPage` class controls the interactions on this page: 
  - Such as scrolling to the "Demo forms for dev" section, opening an accordion and clicking on a form

Screenshot 2025-03-04 at 7 02 57 PM

<br>

- The `FormPage` class controls the interactions on this page:
    - I named it FormPage, because is a page that loads forms. But I'm open to a better name suggestion
Screenshot 2025-03-04 at 7 01 14 PM

latin-panda: From the "Preview" page, the user can navigate to a "Form" page, but both are independent pages.  

A "Form" page can have geopoint question types, but the user does many steps to use this question type (get location, save location, retry), so the "GeopointControl" is created to abstract some of that. The geopoint is a kinda complex component.  A note or input text might be more straightforward (the user does one step to interact with them).

packages/web-forms/e2e/specs/geopoint.test.ts
```diff
@@ -0,0 +1,122 @@
...
+				['Accuracy: 7m', 'Latitude: 80.5128', 'Longitude: -99.9099'],
+				'Good accuracy'
+			);
+		});
+	});
+
+	test.describe('Geolocation permission denied', () => {
+		let formPage: FormPage;
+
+		test.beforeEach(async ({ browser }) => {
```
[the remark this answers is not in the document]
latin-panda: > Does this need to be a beforeEach for one test?

Not necessarily 

> Are you anticipating more tests in this sub-suite?

It's possible; I was hoping to get feedback from QA about other test cases. IMO this test is enough for now

packages/web-forms/e2e/specs/geopoint.test.ts
latin-panda: I didn't want to add these `at index` functions, but I didn't get a better idea fast enough to test these notes automatically (It was also a bit of an experiment to see how to address these cases)

Screenshot 2025-03-04 at 8 32 04 PM

packages/web-forms/e2e/page-objects/pages/PreviewPage.ts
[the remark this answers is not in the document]
latin-panda: Instead of classes, I went with functions for now, both loads the Preview page

```
async goToDevPage() {
  await this.page.goto(DEV_BASE_URL);
}

async goToBuildPage() {
  await this.page.goto(BUILD_BASE_URL);
}
```

packages/web-forms/e2e/README.md
lognaturel: Makes sense to me to remove the `fixture` directory from here and document that form fixtures should be put in `@getodk/common` so they can be used in different contexts and so that there's a single "form library".

packages/web-forms/e2e/README.md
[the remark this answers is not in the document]
lognaturel: How about renaming `specs` to either `test-cases` or `tests` for now and if someone feels strongly about changing the directory structure they can file an issue to pitch the new structure?

packages/web-forms/e2e/page-objects/pages/FormPage.ts
lognaturel: How about something like `FillFormPage` for now? For what it's worth, Collect has [`FillBlankFormPage`](https://github.com/getodk/collect/blob/6a2a4da662db8431c5d92e1a5b6d423f132bc817/collect_app/src/androidTest/java/org/odk/collect/android/support/pages/FillBlankFormPage.java#L24) for the same concept. I agree that in general "form" is overloaded and confusing.

packages/web-forms/e2e/specs/note.test.ts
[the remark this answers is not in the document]
lognaturel: It makes sense to me to have a specific form verifying notes with different value types. It's something that has regressed before in both Enketo and Collect so having an e2e form test here makes sense. I would also expect to see assertions about not having any user-editable control but that could be layered on.

packages/web-forms/e2e/specs/geopoint.test.ts
lognaturel: > so the same logic (especially stateful aspects of it) can be shared as appropriate by scenario, web-forms UI (e.g. paginated presentation of forms, progress indicator), E2E tests

YES, this sounds great!

packages/web-forms/e2e/specs/geopoint.test.ts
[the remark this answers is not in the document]
lognaturel: > We can even start testing that now, we have some demo forms with french translations.

Good idea. I think here T was referring to application translations which we don't have yet. Once we do have them we'll need to decide whether we test against literal strings like you have done here or references to strings. That can come later!

packages/web-forms/e2e/README.md
latin-panda: Thank you, this is done.

packages/web-forms/e2e/specs/build/preview-form-render.test.ts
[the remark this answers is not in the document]
latin-panda: These 3 bullets ⬆️ have been applied.

packages/web-forms/e2e/README.md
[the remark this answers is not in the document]
latin-panda: OK. It's renamed to `test-cases`

packages/web-forms/e2e/specs/note.test.ts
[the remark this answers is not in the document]
latin-panda: Leaving this test file as it is for now and asserting some of those Notes.

packages/web-forms/e2e/specs/geopoint.test.ts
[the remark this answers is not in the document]
latin-panda: In this case, we write test cases to evaluate translations. For example, one suite verifies that everything functions correctly in English, while additional tests specifically check form functionality when Swahili is selected. It can still find by Swahili text on the expected location on the screen.

packages/web-forms/e2e/page-objects/pages/FormPage.ts
latin-panda: I've applied the suggestion to use `FillFormPage`

It breaks a bit the convention of using nouns for page objects, but I couldn't come up with a better name (`FormFillerPage` sounds weird, `FormInteractionPage`?). This new name is still descriptive.

latin-panda:
@lognaturel This is ready for review, I have resolved the new feedback. 

Once approved, I will "dismiss" the pending change request to unblock the `merge` button. So, these tests are available for the PrimeVue upgrade work.

Please let me know if more changes are needed after the merge, and I will open a new PR addressing those requests.

lognaturel (approved):

Issue #319, opened 2025-02-27
Run E2E tests in CI


eyelidlessness:
We don't do this. I could've sworn I filed an issue already, but I can't find it! So I am filing it now. Sorry if it's redundant!

PR #320, opened 2025-02-27
engine (fix): Forms load with any `jr:preload` or `jr:preloadParams` value

eyelidlessness:
Addresses the current failure to load the IFRC Distribution Site Assessment fixture in Web Forms Preview.

Closes # (N/A)

- [x] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [x] Not applicable
- [x] Local build of Web Forms Preview

I just kept relaxing types and removing enforcement until TypeScript stopped saying there were errors. Then I ran build/tests. Finally, I manually tested in WFP to be sure the affected form loads without error.

It's worth noting that the E2E test added in b0f19c3e5fac71444b635afb6ac8b8c7311989fd (#265) would have caught this, but we still don't run E2E tests in CI (now tracked in #319).

The other approach (the change prior to this PR) was parsing a more clear and consistent data model, according to spec.

Near term: no risk.

Longer term: future spec changes affecting preloads may be more prone to dev mistakes.

https://github.com/getodk/web-forms/blob/d7ecc3335f6ccc9804a84fb4ff8447985ecaedfd/packages/common/src/fixtures/preview-service/xforms/ifrc-distribution-site-assessment-v1.xml

## What's changed

- Enforcement of the specified values of `jr:preload` or `jr:preloadParams` attributes on `<bind>` is removed
- Types corresponding to those values are relaxed to reflect that they are only hints of known values, similarly to how we represent appearances


latin-panda (approved):
LGTM!

PR #323, opened 2025-03-04
fix(#324): runs xpath tests with the same timezone

latin-panda:
Closes https://github.com/getodk/web-forms/issues/324

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [x] Not applicable

## What else has been done to verify this works as intended?

**To reproduce the error:** 

1. Change the machine time to March 1st at 23:30 Los Angeles timezone
2. Run the test suite `yarn workspace @getodk/xpath test`
3. See the tests failing like this:

Screenshot 2025-03-03 at 8 37 59 PM

**Other things done:**

After implementing the fix, I tried setting my machine in different timezones, ran the tests, and they passed. The different timezones have been passed to the ["now" function correctly](https://github.com/getodk/web-forms/blob/99295eb1d6ec78cd6a758385793e97859b6a74cc/packages/xpath/src/lib/datetime/functions.ts#L24). 

The offset is being appended to the [localDateTimeString](https://github.com/getodk/web-forms/blob/99295eb1d6ec78cd6a758385793e97859b6a74cc/packages/xpath/src/lib/datetime/functions.ts#L14)

I tried making a test about DST, but it's hard to trick reliably the test to take a different DST of the same timezone. However, I added some debugging code in [localDateTimeString](https://github.com/getodk/web-forms/blob/99295eb1d6ec78cd6a758385793e97859b6a74cc/packages/xpath/src/lib/datetime/functions.ts#L14) function, and I can see the offset switching to/from DST for the same timezone.

```
// The process.env.TZ value is America/New_York
// Current date is March 3rd

console.log(dateTime.offset); // Prints "-05:00"
const dstDate = Temporal.ZonedDateTime.from("2025-07-01T23:30:00[America/New_York]");
console.log(dstDate.offset); // Prints "-04:00"
```

This ensures the tests are run in the same timezone by setting it in the `beforeEach` hook using the `useFakeTimers` utility. However, this doesn't mock the DST, which has been challenging to tick the system timer. 

N/A

I don't think so

## What's changed
- Adds a global beforeEach and AfterEach hook to set the tests' timezone
- Adds a new optional environment variable for the tests' locale

### DIFF
--- a/packages/xpath/test/helpers.ts
+++ b/packages/xpath/test/helpers.ts
@@ -21,8 +21,6 @@ declare global {
 	 */
 	// eslint-disable-next-line no-var
 	var TZ: string | undefined;
-	// eslint-disable-next-line no-var
-	var LOCALE_ID: string | undefined;
 	/**
 	 * The locale string defining the language and regional formatting for tests.
 	 * This follows the BCP 47 language tag format (e.g., 'en-US'). It ensures consistent formatting
@@ -30,6 +28,8 @@ declare global {
 	 *
 	 * @example 'en-US' // American English
 	 */
+	// eslint-disable-next-line no-var
+	var LOCALE_ID: string | undefined;
 	// eslint-disable-next-line no-var
 	var IMPLEMENTATION: string | undefined;
 }
--- a/packages/xpath/test/helpers.ts
+++ b/packages/xpath/test/helpers.ts
@@ -10,14 +10,33 @@ type AnyParentNode =
 	| XMLDocument;
 
 declare global {
+	/**
+	 * The timezone identifier used for all date and time operations in tests.
+	 * This string follows the IANA Time Zone Database format (e.g., 'America/Phoenix').
+	 * It determines the offset and DST behavior for `Date` objects and
+	 * related functions.
+	 *
+	 * @example 'America/Phoenix' // Fixed UTC-7, no DST
+	 * @example 'Europe/London' // UTC+0 (GMT) or UTC+1 (BST) with DST
+	 */
 	// eslint-disable-next-line no-var
 	var TZ: string | undefined;
+	// eslint-disable-next-line no-var
+	var LOCALE_ID: string | undefined;
+	/**
+	 * The locale string defining the language and regional formatting for tests.
+	 * This follows the BCP 47 language tag format (e.g., 'en-US'). It ensures consistent formatting
+	 * across tests.
+	 *
+	 * @example 'en-US' // American English
+	 */
 	// eslint-disable-next-line no-var
 	var IMPLEMENTATION: string | undefined;
 }
 
 globalThis.IMPLEMENTATION = typeof IMPLEMENTATION === 'string' ? IMPLEMENTATION : undefined;
 globalThis.TZ = typeof TZ === 'string' ? TZ : undefined;
+globalThis.LOCALE_ID = typeof LOCALE_ID === 'string' ? LOCALE_ID : undefined;
 
 const namespaces: Record<string, string> = {
 	xhtml: 'http://www.w3.org/1999/xhtml',
@@ -337,4 +356,8 @@ export const getNonNamespaceAttributes = (element: Element): readonly Attr[] => 
 	const attrs = Array.from(element.attributes);
 
 	return attrs.filter(({ name }) => name !== 'xmlns' && !name.startsWith('xmlns:'));
+};
+
+export const getDefaultDateTimeLocale = (): string => {
+	return new Date().toLocaleString(LOCALE_ID, { timeZone: TZ });
 };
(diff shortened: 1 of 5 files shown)


packages/xpath/test/helpers.ts
```diff
@@ -10,14 +10,33 @@ type AnyParentNode =
...
+	var LOCALE_ID: string | undefined;
+	/**
+	 * The locale string defining the language and regional formatting for tests.
+	 * This follows the BCP 47 language tag format (e.g., 'en-US'). It ensures consistent formatting
+	 * across tests.
+	 *
+	 * @example 'en-US' // American English
+	 */
+	// eslint-disable-next-line no-var
 	var IMPLEMENTATION: string | undefined;
```
latin-panda, line 34: I didn't find the usage of this `IMPLEMENTATION`

latin-panda:
@eyelidlessness When you have a chance, can you please review it? Thank you!

lognaturel (commented):
Was the root cause of #324 then that the test was written with the assumption that the timezone was Arizona time and because that wasn't set uniformly the device time was being used in some cases?

packages/xpath/test/helpers.ts
```diff
@@ -10,14 +10,33 @@ type AnyParentNode =
...
+	 * related functions.
+	 *
+	 * @example 'America/Phoenix' // Fixed UTC-7, no DST
+	 * @example 'Europe/London' // UTC+0 (GMT) or UTC+1 (BST) with DST
+	 */
 	// eslint-disable-next-line no-var
 	var TZ: string | undefined;
 	// eslint-disable-next-line no-var
+	var LOCALE_ID: string | undefined;
+	/**
```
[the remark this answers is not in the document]
latin-panda: Right! 🤦🏽‍♀️ 
Fixed

packages/xpath/test/helpers.ts
latin-panda: It was added [in this commit ](https://github.com/getodk/web-forms/commit/0d7c7bcf65113396fb7c4eb42c701865b76f421d#diff-afd1a869e5caed19e5fb890a10ad2875720fc73f6c46cd0c0f0fa9745a8c3816). It's unclear what it is for. I'll leave it there for now and ask Trevor later

packages/xpath/test/xforms/now.test.ts
```diff
@@ -14,8 +14,6 @@ describe('#now()', () => {
 	// > including timezone offset (i.e. not normalized to UTC) as described
 	// > under the dateTime datatype.
 	it('should return a timestamp for this instant', () => {
-		// this check might fail if run at precisely midnight ;-)
```
lognaturel, line 17: I think this comment wasn't related to the issue here but was a tongue and cheek remark about how it's possible for `new Date();` to be run milliseconds before midnight and then the `now()` function to be evaluated milliseconds after midnight in which case the test would fail even if the code is all correct. I do agree with removing the comment.

packages/xpath/vite.config.ts
```diff
@@ -83,6 +96,7 @@ export default defineConfig(({ mode }) => {
 		},
 		define: {
 			TZ: JSON.stringify(timeZoneId),
```
lognaturel, line 98: This was intended/expected to set the global timezone for tests, does that sound right? Is the issue then that this doesn't work or at least not consistently? Is some of what's at https://github.com/vitest-dev/vitest/issues/1575#issuecomment-1439286286 related? 

If there was an attempted alternative approach to set a consistent timezone that didn't work, I think it'd be helpful to remove it so there isn't ambiguity about the source of truth.

latin-panda: Good question

The `define` option in the Vite config is for defining global constants that are replaced at build time or made available as globals during development. The test runner doesn't automatically set the timezone by defining the TZ constant.

We still need the constant to setup the xpath's [evaluator option](https://github.com/getodk/web-forms/blob/4ae10116a9636fcbac044ddd18d4e614dad1df05/packages/xpath/test/helpers.ts#L95) and to get the [default locale](https://github.com/getodk/web-forms/blob/4ae10116a9636fcbac044ddd18d4e614dad1df05/packages/xpath/test/helpers.ts#L362) for mocking the timezone and locale in the [global `beforeEach`](https://github.com/getodk/web-forms/blob/4ae10116a9636fcbac044ddd18d4e614dad1df05/packages/xpath/test/setup.ts#L7)

latin-panda: ref: https://vite.dev/config/shared-options.html#define

lognaturel: Thanks for the explanation! I'm not super familiar with these test options. So now I'm seeing that the evaluator timezone was being set but there was nothing previously setting the vitest timezone. From what I've read, using a global `beforeEach` seems like a good approach to do that.

lognaturel (approved):

packages/xpath/vite.config.ts
latin-panda: > Thanks for the explanation! I'm not super familiar with these test options. 

No worries, I'm also learning as I fix this bug

Issue #41, opened 2024-02-20
Consider alternatives to tree-sitter


eyelidlessness:
I’m filing quickly on mobile ahead of our check in, but briefly:

- tree-sitter adds a fair bit of tooling complexity
- And page weight
- It’s unclear that we benefit from its most compelling distinct features like incremental parsing
- While the grammar is stable, it’s a constant build time drag in CI
- Other tools might be better suited for eg TypeScript integration (as in type gen for aspects of the syntax we’re concerned with)

I briefly explored [Ohm](https://ohmjs.org/) over the weekend and it’s a pretty compelling alternative option (albeit with some reservations around maturity). Beyond its TS support, I was encouraged by how easy it was to build a cromulent XPath grammar, and how much more closely it matches the grammar from the spec.

eyelidlessness:
> [ANTLR4](https://github.com/antlr/antlr4) seems to be quite popular, it has typescript support as well and XPath grammar already written.

Specifically I’m interested in support for generating (or otherwise producing) TypeScript types corresponding to the parsed syntax tree nodes. We use [dts-tree-sitter](https://github.com/asgerf/dts-tree-sitter) for this currently. It’s not amazing but we rely on it and supplemental types built on that to keep the parser and evaluator in sync. That’s something that Ohm does OOTB, and my impression from the spike does it better.

I’m not sure if that’s something ANTLR4 supports, but I didn’t find anything about that when I looked last. It’s not a deal breaker by any means, as the grammar will likely be stable. But it’s something I’m particularly nervous about trying to augment for an existing parser without really being familiar with its output.

eyelidlessness:
It's worth mentioning now that #166 is a really good frame of reference for why a typed grammar is valuable for us. Quite a lot of that change would be much more complicated without the types we have to express which syntax nodes can occur where in the syntax tree.

It's also worth mentioning that I recently did a spike into ANTLR4's XPath grammar. It seems like it partially addresses these concerns: ANTLR4 has a TypeScript target which produces _some grammar-specific types_. Unfortunately, what I found was that the generated parse API would require significant rework for us to use... basically across the board, anywhere we use XPath syntax as anything other than an opaque string. The API is heavily influenced by its Java underpinnings, and heavily generalized around OOP-ish concepts that don't really fit our current model of parsed syntax (as a tree of nodes which are basically plain objects).

eyelidlessness:
[...]

PR #327, opened 2025-03-05
Add config for running web-forms via containers

spwoodcock:

- [x] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Not applicable

- I successfully uploaded a form, [attached](https://github.com/user-attachments/files/19097538/project_form_1.xlsx)
- The form is processed by xlsform-online and then viewed in web-forms preview.

- It's probably not the best possible solution, but it's one option that could make running web-forms easier for devs.

No change.

See above.

## What's changed

- Related to the comment here https://github.com/getodk/web-forms/issues/41#issuecomment-2393935846
- I had trouble setting up web-forms for development due to the tree-sitter library needing WASM compilation.
  - As mentioned in the linked issue, doing this requires either docker or emscripten emsdk.
  - I run almost everything inside containers, meaning docker is not installed unless I used DIND.
  - Installing emsdk would also be ephemeral, unless I decided to add to my dev stack.
- While I admit these issues may be somewhat specific to my config, the linked issue does suggest adding a pathway to build tree-sitter.wasm via docker & this setup does that!

What I included:
- Updating the default ports for the xlsform-online service and web-forms to less standard ports, that are less likely to be in use on a dev machine: 8000 --> 8558 & random port assigned by vite --> 5173 (also used for playwright devserver).
- Adding a Containerfile that builds the packages (including the tree-sitter.wasm file) and installs required dependencies.
- A compose file that runs both xlsform-online and web-forms preview together.
- I added all these files under `contrib/containers`, suggesting it's not maintained as part of the core library (not officially maintained / guaranteed to work). I'm not sure if everyone is as keen on containers as I am, so thought this most appropriate.

Alternatives:
- Alternatively, only the tree-sitter.wasm compilation could be containerised, output to the local development machine, then `yarn` used as normal for local development.
- However this approach wouldn't also add xlsform-online, which I think is quite helpful to bundle.

Open to suggestions of alternative approaches, or feel free to close this PR if it's not useful 😄 

**Update 15-04-2025**:
- I also needed to run tests, so added a small config to run tests via the container too.
- The advantage here is that users don't need to install ~5GB of dependencies, including 3 web browsers to playwright.


lognaturel:
Thanks so much for sharing this and I'm sorry we didn't give you feedback before now!

I'm going to close this because we're in the process of significantly reworking how Web Form's code is organized. In particular, we've spun https://github.com/getodk/tree-sitter-xpath out as a separate repo so that building the Web Forms packages on top of it don't require rebuilding the native dependency. We don't expect that package to change much.

We will be moving the rest of the packages into the Central Frontend repo shortly. We'll continue publishing them as we do now but will develop them in the frontend monorepo so that we can more easily have end-to-end testing with Central which is our true integration target. We'll spin the preview out into the website repository.

We'll post more details about all of this on the forum shortly but wanted to give you some context here first.

PR #328, opened 2025-03-08
feat(#311): adds support for default date question type

### LINKED ISSUE #311: Add default date question type
Implement the default date widget with the Gregorian calendar. 

Support for appearances and additional calendar types—such as Bikram Sambat, Buddhist, Coptic, Ethiopian, Islamic, Myanmar, and Persian—is considered a future enhancement and will not be included at this stage

### User stories

As a user, I want to navigate past or future dates and select a date.

### Component state
- Read-only state 
- Disabled state
- Display of errors

### Libraries

Use the [Primevue's date picker](https://v3.primevue.org/calendar/) component and [figma](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=1381-19538&t=Q0NFu7hT2z2Ajcal-1) for states not in the library. The calendar's footer has an option [with some navigation buttons,](https://primevue.org/datepicker/#button) but we need to check if it's possible to translate those buttons. 

### Design
[Figma](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=1381-12633&p=f&t=jIwntIcjyv8Tv27E-0)

Image

latin-panda:
Closes #311

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [x] Safari (iOS)
- [x] Chrome for Android
- [ ] Not applicable

- [Test plan](https://github.com/getodk/web-forms/issues/311#issuecomment-2685672916)

Test evidence:

<details><summary> Browser compatibility: Chrome, Firefox, Safari (Desktop) </summary>

- Question type becoming relevant 
- Question type as required 
- Clearing question type value 

https://github.com/user-attachments/assets/62bfc715-05c7-4c2b-9464-4d7382a4c155

</details>

<details><summary> Date Notes </summary>

Note values:
- `<date_note>2025-12-21T23:30:05</date_note>`
- `<date_note>2025-12-21</date_note>`

Screenshot 2025-03-10 at 9 28 57 PM

</details>

<details><summary> Responsive and readonly </summary>

- Responsive
  - The calendar closes when resizing the screen
  - The calendar might not fit on the screen but allows scrolling to see more of the calendar 
  - The read-only state won't allow editing the question type

https://github.com/user-attachments/assets/d56caa16-d457-449d-8ea8-0e7d11882d5a

</details>

<details><summary> Allows navigating across years and months. Prevent setting invalid dates. </summary>

https://github.com/user-attachments/assets/d3ee76cc-54c8-4160-8c95-e8d9d9fb48fc

</details>

<details><summary> Compatibility with Chrome for Android</summary>

https://github.com/user-attachments/assets/c07f8f7e-5144-4ff5-bcaf-f0942b78feda

</details>

<details><summary> Form's error message is closable and doesn't overlap calendar </summary>

https://github.com/user-attachments/assets/cb936147-155b-45e5-94ca-82f556bcad1e

</details>

- Datetimes with a timezone offset will result in a null / empty string. We need user research to determine if it should honor the timezone (in local time, it could be one date back or forward) or truncate to the date part (yyyy-mm-dd) as Enketo does. 

A test plan is in the ticket, and the test evidence is in the previous section.

### What's changed

- The `@js-temporal/polyfill` library was replaced with `temporal-polyfill`
- Date constant file was relocated from the 'packages/xpath' to packages/common/src/constants/datetime.ts
- A new demo form for dates was added, and also the date note was added to the note's demo form
- New validation was added to `packages/xpath/src/lib/datetime/coercion.ts` to prevent out-of-range offsets 
   - Webkit (Safari) parses invalid offsets like "-24:00" while Chrome and Firefox reject them.
- Adds a new Vue component for the Input Date and adds support for date notes.
- Adds a new Date encode and decode that interprets date like or datetime like without timezone offset,  to date only (`yyyy-mm-dd`).
- Adds Z-index Layer System with documentation 
  - Error banner doesn't overlap the calendar and hamburger menu
  - Feedback button doesn't overlap the dialog

Note: Calendar's translation is to be resolved in this ticket: https://github.com/getodk/web-forms/issues/332


latin-panda:
@eyelidlessness 
There will be a bit of change in the CSS
- The calendar is showing behind the floating error message from the top - I'm working on figuring out a fix (so far, `z-index` with a different `position` doesn't seem to work) 
- Maybe some adjustments in the disabled style - waiting for feedback from Aly

I'll open a separate PR for the e2e tests since we haven't merged the other PR that introduces the structure (https://github.com/getodk/web-forms/pull/313) 

Other than that, this is ready for a review. Please have a look when you have time 🙂

latin-panda:
PR is up to date with the `main` branch.

packages/scenario/test/bind-types.test.ts
```diff
@@ -616,6 +644,62 @@ describe('Data (<bind type>) type support', () => {
 				}
 			);
 		});
+
+		describe('type="date"', () => {
```
latin-panda, line 648: This is a good place to start reviewing and see if the encode/decode is doing what we expect.

packages/web-forms/src/components/controls/Input/InputDate.vue
```diff
@@ -0,0 +1,49 @@
...
+			return null;
+		}
+
+		const temporalValue = props.question.currentState.value.toString();
+		if (!ISO_DATE_LIKE_PATTERN.test(temporalValue)) {
+			return null;
+		}
+
+		// Convert to ISO string (yyyy-mm-dd) and append time for start of day local
+		return new Date(temporalValue + 'T00:00:00');
```
latin-panda, line 25: PrimeVue's date picker doesn't support temporal, so we create a Date object at the start of the local day

packages/xpath/src/context/Context.ts
```diff
@@ -37,5 +37,5 @@ export interface Context<T extends XPathNode> {
 
 	readonly functions: FunctionLibraryCollection;
 	readonly namespaceResolver: NamespaceResolver<T>;
-	readonly timeZone: Temporal.TimeZone;
+	readonly timeZone: Temporal.TimeZoneLike;
```
latin-panda, line 40: Temporal defines `TimeZoneLike` like this:

```js
  /**
   * Any of these types can be passed to Temporal methods instead of a time zone ID.
   */
  type TimeZoneLike = string | ZonedDateTime;
```

In case this is useful, [ZonedDateTime definition](https://github.com/fullcalendar/temporal-polyfill/blob/bc1baf3875392ecd8522d40e7eecb55fa582808c/packages/temporal-spec/index.d.ts#L979).

latin-panda:
@lognaturel can you please have a look when you have time? Thanks!

packages/web-forms/README.md
```diff
@@ -81,6 +81,21 @@ We use CSS variables for theming with two prefixes:
 - `--p-` Prefix: Variables prefixed with `--p-` (e.g., `--p-primary-50`) come from PrimeVue and control its component styles (e.g., colors, borders). These are part of PrimeVue’s [theming system](https://primevue.org/theming/styled/).
 - `--odk-` Prefix: Variables prefixed with `--odk-` (e.g., `--odk-font-family`) are custom to this project and manage styles outside PrimeVue or PrimeFlex, such as application-specific typography.
 
+#### Z-Index Layering System
```
latin-panda, line 84: Maybe I can remove this documentation; the `style.scss` is already enough. This is good for onboarding new developers and better in another document/file with other stuff

lognaturel:
> The @js-temporal/polyfill library was replaced with temporal-polyfill

Was this for size/perf reasons? Something else?

packages/web-forms/README.md
lognaturel: Seems fine to leave here for now and spin out if the readme gets too long.

latin-panda:
> Was this for size/perf reasons? Something else?

It's something Trevor requested a while ago. Yes, it's a lightweight option, and since their focus is to keep it light, it may mean fewer breaking changes or breaking changes that get optimization benefits. There's a comparison [table here](https://github.com/fullcalendar/temporal-polyfill?tab=readme-ov-file#comparison-with-js-temporalpolyfill). 
The [js-temporal/polyfill](https://github.com/js-temporal/temporal-polyfill/blob/5830124d0ea0b6229a32af8708829a8b5cade9bd/lib/calendar.ts#L2394) and [temporal-polyfill](https://github.com/fullcalendar/temporal-polyfill/blob/bc1baf3875392ecd8522d40e7eecb55fa582808c/packages/temporal-polyfill/src/internal/calendarConfig.ts#L31) support many calendar systems, like Persian, Buddhist, Islamic, etc. The new one doesn't rely on BigInt, but since we aren't supporting old browsers, the dependency on BigInt is not a deciding factor. 
So, the bundle size is the immediate benefit we get.

packages/scenario/test/bind-types.test.ts
```diff
@@ -616,6 +644,62 @@ describe('Data (<bind type>) type support', () => {
...
+				expect(answer.value).toBeNull();
+				expect(answer.stringValue).toBe('');
+			});
+
+			it.each([
+				'13:30:55',
+				'2025-23-23',
+				'ZYX',
+				'2025-03-07T14:30:00+invalid',
+				'2025-03-07T14:30:00-08:00',
```
lognaturel, line 676: From PR description: "Datetimes with a timezone offset will result in a null / empty string. We need user research to determine if it should honor the timezone (in local time, it could be one date back or forward) or truncate to the date part (yyyy-mm-dd) as Enketo does."

packages/web-forms/src/components/controls/Input/InputDate.vue
```diff
@@ -0,0 +1,49 @@
...
+const value = computed({
+	get: () => {
+		if (props.question.currentState.value == null) {
+			return null;
+		}
+
+		const temporalValue = props.question.currentState.value.toString();
+		if (!ISO_DATE_LIKE_PATTERN.test(temporalValue)) {
+			return null;
+		}
```
lognaturel, line 22: Given that we know the question is a `DateInputNode`, can't we use `value` directly and know that will be a `DatetimeRuntimeValue`? Or use `decodeValue` to get a string rather than `toString`?

Maybe the functionality to convert to Date should also be in the engine? Less sure about this one because it really is just for the UI.

packages/web-forms/src/components/controls/NoteControl.vue
```diff
@@ -47,8 +47,10 @@ const value = computed<NoteRenderableValue>(() => {
 		case 'geopoint':
 			return question.currentState.value;
 
-		case 'boolean':
 		case 'date':
+			return question.currentState.value?.toString();
```
lognaturel, line 51: Use the codec?

packages/web-forms/src/components/controls/Input/InputDate.vue
latin-panda: Using a string (export type DatetimeRuntimeValue = string | null;) is fine for the current needs of the Web Form client, since we don't expect to perform any calculations where the precision of T`emporal.PlainDate` would have been nice.

The original intention was to expose the Temporal classes and let the front-end handle things as needed, providing more flexibility.

What do you think?

latin-panda: Given that Temporal is not natively available in Browsers yet. Now I'm more inclined not to expose any Temporal class to the clients, to reduce incompatibilities and put us in a weird spot of support. 

In theory, we control these things more since we manage central releases.

lognaturel: I see the benefits of having the client work with strings for now. To get that string, can we use the codec's `decodeValue`?

latin-panda: Yes, we can. I'm thinking of returning `YYYY-MM-DDTHH:MM:SS` even when it's a "Date" only (the time is: `T00:00:00`), which will reduce further checks in the front-end.

packages/web-forms/src/components/controls/NoteControl.vue
latin-panda: Fixed

lognaturel (approved):
Thanks for exploring some alternative ideas for type conversion for client presentation. Let's go with this for now and keep exploring alternative ideas as we support more types.

PR #329, opened 2025-03-10
Add `stepper-layout` prop to OdkWebForm component for Collect-like experience

spwoodcock:

- [x] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Not applicable

- Tested the original 'list' view layout works too.

- I tried my best to not add new superfluous code.
- It's mostly either:
  - Using built-in PrimeVue components.
  - Using existing functionality built into web-forms already (such as the validation code).

- No change for the default view - the stepper has to be enabled by a prop.

I was testing with this one:
[mdp_fmtm_xlsform.xlsx](https://github.com/user-attachments/files/19163618/mdp_fmtm_xlsform.xlsx)

### Included

- I added a new component `QuestionStepper` that mirrors the `QuestionList` component.
- It's enabled by simply passing the `stepper-layout` prop to `OdkWebForm`.
- Instead of displaying questions and groups via collapsible panels, instead a PrimeVue Stepper component is introduced, displaying one question or group of questions per page.
- I kind of abused the `submitPressed` dependency to trigger the existing validation logic, without needing to add extra code - it seems to work nicely.
- Video is in desktop view for easier recording, but mobile view is the preferred way to use the stepper:

https://github.com/user-attachments/assets/132bafd7-0177-4a1a-9b09-cb39d5ee0669

### Things I didn't do

  - [ ] I updated the `FormPanel` to have an optional `toggleable` prop. This allows us to disable the panel collapsing behaviour for the stepper. I didn't update RepeatInstance and RepeatRange to have a toggleable prop passed through to FormPanel too, but assume this may be needed?
  - [ ] I attempted to add a PrimeVue ProgressBar to the stepper. It works well in the layout, however, it's pretty tricky to assess the percentage complete for the form. This needs to be done via the available nodes, but based on `relevant` fields, additional nodes can be added causing the percentage complete to drop. I need to look into this in more detail - it's probably doable - if this is a desired feature.
- [ ] I had to hide the 'Powered by ODK' button when using the stepper, as it has absolute positioning below the question panel. This could be fixed somehow if needed.

### Note on the FormPanel border

  - I removed this entirely for the stepper based view, as it's not necessary imo:
    Stepper view:
    image
    List view:
    image
  - I kept the left border on collapsible FormPanel as it nicely demarcates where each group starts and ends.
    - I assume that was why it is present? Else, easily removable via `box-shadow` CSS.

### Extra notes

- Styling is a mix of https://www.figma.com/design/KFi9hIQdrRgo0rLqIC7Sg8/ODK-Collect?node-id=5504-114737 and trying to keep in line with Collect - let me know your thoughts! @lognaturel @DanielJDufour 
- The send button displays in center screen on Collect, but here I added it in place of the next button on the stepper instead.
  - This could be changed, but it complicates things a bit.

### DIFF
--- a/packages/web-forms/src/components/FormPanel.vue
+++ b/packages/web-forms/src/components/FormPanel.vue
@@ -12,6 +12,7 @@ export interface PanelProps {
 	class?: string[] | string;
 	labelIcon?: string;
 	labelNumber?: number;
+    toggleable?: boolean;
 }
 
 const props = withDefaults(defineProps<PanelProps>(), {
@@ -21,6 +22,7 @@ const props = withDefaults(defineProps<PanelProps>(), {
 	class: undefined,
 	labelIcon: undefined,
 	labelNumber: undefined,
+    toggleable: false,
 });
 
 const panelClass = computed(() => [
@@ -31,7 +33,9 @@ const panelClass = computed(() => [
 const panelState = ref(false);
 
 const toggle = () => {
-	panelState.value = !panelState.value;
+    if (props.toggleable) {
+		panelState.value = !panelState.value;
+    }
 };
 
 const menu = ref<Menu & MenuState>();
@@ -40,10 +44,11 @@ const toggleMenu = (event: Event) => {
 	menu.value?.toggle(event);
 };
 </script>
+
 <template>
-	<Panel v-if="!noUi" :class="panelClass" :toggleable="true" :collapsed="panelState">
+    <Panel v-if="!noUi" :class="[panelClass, { 'toggleable-enabled': toggleable }]" :toggleable="toggleable" :collapsed="toggleable ? panelState : false">
 		<template #header>
-			<div class="panel-title" role="button" @click="toggle">
+            <div v-if="toggleable" class="panel-title" role="button" @click="toggle">
 				<h2>
 					<span class="chevron" :class="panelState ? 'icon-keyboard_arrow_down' : 'icon-keyboard_arrow_up'" />
 					<span v-if="labelNumber" class="label-number">{{ labelNumber }}</span>
@@ -51,6 +56,13 @@ const toggleMenu = (event: Event) => {
 					<span v-if="labelIcon" class="ml-2" :class="labelIcon" />
 				</h2>
 			</div>
+            <div v-else>
+                <h2>
+                    <span v-if="labelNumber" class="label-number">{{ labelNumber }}</span>
+                    <span>{{ title }}</span>
+                    <span v-if="labelIcon" class="ml-2" :class="labelIcon" />
+                </h2>
+            </div>
 		</template>
 		<template v-if="menuItems && menuItems.length > 0" #icons>
 			<Button severity="secondary" rounded class="btn-context" :class="{ 'p-focus': menu?.overlayVisible }" icon="icon-more_vert" aria-label="More" @click="toggleMenu" />
@@ -129,15 +141,19 @@ h2 {
 	}
 
 	:deep(.p-panel-content) {
-		border-left: 2px solid var(--gray-200);
+		box-shadow: none;
 		margin-left: 10px;
-		border-radius: 0;
 		padding: 0 0 0 1.5rem;
 	}
 
 	:deep(.p-panel-toggler) {
 		display: none;
 	}
+}
+
+.toggleable-enabled :deep(.p-panel-content) {
+	border-left: 2px solid var(--gray-200);
+	border-radius: 0;
 }
 
 .content-wrapper {
--- a/packages/web-forms/src/components/QuestionList.vue
+++ b/packages/web-forms/src/components/QuestionList.vue
@@ -45,7 +45,7 @@ const isControlNode = (node: NonStructuralNode): node is ControlNode => {
 	<template v-for="node in nodes" :key="node.nodeId">
 		<template v-if="node.currentState.relevant">
 			<!-- Render group nodes -->
-			<FormGroup v-if="isGroupNode(node)" :node="node" />
+			<FormGroup v-if="isGroupNode(node)" :node="node" toggleable />
 
 			<!-- Render repeat nodes -->
 			<RepeatRange v-else-if="isRepeatRangeNode(node)" :node="node" />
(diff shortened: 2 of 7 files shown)


packages/web-forms/src/components/QuestionList.vue
```diff
@@ -48,7 +48,7 @@ const isControlNode = (node: NonStructuralNode): node is ControlNode => {
 	<template v-for="node in nodes" :key="node.nodeId">
 		<template v-if="node.currentState.relevant">
 			<!-- Render group nodes -->
-			<FormGroup v-if="isGroupNode(node)" :node="node" />
+			<FormGroup v-if="isGroupNode(node)" :node="node" toggleable />
```
spwoodcock, line 51: I set the default for the prop `toggleable` to `false`, meaning the prop needs to be passed here to allow for the collapsable panels in the list view

packages/web-forms/src/components/FormPanel.vue
```diff
@@ -140,6 +151,11 @@ h2 {
 	}
 }
 
+.toggleable-enabled :deep(.p-panel-content) {
```
spwoodcock, line 154: Here I only add the style for the line on the left of the group if the panel is set to `toggleable`

Screenshot 2025-03-10 135310

packages/web-forms/src/components/FormPanel.vue
```diff
@@ -40,17 +44,25 @@ const toggleMenu = (event: Event) => {
 	menu.value?.toggle(event);
 };
 </script>
+
 <template>
-	<Panel v-if="!noUi" :class="panelClass" :toggleable="true" :collapsed="panelState">
+    <Panel v-if="!noUi" :class="[panelClass, { 'toggleable-enabled': toggleable }]" :toggleable="toggleable" :collapsed="toggleable ? panelState : false">
```
spwoodcock, line 49: Adding the `toggleable-enabled` class dynamically

lognaturel:
Thanks so much for exploring this and sharing what you've come up with! The video looks so good and it's great to see that the implementation is largely contained.

I'm going to change this to a draft for now to help with our review processes. We'd want to make sure there are tests and a design pass before considering for merge.

I think it makes sense for us to prioritize image capture and selects with images over reviewing and finalizing this, does that sound right?

spwoodcock:
> Thanks so much for exploring this and sharing what you've come up with! The video looks so good and it's great to see that the implementation is largely contained.
> 
> I'm going to change this to a draft for now to help with our review processes. We'd want to make sure there are tests and a design pass before considering for merge.
> 
> I think it makes sense for us to prioritize image capture and selects with images over reviewing and finalizing this, does that sound right?

Yep sounds good to me! We can finalise and discuss this further in the long term - I'll add tests too 😁

We can use this layout on our fork for the time being 👍

spwoodcock:
Note to self (sorry for the ping).

If we have groups within groups, then then breaks down a bit:

image
(ignore the funny column-pack appearance in the screenshot - unrelated)

- This PR will only consider one group level, placing one group on one page.
- If then is a group within a group, then they are all displayed on one page, with the default collapsible nesting for the second group.

We should consider what to do in this scenario 😄

alyblenkin:
Thanks for sharing @spwoodcock. I'm working on some new iterations for groups and repeats – trying to find a way to reduce the visual noise of the lines, while still showing nesting. This redesign work will come a bit later in the roadmap, but we are starting to think about it!

Screenshot 2025-04-17 at 8 27 40 AM

spwoodcock:
Amazing!

I can already see some new design choices in the latest Web Forms - it looks very nice 👍

---

Another consequence of this PR is having to consider translations for the 'Back', 'Next', and 'Submit' buttons.

I believe Collect doesn't do this currently anyway, but we had feedback that its confusing for users having an English button when they speak Portuguese 😅

I saw the other designs that were using icons instead & I'm sure you are considering this already, just food for thought!

alyblenkin:
> I can already see some new design choices in the latest Web Forms - it looks very nice 👍

So glad you like the changes! If you get any feedback from users on the new designs let us know. 

> I believe Collect doesn't do this currently anyway, but we had feedback that its confusing for users having an English button when they speak Portuguese 😅

Do you mean in Collect the button label says "Change language" rather than the language selected? 

> I saw the other designs that were using icons instead & I'm sure you are considering this already, just food for thought!

You're right we've tried a few versions, but adding the label does make it more discoverable. It's good feedback!

spwoodcock:
> Do you mean in Collect the button label says "Change language" rather than the language selected? Or maybe English is the default language when it should be Portuguese?

No sorry, I just mean that Collect doesn't translate the 'Next', 'Back', 'Submit' UI buttons when you select a language.
It's English only for the UI, with translations possible in the form question only.

I wanted to keep the experience as 'Collect-like' as possible in this PR, but I'm actually leaning more in favour of buttons like this:
image
(it would save the UI translation headache)

alyblenkin:
@spwoodcock oh sorry! I misunderstood! That makes sense

lognaturel:
Thanks again for sharing this, @spwoodcock! We decided this functionality was not a must-have for making ODK Web Forms the default web forms client in Central so we deprioritized it. We're about to make it the default for new forms and are getting close to looking at this again!

I'm going to close this PR for now for the reason described at https://github.com/getodk/web-forms/pull/327#issuecomment-4425937071

spwoodcock:
Thank you @lognaturel!!

As a quick update from our side:
- We made a big push to use web-forms on mobile devices, as I though (and still think) this would make a really nice mapping app experience (especially as it includes iOS, and has the potential to pull in other JS local-first libraries).
- However, as much as I would like to pursue that further, we decided that maintaining a mapping app was too much of a maintenance burden & didn't have enough developer time to make a success of it.
- I might return to this in the future, but for now we decided to focus on making a 'coordinator only' and relying on ODK Central or other tools to track all of the data and task statuses (embedding an entity property for task `status` per project, then updating the status field as part of the form submission).

Would love to touch bases at some point, to better optimise our ODK Collect workflow in FieldTM - I'm sure there is functionality I haven't kept on top of over time that would really benefit the coordination of 10's - 100's of mappers using ODK! Thanks for engaging so much with this - we learn many valuable lessons that have shaped our current approach🙏

Issue #316, opened 2025-02-26
Prepare a new instance for filling from an active form definition and form attachments


lognaturel:
## User story
As the developer of a host application,
I want the option to request preparing a new instance for form filling after a user has made a submission
So that I can support a workflow that requires filling a form multiple times

As a form end user,
I want the option to immediately fill out another instance of the active form after submitting
So that I can quickly carry out my next task

## Notes
* This is the current behavior from Central when accessing form filling from Submissions > New
* This needs to be opt-in -- there are many cases in which the host application will want to redirect or do something else after submission
* Preparing a new instance involves resetting all values to defaults, triggering all preloads, and triggering `odk-instance-first-load`.

lognaturel:
Closed by https://github.com/getodk/web-forms/pull/355

```
<script setup>
import { POST_SUBMIT__NEW_INSTANCE } from '@getodk/web-forms';

// Given whatever existing logic handles the submission data, I/O, etc...
declare const submitPayload: (payload) => Promise<any>;

const onSubmit = (payload, callback) => {
  callback(submitPayload(callback).then(() => {
    return { next: POST_SUBMIT__NEW_INSTANCE };
  }));
}
</script>

<template>
<OdkWebForm @submit="onSubmit" ... />
</template>
```

Issue #247, opened 2024-11-09
Epic: Submission edit


alyblenkin:
## User stories

As a project manager
I want to edit a submission 
So that I can adjust or fix any mistakes

## Sub-tasks related to this Epic 
More sub-tasks may be identified later in the implementation 
- [x] https://github.com/getodk/web-forms/issues/299
- [x] https://github.com/getodk/web-forms/issues/36
- [x] https://github.com/getodk/web-forms/issues/37
- [ ] https://github.com/getodk/web-forms/issues/314

lognaturel:
Now that #355 is in and released, let's close this! #314 feels like it stands alone. We may want to track it as part of a different epic that's more like "communicate WF state to user" that would involve indicating when a form is a preview, edits, etc.

PR #335, opened 2025-03-12
[Edits] Separate form load and instance creation in engine entrypoints

eyelidlessness:
Part 2/4 in preparation for #247.

Also addresses most of #316. Has a _little bit of a hint_ toward how we might start approaching #202.

Branched from #334 ([diff](https://github.com/getodk/web-forms/compare/edit-epic/instance-terminology...edit-epic/engine-entrypoint-split)).

- [x] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [x] Not applicable

I don't think there's anything browser specific here.

As of the commits in this branch, most of the verification was done by a combination of type check, tests, and manual spot checking. The feature is exercised more by demo-focused changes integrating downstream work (PRs incoming!).

The approach here was to identify _in the existing `initializeForm` flow_ where "load form" and "create instance" conceptually split, and to introduce an interface boundary between the two **without exposing much more than that** outside of the engine.

An alternative that's definitely worth considering (and working toward) is to expose more detail _about the form itself_ from the engine to its clients.

We may also want to introduce _another split_: between parsing the _form definition_ and loading `jr:` URL resources, and also exposing more detail _about the form itself_ to that end. (I could've sworn we had an issue for this, but I'm having trouble finding it now! I'd bet it's probably at least mentioned in commentary on #201.)

I think the biggest risk this introduces is **potential** for growing memory usage over time. This change doesn't introduce any kind of interface for cleanup/teardown, for form and instance resources alike. A client could **hypothetically** load any number of forms, and for each form create any number of instances, in a given session; if the client held onto all of the references produced by that, they'd have N forms × M instances in memory.

But I think it's important to caution: this risk is _entirely theoretical_, and it's not consistent with the workflows we anticipate! In those workflows we'd tend to either hold onto 1 form reference (or in one case I can think of, very-small-N form references) and 1 instance reference (or in the same case, 1-or-very-small-N forms × 1-or-very-small-M instances) at any given time.

I'd also mention that we do have some mitigations we can already leverage here: we already `dispose` computations on removed nodes (repeat instances and their descendants) in the engine, and we can expand that to release quite a lot of memory which might otherwise be hard to GC.

N/A

## What's changed

Explicitly supports the workflow described in #316.

Note: I considered demoing this in the UI by showing a 2-up pair of instances from the same form. If that would be useful I'd be happy to whip that up! Shouldn't take more than a few minutes.


.changeset/ten-wolves-reflect.md
```diff
@@ -0,0 +1,13 @@
...
+
+**BREAKING CHANGE**
+
+The main engine entrypoint (formerly `initializeForm`) has been split into:
+
+- `loadForm`, producing an intermediate result from which many instances of the same form can be created (with a `createInstance` method on that result)
+
+- `createInstance`, a convenience wrapper composing the result from `loadForm` and its `createInstance` method to create a single instance; this entrypoint effectively restores the behavior of `initializeForm`
+
+Some interfaces related to the former `initializeForm` have also been refined to reflect this change.
```
eyelidlessness, line 13: The other changesets currently in the diff are part of #334 (and as I just mentioned there, I think the PR warrants a broader one as well). But I am pretty sure this changeset adequately covers the delta from that branch to this one!

packages/xforms-engine/src/error/LoadFormFailureError.ts
```diff
@@ -0,0 +1,114 @@
...
+
+				return `- ${aggregatedMessage}`;
+			}),
+		];
+
+		if (rawData != null) {
+			messageLines.push('\n- - -\n', 'Raw resource data:', rawData);
+		}
+
+		const message = messageLines.join('\n');
```
latin-panda, line 100: Suggestion:

I’d suggest refactoring this into a separate function. The message’s construction follows a specific order that we can preserve, and keeping it inline in the constructor could lead to errors (while making it harder to read)

packages/xforms-engine/src/error/LoadFormFailureError.ts
```diff
@@ -0,0 +1,114 @@
...
+
+		super(errors, message);
+
+		const [head, ...tail] = errors;
+
+		if (head != null && tail.length === 0) {
+			const { stack } = head;
+
+			if (typeof stack === 'string') {
+				this.stack = stack;
```
latin-panda, line 110: Question:

Could this stack expose sensitive information we’d need to handle for privacy compliance?
For example, should we redact specific patterns from the stack trace to protect data?

latin-panda (approved):
Very nice to see this coming through, a lot of thinking here ✨
I left a couple of suggestion and questions

packages/xforms-engine/src/error/LoadFormFailureError.ts
eyelidlessness: I'm not sure I understand the suggestion, or the reasoning for it. The logic here is constructing a set of message lines, in order to populate the `message` parameter to `super` (and ultimately the error's own `message` property). The logic I see as having significant complexity is already extracted to a separate function (`formResourceMetadata`, called here). The only logic remaining inline is building a list:

- With a prefix
- With an item for each aggregated input item
- With a suffix, if there's a basis for the suffix

I'm not opposed to calling that complex enough for a function, and I'm more than happy to extract it! What I'm not following is...

> keeping it inline in the constructor could lead to errors (while making it harder to read)

Are you imagining a more complex evolution of this class over time? As I hoped to convey in the JSDoc/`@todo` on the class itself, I think it's much more likely for the whole class be removed/replaced with much more thoughtful models of much better known error conditions. If it gets any more complex than this, I'd raise alarms that we're putting too much effort on the wrong part of (avoiding) #202.

packages/xforms-engine/src/entrypoints/loadForm.ts
```diff
@@ -0,0 +1,154 @@
...
+ *   the engine drops access to the loaded form
+ *
+ * **IMPORTANT:** this **MUST** be called synchronously. If it is called in an
+ * `async` function, it **MUST** be called before any `await` expression; if it
+ * is called in any other flow with mixed synchrony, it must be called before
+ * yielding to the event loop. Failing to do this will cause the engine to lose
+ * access to a client's Solid reactive context, potentially leaking form
+ * reactivity indefinitely.
+ */
+const createPotentiallyClientOwnedReactiveScope = (): ReactiveScope => {
```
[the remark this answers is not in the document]
eyelidlessness: I'm also not fond of this name, but it's... more direct about what it does and _why_, than anything I could come up with. It's awkward, but it's precise:

1. Create a `ReactiveScope`
2. Where it's explicitly **important** that the `ReactiveScope` is _potentially_ (but not necessarily!) client-owned

Which is to say: "potentially client-owned" is the _reason_ I finally decided this needed to be extracted into a function: to give its potentially client-owned-ness a name and to protect that intent from future mistakes. It's the reason the `ReactiveScope` must be created when it's created. And losing track of that intent as the invocation moves around is a known risk factor that I wanted to make more clear (after messing it up... again).

I don't think it's "client aware". At least, it's not aware of anything about a client _other than_ the fact that a client may invoke `loadForm` within an active Solid reactive scope (i.e. the engine's `ReactiveScope` may _potentially be client-owned_).

I definitely don't think it's "hybrid". If anything, that is a more accurate description of a _non-client-owned_ `ReactiveScope`... as in it may be incorporated into _another_ reactivity system, with observable state between the engine and client being managed as a hybrid of two reactivity implementations.

- - -

Like I said, I'm not thrilled about this name! I have some better ideas in mind, in the vein of "make illegal states unrepresentable". That'd look similar to how we model quite a lot of things that have different requirements at an outer boundary and within inner layers, i.e. probably something like:

- An `abstract` base class for shared `ReactiveScope` concepts in the engine;
- some "root"-like (outermost, parentless) `ReactiveScope` subclass, with certain guarantees about its interface with the outside word (client)
- some "descendant"-like (parented/traceable up to the "root"-like scope) `ReactiveScope` subclass, with different guarantees (i.e. that it isn't the outermost and doesn't have responsibilities to the outside world)

I want to do this! But I'm hesitant to bring it into this scope. Even writing all of the above feels like time I wish I'd spent writing it up as an issue to unblock this. I'm happy to file that issue if you think we can live with this awkward name in the meantime?

packages/xforms-engine/src/error/LoadFormFailureError.ts
latin-panda: The suggestion is on the line of having minimalistic constructors and leaving the "building" process to methods so new code doesn't get in between the lines of the "building" message / stack process. 

```
const buidXYZMessage() => {
const messageLines: string[] = [
			`Failed to load form resource: ${description}`,
			'\n',

			...errors.map((error) => {
				const aggregatedMessage = error.message;

				if (aggregatedMessage == null) {
					return '- Unknown error';
				}

				return `- ${aggregatedMessage}`;
			}),
		];

		if (rawData != null) {
			messageLines.push('\n- - -\n', 'Raw resource data:', rawData);
		}

		return messageLines.join('\n');
}

class MyClass {

private setStack() => {
const [head, ...tail] = errors;

		if (head != null && tail.length === 0) {
			const { stack } = head;

			if (typeof stack === 'string') {
				this.stack = stack;
			}
		}
}

constructor(resource: FormResource, errors: readonly Error[]) {
		const { description, rawData } = formResourceMetadata(resource);
		const message = buidXYZMessage(description, rawData);
		
                super(errors, message);

                 this.setStack(errors);
	}
```

I find this straightforward way to make sense of a constructor at a high level faster. 
Then later, if I need to understand what makes a `message` I'll check the buidXYZMessage function.

packages/xforms-engine/src/error/LoadFormFailureError.ts
eyelidlessness: > Could this stack expose sensitive information we’d need to handle for privacy compliance?

Maybe? I doubt it? It's an `Error` call stack, captured automatically by the JavaScript runtime.

> For example, should we redact specific patterns from the stack trace to protect data?

Maybe? I'm less sure than "I doubt it", but I don't know if this should be the trigger for determining what we should redact. If it's sensitive here, it was already sensitive before this wrapper was introduced. And we had much less control over it before we were capturing it as a value.

Which will be more and more true as we address #202 in earnest. Maybe "is `Error.stack` potentially sensitive information?" is a question best suited for the broader error story?

packages/xforms-engine/src/entrypoints/loadForm.ts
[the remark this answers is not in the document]
latin-panda: I can live with the awkward name for now :) Thanks for checking!

PR #339, opened 2025-03-13
Automate release

lognaturel:
Closes #

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [x] Not applicable

Nothing! Unfortunately I think we just have to try this one.

I considered automating more but I think this is a good balance of keeping a human in the loop and automating the tedious. We'll still do a PR to generate and verify the changelogs. Then once that gets merged we'll push tags and that will trigger automatic publish. We can add more automation later if we want.

Should have no effect other than incentivizing us to release more!

### DIFF
--- a/README.md
+++ b/README.md
@@ -423,3 +423,4 @@ If you'd like to try the functionality available on `main`, see the preview [on 
 1. Run `yarn changeset version` to generate changelog files and version bumps from the changeset files
 1. Verify that the changelogs look good, commit changes, open a PR, merge the PR
 1. Push tags for each package in the format `package@x.x.x`. A Github action will publish the packages on NPM
+1. Update dependencies to kick off the new release cycle. We do this so that dependency updates get verified implicitly during development.
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 With ODK Web Forms, you can define forms with powerful logic using the spreadsheet-based [XLSForm standard](https://docs.getodk.org/xlsform/). Use [our Vue-based frontend](/packages/web-forms/) or build your own user experience around [the engine](/packages/xforms-engine/)!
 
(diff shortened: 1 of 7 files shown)


packages/scenario/package.json
```diff
@@ -54,5 +54,8 @@
     "solid-js": "^1.9.3",
     "vite": "^5.4.11",
     "vitest": "^2.1.8"
+  },
+  "publishConfig": {
```
latin-panda, line 58: Double checking:

Why do we publish scenario? 
Should we publish common? As it has things that other packages might need - perhaps is already bundled when publishing then please ignore the question

lognaturel: 😬 we don't publish scenario. I've removed that change.

I believe that the Vue client does end up bundling common so that ends up working. But I think currently `xforms-engine` doesn't work on its own. Let's consider that a separate work item.

latin-panda (approved):
LGTM! Let's try it

README.md
```diff
@@ -2,7 +2,7 @@
 
 With ODK Web Forms, you can define forms with powerful logic using the spreadsheet-based [XLSForm standard](https://docs.getodk.org/xlsform/). Use [our Vue-based frontend](/packages/web-forms/) or build your own user experience around [the engine](/packages/xforms-engine/)!
 
-You can try a preview [on the ODK website](https://getodk.org/web-forms-preview/).
+You can try a preview [on the ODK website](https://getodk.org/web-forms-preview/). This gets updated nightly to reflect `main`.
```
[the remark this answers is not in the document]
lognaturel: I think it runs at 11PM GMT? I don't really want to commit in case we change it, though. Maybe you can follow up with more precise/accurate language if it keeps bugging you!

PR #343, opened 2025-03-14
Specify NPM registry to use auth token

lognaturel:
The publish action now runs but fails with an auth error. This attempt at a fix follows https://github.com/npm/cli/issues/6184#issuecomment-1544496444

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,8 +10,13 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout repository
+      - name: Check out repository
         uses: actions/checkout@v4
+
+      - name: Set up npm registry
+        uses: actions/setup-node@v4
+        with:
+          registry-url: "https://registry.npmjs.org"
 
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "private": true,
   "repository": {
     "type": "git",
-    "url": "https://github.com/getodk/web-forms"
+    "url": "git+https://github.com/getodk/web-forms.git"
```
lognaturel, line 6: This was changed by `npm pkg fix`

.github/workflows/release.yml
```diff
@@ -10,9 +10,14 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout repository
+      - name: Check out repository
         uses: actions/checkout@v4
 
+      - name: Set up npm registry
+        uses: actions/setup-node@v4
```
lognaturel, line 17: Note that the action gets the node version from Volta automatically

yanokwa (approved):

Issue #338, opened 2025-03-13
[TASK] Apply code style for "enum" like string constants


latin-panda:
### Description

Implement the agreed approach to "enum" like string constants. Benefits:

- Self-explanatory business rules: The code reads clearly. For example `InstancePayloadStatus.READY` instantly tells you what it’s about.
- Distinct grouping: It’s obvious that `InstancePayloadStatus.READY` is separate from `TransactionStatus.READY` (just example names), avoiding confusion between similar values.
- Fewer mistakes: It reduces typos and enforces a clear scope for these strings, making the code more reliable.

---

### Checklist
- [x] **Does it need UI/UX design?**
  - No
- [x] **Does it need API design?**
  - No
- [x] **Does it need design around state flow?**
  - No
- [x] **Does it need a test plan?**
  - No

---

### Design

- Do a full pass in the project to ensure consistency across  
- Continue to use the type EnumerationName = UNION | OF | STRING | LITERALS, but use a shared prefix for each member of the union
- Use CAPS_SNAKE_CASE across the board
- Hopefully, with a lint rule in tow so, it's more than a convention-by-handshake
- Add documentation about this approach in a code style section

---

### Dependencies
To do after the editing work is merged

- https://github.com/getodk/web-forms/pull/334
- https://github.com/getodk/web-forms/pull/335
- https://github.com/getodk/web-forms/pull/336
- + the other upcoming 1 or 2 PRs related to edit

---

### Related Issues

- [Slack](https://getodk.slack.com/archives/C051L6AKG4F/p1741888706927129)

Issue #66, opened 2024-03-26
Image question type (`<upload>`) without appearances


lognaturel:
Quick links: [Figma](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=1466-5080&t=agV5ZyoLIHkpOQus-0) · [thinking doc](https://docs.google.com/document/d/1PN2tD4-lk0TCSiaYLIy7xkydC1yW4kAIY5RT004IV2s/edit?tab=t.0) · [Collect user-facing docs](https://docs.getodk.org/form-question-types/#default-image-widget) · [XForms spec](https://getodk.github.io/xforms-spec/#media)

### Description

The image question type without appearances allows the user to select an image from their library or, if their device permits, capture a new one.

Out of initial scope:
* default value
* size-based validation (i.e. using max-size prop passed in from host)
* downsizing using the `max-pixels` property
* appearances (like the “new” appearance to only capture a picture)

---

### Checklist
- [x] **Does it need UI/UX design?**
  - If yes, describe the UI/UX requirements or attach a Figma link in the "Design" section.
- [x] **Does it need API design?**
  - If yes, specify the endpoints, methods, or data structures needed.
- [ ] **Does it need design around state flow?**
  - If yes, outline the state flow requirements.
- [x] **Does it need a test plan?**
  - If yes, add the test plan as a comment in this task or attach the document.
---

### User Stories
As a user filling out a form I want to:
* attach an image from my device
* capture a picture from my built-in camera if my device has one
* see an attractive preview of my image that is clear, doesn't take up too much vertical space, and scales as needed if my browser size changes
* replace an existing image
* submit any attached images when I send the full submission

Standard question type considerations:
* read-only: if there's a value, a preview is displayed (out of scope for first pass). Buttons are disabled.
* relevance: full question is hidden if non-relevant
* required: if required, the asterisk and error message should be shown appropriately
* constraint: constraint message is shown in reasonable location
* clear value: x overlaid on image to remove
* default value: OUT OF SCOPE FOR FIRST PASS TODO FILE ISSUE

---

### Design

- **[Figma](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=1466-5080&t=agV5ZyoLIHkpOQus-0)**
[...]

PR #345, opened 2025-03-14
[Edits] Restore serialized 

eyelidlessness:
Part 4/4 in preparation for #247.

Branched from #336 ([diff](https://github.com/getodk/web-forms/compare/edit-epic/decouple-definition-instance-input...edit-epic/restore-serialized-instance-state)).

Opening as draft before my brain melts. Will write up this weekend!

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Not applicable


packages/web-forms/src/demo/InstanceCache.vue
```diff
@@ -0,0 +1,137 @@
...
+	return instance;
+};
+</script>
+
+<template>
+	<section v-if="instantiableFormResult" class="demo-instance-cache">
+		<details>
+			<summary>Instance cache</summary>
+
+			<p>This is an in-memory demo simulating storage and restoration of instance state, which might be similar to the experience of a Web Forms offline mode.</p>
```
latin-panda, line 41: Fix:

Based on this explanation. Since it's a simulation, it feels right to make it clear in the component name: 
`InstanceCacheSimulation.vue`

packages/web-forms/src/lib/shared-state/form-state.ts
```diff
@@ -0,0 +1,117 @@
...
+export type InstantiableFormResult =
+	| LoadFormSuccessResult
+	| LoadFormWarningResult;
+
+export const formResultState: Ref<LoadFormResult | null> = ref(null);
+
+const isInstantiableFormResult = (
+	formResult: LoadFormResult | null
+): formResult is InstantiableFormResult => {
+	return formResult != null && formResult.status !== 'failure';
```
latin-panda, line 25: Fix:

This status ('failure') is safer coming from the xforms-engine, one source for these string constants. For example:

```suggestion
	return formResult != null && formResult.status !== FORM_STATE_STATUS_FAILURE;
```

I know we said about doing it at once in another PR, but if this is a low-hanging fruit, it's worth doing it now.
Since this status is being used outside the engine, I feel extra cautious

packages/xforms-engine/src/client/serialization/InstanceData.ts
```diff
@@ -2,6 +2,11 @@ import type { INSTANCE_FILE_NAME, InstanceFile } from './InstanceFile.ts';
 
 export interface InstanceData extends FormData {
 	get(name: INSTANCE_FILE_NAME): InstanceFile;
+
+	/**
+	 * @todo Can we guarantee (both in static types and at runtime) that
```
latin-panda, line 7: Question:

What do we need to resolve this question?

latin-panda (changes requested):
I need to give another pass on Monday, but submitting some feedback. 

Question: Have you test the performance of the Repeat? Just for sanity :)

packages/common/src/test/fixtures/xform-dsl/PROPOSED_XMLLiteralXFormsElement.ts
```diff
@@ -0,0 +1,37 @@
+import { UpsertableMap } from '../../../lib/collections/UpsertableMap.ts';
+import type { XFormsElement } from './XFormsElement.ts';
+
+const xmlLiteralNameCache = new UpsertableMap<string, string>();
+
+/**
+ * PROPOSED: a convenience wrapper for raw XML implementing the common
+ * {@link XFormsElement} interface.
+ */
+export class PROPOSED_XMLLiteralXFormsElement implements XFormsElement {
```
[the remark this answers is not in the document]
eyelidlessness: This is not seeking feedback _per se_. It's a _proposed addition_ to an interface which we share with JavaRosa. Just as you added to prefix `rankDynamic` in 81a57c3 based on [my feedback](https://github.com/getodk/web-forms/pull/288#discussion_r1936128936) along the same lines. The same pattern is used for additive changes to all of the interfaces we share with JavaRosa for testing.

It's not my favorite pattern in the world. It began as a compromise during the effort to port JavaRosa tests to Web Forms, which was also rushed. It's unclear when (or if?) it will become a priority to synchronize these types of changes between the projects; until it is prioritized, the tradeoff is/has been this use of a single, easily searchable prefix to identify all such cases, rather than to add the overhead of filing an issue or opening a PR in JavaRosa with equivalent changes which may or may not be used.

packages/scenario/src/jr/Scenario.ts
```diff
@@ -1,24 +1,28 @@
 import type { XFormsElement } from '@getodk/common/test/fixtures/xform-dsl/XFormsElement.ts';
```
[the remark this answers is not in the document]
eyelidlessness: Before I address the specific concern, I want to make sure we have a shared understanding of some of the context of what's going on with _this particular file_.

There are some hard-ish constraints on the size of this file which aren't entirely up to us because, like the XForms DSL, `Scenario` is an interface we share with JavaRosa. While the languages are not exactly 1:1, I think we can take the [current state of JavaRosa's equivalent](https://github.com/getodk/javarosa/blob/7986abae78dc0dfe8715fa0071cd04ab698db9bb/src/main/java/org/javarosa/test/Scenario.java) as a decent estimate of a practical floor on its potential line count: 934.

We have already extracted some supporting logic into separate modules, where it's been practical to do (for a variety of reasons, though I'm not sure file size has been a primary motivation). And we've also, by necessity in supporting _the interface as defined in Java_, had to inflate some aspects of this module because: JavaScript _the runtime language_ doesn't have method overloading semantics... which are used a fair bit by JavaRosa's equivalent. (And TypeScript's signature overloading doesn't introduce runtime semantics.)

The additive changes _in this PR_ make up almost the entire delta between our line count and JavaRosa's. Those additive changes are _almost all comments_. Coming to understand those comments' contents, and writing them for posterity and future reference, was **crucial** for me to understand the semantics of _existing_ `Scenario` usage (i.e. ported from JavaRosa) in existing test coverage (again, ported from JavaRosa) affected by this change/feature. Gaining that understanding was also invaluable for me to understand the approach to exercising core semantics of the feature for which I couldn't find existing coverage.

The remaining additive changes in this PR:

- Add a proposed API with clearer semantics and intention _for the feature under test_: 2 SLOC
- Update an existing method (which we skipped in the original porting effort) to call into that added API: 1 assertion (which Prettier expanded to 4 lines so a failure can produce a helpful message) + 1 SLOC
- Improve clarity of a still-unimplemented function: +0 SLOC (Prettier added whitespace)
- Add a few properties to support all of the new test surface area, all of which necessarily goes through `Scenario` or implies a significant rethink on _our approach to engine integration testing_

I'm sure we could **get creative** about extracting some aspects of `Scenario` into other modules, but I want to seriously consider whether the downsides that would entail are worth the goal of a smaller file size _in this case_.

There's a very small amount of code which is _not directly part of the `Scenario` class_. Some of it those are lines which are inherently part of the class's _interface_ and don't belong anywhere else. There are a couple of helper functions we could hypothetically extract to... somewhere. Most of that potential reduction in SLOC would be countered by adding to the already **massive** stack of imports at the top of the module.

And I really think it's worth highlighting that massive stack of imports: quite a lot of `Scenario`, the class and its behavior, _is already being imported from other modules_.

- - -

So with all of that context...

Can you help me understand how you've determined that this file is too big? What is the maximum number of lines you'd like to see in a file, and what is the absolute consequence of exceeding it? How important is this relative to cohesion and comprehension of the module itself? How should we weigh these against potential synchronization overhead (if it should occur) where a module's interface is shared with another project?

- - -

I want to also be explicit that I could have probably answered this with the last two paragraphs in the first section. But I think the context in this case is important. And I think it's just as important to ask that we try to consider context (and seek it out) when we think about introducing new (especially arbitrary) code style rules which don't match the current project state. (An oft-cited name for this principle is [Chesterston's fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence).)

Lastly, I really don't want this to come across as dismissive. If we do want to set a cap on module size, I'm open to that... with exceptions as appropriate, and I think this module is likely to be one such exception. Line count aside, maybe there are other ways we can help make this module more accessible for you. For instance, I've found tools like the TS Language Server incredibly helpful, even _indispensable_, for navigating large modules (and large numbers of small modules too!).

packages/scenario/test/serialization.test.ts
```diff
@@ -252,3 +287,789 @@ describe('ExternalSecondaryInstanceParseTest.java', () => {
...
+
+/**
+ * @todo Maybe we should consider either:
+ *
+ * - A corresponding "deserialization" suite/module
+ * - Renaming this suite/module "serde" (or come up with some more obvious term
+ *   for this concept as it applies to the scope of serializing and
+ *   deserializing instance state)?
+ */
+describe('Restoring serialized instance state', () => {
```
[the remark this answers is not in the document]
eyelidlessness: Yeah, so I included that comment (the one directly above this line) because I agree it's probably best to split out another suite. The intent of the comment is to prompt some discussion about a good structure. Again, this is stuff we at least nominally share with JavaRosa, so I think it's worth putting some thought into that.

Without that consideration, my thinking would go something like:

1. I think it would make sense to continue grouping these concepts (serialization/deserialization) together since they have a mostly inverse relationship ("mostly" because edit will go beyond the direct overlap soon enough). They could be grouped in a subdirectory rather than a single module (_for instance_, but...)
2. A common term for grouping these concepts is the abbreviated portmanteau "serde", but that has all the downsides of abbreviations and portmanteaus.
3. We could also cash in on some of the recent gains we've made clarifying terminology in #334, with a shared prefix rather than a directory, e.g.:
    - `serialization.test.ts` -> `instance-output.test.ts`
    - [New tests] -> `instance-input.test.ts`
    - [And hypothetically soon?] `instance-edit.test.ts`, although I don't expect there to be a ton of test surface area there. Edit functionality will just be restore (or "instance input") + `deprecatedId` + I/O

packages/web-forms/src/demo/InstanceCache.vue
eyelidlessness: I wish I'd had the energy to write up PR notes before EOD yesterday. But since I didn't... apologies for not explaining sooner: 

Unless folks feel strongly otherwise, I was planning to use the last (Vue UI) commit as a temporary internal demo, and drop the commit before we merge this. I included it because I thought it was valuable to be able to show:

- The functionality working _visibly within the product_
- Engine/client API integration in practice, in the UI context
- A little hint of how the functionality overlaps with other pending features (e.g. offline, restoring incomplete instance state, etc)

But I don't think this demo is something we want to land on `main`. If we do want it to be part of this scope, I can spend a few minutes writing up what I think it still needs to be suitable for that. Because there's more to address than the (great!) feedback you've already left here.

packages/web-forms/src/lib/shared-state/form-state.ts
eyelidlessness: I wouldn't be super concerned about addressing this case in this PR if that turns out to be important. However...

- We already merged the [engine side](https://github.com/getodk/web-forms/blob/faa81e48f9a2959a815ee8f1c7e3e22cbec07482/packages/xforms-engine/src/client/form/LoadFormResult.ts#L10-L14) in #335 

- As I mentioned [above](https://github.com/getodk/web-forms/pull/345#discussion_r1997401043), I'm not sure we want to land this demo in its current state anyway

- I'm _much more_ concerned that this suggestion of a constant doesn't reflect the approach we decided to take when we discussed it a couple days ago, nor #338 which captured that decision. I get the sense that we may need to revisit that decision. Which is probably all the more reason not to be hasty about adopting a new approach in this PR.

- I'm **even more** concerned about the impact that continued and growing focus on code style during review is having on productivity, and may have on _functionality_. The former is something I think we've all been feeling in one way or another. The latter is always a concern: in my experience, review focused on code style very often leads to more mistakes in behavior/approach slipping through, sometimes fairly big ones with significant impact. This is a big part of why I advocate for automated tools as a primary solution: I've seen the before/after, and it's pretty stark!

packages/xforms-engine/src/client/serialization/InstanceData.ts
eyelidlessness: A little bit of time and effort. I wrote the comment so I wouldn't forget to apply some of both when it comes up again and when I can afford the cognitive overhead of a context switch.

packages/common/src/test/fixtures/xform-dsl/PROPOSED_XMLLiteralXFormsElement.ts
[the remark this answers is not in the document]
latin-panda: I understand. The "proposed" word is distracting to me; this is how I process it: "proposed" -> "needs agreement" -> "might go away". 

But here, it seems that these "proposes" will likely stay. It'd have been more evident to me to see something like "new extension" or similar. 

For now, this is okay, and I am resolving the comment.

packages/web-forms/src/demo/InstanceCache.vue
latin-panda: This demo seems valuable to me, so it’s fine to remove it from the `main` branch. However, I think it would be helpful to include it in a documentation file as a reference for others. A commit is not easily discoverable, but having it in the documentation would make it a much better experience for others to find and use as an example.

packages/xforms-engine/src/client/serialization/InstanceData.ts
latin-panda: Is that coming in the following PR? Or more like another day sometime in the future? 

My question is in the line to know if we should open a GitHub ticket for this and get it scheduled with the team.

packages/scenario/src/jr/Scenario.ts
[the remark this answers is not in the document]
eyelidlessness: Thank you for elaborating on your concerns. I am totally on board for reducing unnecessary cognitive load. That's a worthy goal. I would like to consider opening an issue about:

- addressing module size as a general factor affecting cognitive load
- how best to balance that general concern in specific cases (such as `Scenario`) where we have collaboration goals outside of this repo

I think we can probably come up with a lot of good ideas on both fronts, but I'm cautious about trying to address them in this PR.

That caution is partly motivated by the sense of urgency we all feel to land edit functionality and unblock Central. Also, since so much of the additive change here is in the form of JSDoc, a big part of my motivation for being thoughtful about addressing it is to try to keep documentation context close to what it's documenting as much as reasonably possible. A lot of the `Scenario` interface carries a lot of semantic weight that isn't necessarily as obvious _where it's used_, and JSDoc provides invaluable leverage for exactly that: editors can/do also make that information available directly where the interface is accessed.

I've found that leverage valuable enough, _especially_ for `Scenario` and similarly semantics-heavy interfaces, that I'd be more inclined to extract **code** in ways that might seem awkward, if it helps keep the relationship between interface/documentation intact.

So my proposal is to file an issue with something like the two bullet points above, and hold addressing this module's size for this PR. Does that sound okay to you for now?

packages/scenario/test/serialization.test.ts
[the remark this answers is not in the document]
eyelidlessness: I feel pretty good about it. I was hoping to be able to do it in a single commit without losing the history of the original file at its new name, but that didn't pan out. So I ended up doing it in three commits to make sure it happened explicitly.

packages/web-forms/src/demo/InstanceCache.vue
eyelidlessness: I agree that it's probably valuable, but I also expect we'll be doing a very similar integration ~~the right way~~ with the intent of supporting Central integration of #247 and #316. While those integration points won't directly exercise `restoreInstance` right away, I think the `editInstance` call site would be an excellent place to link directly to the commit.

If you think that sounds like a reasonable way to keep a reference it, I'd be happy to (er...) commit to making sure that happens as soon as there's an `editInstance` API to call (likely today).

packages/xforms-engine/src/client/serialization/InstanceData.ts
eyelidlessness: I think we'll want to look into it for either #247 or #66.

Is your concern about having TODOs in code comments generally? Because if that's the case, I have some **really bad news**[^1].

Or is this specific TODO of particular concern for you? (Maybe as relates to #66 and/or
#346?) If that's the case, I'd be happy to put some thought into it as soon as you're ready to address it.

Or maybe some other concern I haven't thought of?

[^1]: And it's much less theoretical than the enum concern: we have a **ton** of TODOs in code comments already. For what it's worth, I've been wanting to add linting for them so they can be tracked more intentionally in some way. It just hasn't become a priority yet, since we've had pretty decent TODO comment hygiene thus far.

packages/web-forms/src/demo/InstanceCache.vue
eyelidlessness: Heh. Or even simpler/more immediate: I could see linking to the commit in JSDoc on both the `createInstance` and `restoreInstance` client interface methods in this PR. That way it’s linked right away, without any need to follow up afterwards, right where someone would want to start looking into integrating them.

packages/scenario/src/jr/Scenario.ts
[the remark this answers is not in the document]
latin-panda: > So my proposal is to file an issue with something like the two bullet points above, and hold addressing this module's size for this PR. Does that sound okay to you for now?

Yes, it sounds good to me!

packages/scenario/test/serialization.test.ts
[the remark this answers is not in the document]
latin-panda: That's okay, thank you!

packages/web-forms/src/demo/InstanceCache.vue
latin-panda: > I could see linking to the commit in JSDoc on both the createInstance and restoreInstance client interface methods in this PR. That way it’s linked right away, without any need to follow up afterwards, right where someone would want to start looking into integrating them.

This sounds good to me, thanks!

packages/web-forms/src/lib/shared-state/form-state.ts
latin-panda: Thanks for testing the assumptions. It gives reassurance

 >  think it's a lot more likely that we'll develop other clients before anyone else even wants to

Really looking forward to this!

packages/xforms-engine/src/client/serialization/InstanceData.ts
latin-panda: > Is your concern about having TODOs in code comments generally?

I'm fine with TODO comments. I just wanted to understand what’s needed to address this one—whether I can help now or if we need to create a GitHub ticket and prioritize with the team (project management). But that’s all I had in mind.

latin-panda (approved):
I did another quick pass, and it looks good. I think all the conversations has been discussed. I'm approving now since I won't be able to do it tomorrow, to unblock and you can merge it 

Thanks!

packages/scenario/src/jr/Scenario.ts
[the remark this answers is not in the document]
eyelidlessness: Tracked in #350

PR #349, opened 2025-03-17
[Edits] Engine I/O support for editing submitted instances

eyelidlessness:
🚨 This is not ready for review! 🚨

I'm working toward completing the engine side of #247. I am opening this PR now to **track and show progress**.

[Diff](https://github.com/getodk/web-forms/compare/edit-epic/review/restore-serialized-instance-state-typo-example...edit-epic/edit-instance-interface) from #345

- [x] Engine/client interface: `editInstance`
- [ ] Resource resolution (I/O)
    - [x] Instance XML (I/O)
    - [x] Instance attachments (I/O) **but we might want to roll this part back, and handle separately!**
    - [ ] Test: resource resolution (for at least resolving instance XML)
- [ ] Maybe split into two PRs here
- [ ] `instanceId` -> `deprecatedId`
    - [ ] Test:
        - value assigned
        - implicit element creation when missing
        - namespace behavior consistency with `meta` and `instanceId`
        - not performed for restore
- [ ] Update `instanceId` + `preload="uid"`
    - [ ] Test:
        - after `deprecatedId` value assigned
        - same behaviors as create
        - not performed for restore

PR #352, opened 2025-03-18
[Edits] Support for `deprecatedID` and `instanceID` edit semantics

eyelidlessness:

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [x] Not applicable

I'm not entirely sure how I could validate this in-browser, without either a host app integration or a simulation of one. In any case, I don't think there's anything that should be browser specific here!

Leans heavily on testing (which caught a slightly embarrassing mistake dealing with namespaces!) and to some extent types (which guided most of the `deprecatedID` logic).

Most of this is expanding on prior work (#349 and prior), so there wasn't much new to consider.

There's a big open question about a naming pattern introduced in this change. I'd like us to address it in review.

I can't think of any meaningful regression risks off the top of my head, but I might update this if any come to mind later.

N/A

## What's changed

When `editInstance` is used to initialize instance state:

- If the incoming instance XML has an `instanceID` metadata element, its value is populated in a `deprecatedID` metadata element...
     - ... creating one if it doesn't already exist
     - ... updating its value otherwise
         - ... forming a linked list of sequenced edits
         - ... forming a tree of concurrent edits
     - ... retaining the namespace URI of its `meta` parent and `instanceID` sibling
- If the form defines a `preload="uid"` [implied: for the `instanceID` metadata element], a new value is computed on load...
    - ... but not for "restore" (not an implementation change, newly covered by tests)
    - ... after the incoming `instanceID` value is captured as `deprecatedID` (implicitly covered by the combined set of tests covering the above)

[^1]:

### DIFF
--- a/packages/common/src/constants/regex.ts
+++ b/packages/common/src/constants/regex.ts
@@ -1,0 +1,2 @@
+export const PRELOAD_UID_PATTERN =
+	/^uuid:[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}$/;
--- a/packages/scenario/src/assertion/extensions/submission.ts
+++ b/packages/scenario/src/assertion/extensions/submission.ts
@@ -1,6 +1,12 @@
+import { PRELOAD_UID_PATTERN } from '@getodk/common/constants/regex.ts';
+import {
+	OPENROSA_XFORMS_NAMESPACE_URI,
+	XFORMS_NAMESPACE_URI,
+} from '@getodk/common/constants/xmlns.ts';
 import { assertUnknownArray } from '@getodk/common/lib/type-assertions/assertUnknownArray.ts';
 import { assertUnknownObject } from '@getodk/common/lib/type-assertions/assertUnknownObject.ts';
 import { getBlobText } from '@getodk/common/lib/web-compat/blob.ts';
+import type * as CommonAssertionHelpers from '@getodk/common/test/assertions/helpers.ts';
 import type { DeriveStaticVitestExpectExtension } from '@getodk/common/test/assertions/helpers.ts';
 import {
 	ArbitraryConditionExpectExtension,
@@ -11,6 +17,7 @@ import {
 } from '@getodk/common/test/assertions/helpers.ts';
 import type { SimpleAssertionResult } from '@getodk/common/test/assertions/vitest/shared-extension-types.ts';
 import type { AssertIs } from '@getodk/common/types/assertions/AssertIs.ts';
+import type { ExpandUnion } from '@getodk/common/types/helpers.js';
 import type { InstanceFile, InstancePayload, InstancePayloadType } from '@getodk/xforms-engine';
 import { constants } from '@getodk/xforms-engine';
 import { assert, expect } from 'vitest';
@@ -87,6 +94,135 @@ const getInstanceFile = (payload: AnyInstancePayload): InstanceFile => {
 	return file;
 };
 
+const META_NAMESPACE_URIS = [OPENROSA_XFORMS_NAMESPACE_URI, XFORMS_NAMESPACE_URI] as const;
+
+type MetaNamespaceURI = (typeof META_NAMESPACE_URIS)[number];
+
+type AssertEnumeratedString<T extends string> = (actual: unknown) => asserts actual is T;
+
+/**
+ * @todo This is probably general enough to be exported from {@link CommonAssertionHelpers}
+ */
+const enumeratedStringAssertion = <T extends string>(
+	expected: readonly T[]
+): AssertEnumeratedString<T> => {
+	return (actual) => {
+		assertString(actual);
+
+		expect(expected).toContain(actual);
+	};
+};
+
+const assertMetaNamespaceURI: AssertEnumeratedString<MetaNamespaceURI> =
+	enumeratedStringAssertion(META_NAMESPACE_URIS);
+
+interface SerializedMetaChildValues {
+	readonly instanceID: string | null;
+	readonly deprecatedID: string | null;
+}
+
+type MetaChildLocalName = ExpandUnion<keyof SerializedMetaChildValues>;
+
+interface SerializedMeta extends SerializedMetaChildValues {
+	readonly meta: Element | null;
+}
+
+type MetaElementLocalName = ExpandUnion<keyof SerializedMeta>;
+
+const getMetaElement = (
+	parent: ParentNode | null,
+	namespaceURI: MetaNamespaceURI,
+	localName: MetaElementLocalName
+): Element | null => {
+	if (parent == null) {
+		return null;
+	}
+
+	for (const child of parent.children) {
+		if (child.namespaceURI === namespaceURI && child.localName === localName) {
+			return child;
+		}
+	}
+
+	return null;
+};
+
+const getMetaChildValue = (
+	metaElement: Element | null,
+	namespaceURI: MetaNamespaceURI,
+	localName: MetaChildLocalName
+): string | null => {
+	const element = getMetaElement(metaElement, namespaceURI, localName);
+
+	if (element == null) {
+		return null;
+	}
+
+	expect(element.childElementCount).toBe(0);
+
+	const { textContent } = element;
+
+	assert(typeof textContent === 'string');
+
+	return textContent;
+};
+
+interface MetaNamespaceOptions {
+	readonly [key: string]: unknown;
+	readonly metaNamespaceURI: MetaNamespaceURI;
+}
+
+type AssertMetaNamespaceOptions = (value: unknown) => asserts value is MetaNamespaceOptions;
+
+const assertMetaNamespaceOptions: AssertMetaNamespaceOptions = (value) => {
+	assertUnknownObject(value);
+	assertMetaNamespaceURI(value.metaNamespaceURI);
+};
+
+const getSerializedMeta = (scenario: Scenario, namespaceURI: MetaNamespaceURI): SerializedMeta => {
+	const serializedInstanceBody = scenario.proposed_serializeInstance();
+	/**
+	 * Important: we intentionally omit the default namespace when serializing instance XML. We need to restore it here to reliably traverse nodes when {@link metaNamespaceURI} is {@link XFORMS_NAMESPACE_URI}.
+	 */
+	const instanceXML = `<instance xmlns="${XFORMS_NAMESPACE_URI}">${serializedInstanceBody}</instance>`;
+
+	const parser = new DOMParser();
+	const instanceDocument = parser.parseFromString(instanceXML, 'text/xml');
+	const instanceElement = instanceDocument.documentElement;
+	const instanceRoot = instanceElement.firstElementChild;
+
+	assert(
+		instanceRoot != null,
+		`Failed to find instance root element.\n\nActual serialized XML: ${serializedInstanceBody}\n\nActual instance DOM state: ${instanceElement.outerHTML}`
+	);
+
+	const meta = getMetaElement(instanceRoot, namespaceURI, 'meta');
+	const instanceID = getMetaChildValue(meta, namespaceURI, 'instanceID');
+	const deprecatedID = getMetaChildValue(meta, namespaceURI, 'deprecatedID');
+
+	return {
+		meta,
+		instanceID,
+		deprecatedID,
+	};
+};
+
+const assertPreloadUIDValue = (actual: string | null) => {
+	assert(actual != null, 'Expected preload uid value to be serialized');
+	expect(actual, 'Expected preload uid value to match pattern').toMatch(PRELOAD_UID_PATTERN);
+};
+
+interface EditedMetaOptions extends MetaNamespaceOptions {
+	readonly sourceScenario: Scenario;
+}
+
+type AssertEditedMetaOptions = (value: unknown) => asserts value is EditedMetaOptions;
+
+const assertEditedMetaOptions: AssertEditedMetaOptions = (value) => {
+	assertMetaNamespaceOptions(value);
+	assertScenario(value.sourceScenario);
+};
+
 export const submissionExtensions = extendExpect(expect, {
 	toHaveSerializedSubmissionXML: new AsymmetricTypedExpectExtension(
 		assertScenario,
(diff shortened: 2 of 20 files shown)


packages/common/src/constants/regex.ts
```diff
@@ -0,0 +1,2 @@
+export const PRELOAD_UID_PATTERN =
+	/^uuid:[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}$/;
```
eyelidlessness, line 2: This module is named generally, because I anticipate at least a few other patterns moving here.

packages/scenario/src/assertion/extensions/submission.ts
```diff
@@ -152,6 +288,88 @@ export const submissionExtensions = extendExpect(expect, {
 			return compareSubmissionXML(actualText, expected);
 		}
 	),
+
+	toHaveComputedPreloadInstanceID: new AsymmetricTypedExpectExtension(
+		assertScenario,
+		assertMetaNamespaceOptions,
+		(scenario, options): SimpleAssertionResult => {
```
eyelidlessness, line 295: Hopefully a helpful hint for new concepts...

Each of our assertion extensions follow a similar pattern:

- Treat the assertion's inputs (e.g. arg0 and arg1 corresponding to the call site shape: `expect(arg0).assertionName(arg1)`) as `unknown`
- Parse-validate each input from `unknown` to whatever actual type (runtime and static in tandem) is expected for the actual assertion logic
- Perform assertion logic, returning:
    - `true` -> Vitest considers this a passing assertion, test proceeds
    - `Error` -> Vitest considers this a failed assertion, reports as test failure
- Static types are derived from the parse-validate logic, and used to extend the static types of Vitest itself. This is the best way we've found to ensure `expect(...).toWhatever(...)` type checks correctly (and stays in sync with the custom assertion if/whenever it changes)

packages/xforms-engine/src/instance/children/collectChildInputs.ts
```diff
@@ -0,0 +1,135 @@
...
+ *   interface whose name is consistent with that node
+ * - Update this type to be a union of those interfaces
+ * - Implement that in {@link collectChildInputs}
+ * - Update downstream construction to switch over whatever narrows the union
+ * - Bonus points: revise eeach concrete {@link DescendantNode} to use a common
+ *   constructor API (i.e. a static `from` method, since several such classes
+ *   already have private constructors). Then downstream isn't even a switch
+ *   statement, it's just a lookup table.
+ */
+export interface InstanceNodeChildInput {
```
eyelidlessness, line 90: Naming: everything to do with "child input"s felt right at the start (as a naming convention deriving from "instance input"). In hindsight, this name and many of the others following from it are **confusing**!

I'm not going to delay other aspects of review for this, but I want to make sure to highlight that: I haven't addressed this, I know it's awkward, I want to change it, and I'm open to suggestions!

packages/xforms-engine/src/instance/children/normalizeChildInputs.ts
```diff
@@ -0,0 +1,355 @@
...
+ * on {@link populateDeprecatedID}.
+ */
+const buildMetaValueElement = (
+	parent: MetaParent,
+	localName: string,
+	value: string
+): StaticLeafElement => {
+	const { qualifiedName, nodeset } = parent.definition;
+	const { namespaceURI, prefix } = qualifiedName;
+	const { root } = new StaticDocument({
```
eyelidlessness, line 230: Note: it's a bit weird to new up a `StaticDocument` and immediately discard it in favor of its `root`. The alternatives would be:

- New up a `StaticElement` directly, referencing `parent`... as its `parent`. But this would create a hierarchical relationship which doesn't make sense! The newly created element would not be a child of its `parent`.
- Create a concept in "static-dom" similar to [`DocumentFragment`](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment). This is compelling, but I didn't want to stray into new concepts. If we did ever do this, we'll have to make some choices about how it would interop with `xpath` (because it wouldn't be a valid context node in strict XPath semantics).

packages/xforms-engine/src/instance/children/collectChildInputs.ts
latin-panda: Ok, sounds good, thanks for the highlight

latin-panda (commented):
I couldn't finished it today, there are couple of pieces I need to dig more to fully understand them but that'll be on Monday.

packages/common/src/constants/regex.ts
eyelidlessness: Not related to edit, at least that I'm aware of. I would probably have moved those now if I expected to use them right away.

A few that come to mind:

- [XML/XPath names](https://github.com/getodk/web-forms/blob/cbfe1d3480c2e1a3d3f6785f14d86e67aace7ec6/packages/tree-sitter-xpath/src/grammar.ts#L25-L44)
- [Date/time](https://github.com/getodk/web-forms/blob/cbfe1d3480c2e1a3d3f6785f14d86e67aace7ec6/packages/xpath/src/functions/xforms/datetime.ts#L246-L249) (I think you've linked others at some point as well)

There may be others that aren't bubbling up in my memory, my recall isn't exactly at its best right now!

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
```diff
@@ -9,6 +9,15 @@ const isInstanceFirstLoad = (context: InstanceValueContext) => {
 	return context.rootDocument.initializationMode === 'create';
```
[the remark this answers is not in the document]
eyelidlessness: This is actually an intentional naming dichotomy.

- `isInstanceFirstLoad` is a reference to the ODK XForms specified [`odk-instance-first-load` event](https://getodk.github.io/xforms-spec/#event:odk-instance-first-load). While we haven't yet implemented actions and events, the same event is referenced by [`preload="uid"`](https://getodk.github.io/xforms-spec/#preload-attributes). In the spec, this is a cue to its ordering semantics on form/instance load. In our code, it's a cue to the spec.

    The intent of this name is to express the spec semantics, both as they've been implemented and as the remaining aspects of actions/events will be introduced[^1].

- `isEditInitialLoad` is a special case, which doesn't correspond to any specified event (as its JSDoc suggests). That special case [is `preload="uid"`](https://github.com/getodk/web-forms/blob/4a8cadd186aa6f8fb23eedb229022309a88a8103/packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts#L91-L97). This isn't actually expressed anywhere in the ODK XForms spec (as far as I know), but it's the core assumption for the second half of this PR's semantics (first: `instanceID` -> `deprecatedID`; second: new UUID -> `instanceID`).

    The intent of this name is to express... the absence of `isInstanceFirstLoad`'s intent. In other words, the intent is to specifically call it out as a special case, and disconnect the concept from any implication that it has a basis in spec event naming, or any parity with `isInstanceFirstLoad`[^2].

[^1]: Note: I think it's likely many of these aspects will be moved out of this module when we do implement actions/events. The naming as prep for that is still valuable because it leaves a clear breadcrumb for _ordering of computations_, which is an important detail we'll want to preserve, especially as the code is refactored.

[^2]: Other than their shared `is` prefix: they do have parity in being predicates, but any other parity between them is incidental.

packages/xforms-engine/src/instance/children/normalizeChildInputs.ts
```diff
@@ -0,0 +1,355 @@
...
+interface MetaGroup extends Group {
+	readonly parent: Root;
+	readonly definition: MetaGroupDefinition;
+}
+
+type MetaParent = MetaGroup | MetaSubtree;
+
+const isMetaParent = (parent: GeneralParentNode): parent is MetaParent => {
+	const { nodeType } = parent;
+	switch (parent.nodeType) {
```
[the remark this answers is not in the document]
eyelidlessness: Yes. To be precise, the "meta" naming pattern throughout the module refers to that `<meta>` element subtree, _including its descendants_, as "metadata about an instance" broadly. E.g.

- `<meta>` -> `MetaParent` (more in next paragraph)
- `<meta><instanceID/>` (and `<meta><deprecatedID/>`) -> "meta child"

I guess I should add to my "names I'm not thrilled about" list: `MetaParent`. What I was looking for here was something referring to the `<meta>` element... which is a _subtree_, but at this point in the flow we use the name "subtree" to distinguish from "group". The existing naming is probably a mistake (or at least, probably isn't clear enough), and my reluctance to address that now (out of scope, keeping scope smaller) leads straight to the awkward naming here.

<details>
<summary>Aside(?): "meta" naming convention thoughts</summary>

Another candidate for the `MetaParent` name, probably the most obvious, would be `MetaElement` (or `MetaNode`). Unfortunately, those create ambiguity where the "meta" naming pattern applies also to its descendants.

Since I'm starting to ponder other names this morning, I'll start here: `MetaSubroot` might be more clear.

Some other options:

- Accept that upstream naming is the problem and fix it now:
    - Rename `SubtreeNode` (client interface) and `Subtree` (internal implementation) to `ModelSubtreeNode` and `ModelSubtree` respectively
    - Rename `MetaParent` to its more natural name: `MetaSubtree` (which could be either a "model subtree" or a "group", doesn't matter, it's still a "subtree")
- `MetaParent` -> `MetaElement` + some other naming for descendants
</details>

<details>
<summary>Aside(?): useless switch statement</summary>

I'm not sure if the confusion here _began_ with this useless `switch` statement that I forgot to delete. But that sure confused me seeing it with fresh eyes! I'm removing it now. In the future, I wouldn't be surprised if [`eslint-plugin-unicorn/no-useless-switch-case`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-useless-switch-case.md) could have caught this.

I don't know if there are other (erm) cases like this in the project. Seems unlikely, but there might be! I often try out a `switch` first, before starting to write an `if` statement like the one below (like the one below == "I know it will need to narrow cases of a union type" + "has a set of additional constraints on some narrowed subset of that union"). I just usually remember to back out of that when it doesn't seem like it'll add clarity. This is definitely an artifact of my brain being exhausted!</details>

packages/scenario/test/instance-edit.test.ts
```diff
@@ -0,0 +1,383 @@
...
+				const leaf2 = await branch2.proposed_editCurrentInstanceState();
+
+				assertChained(source, branch1);
+				assertChained(source, branch2);
+				assertChained(branch1, leaf1);
+				assertChained(branch2, leaf2);
+
+				assertNotChained(branch1, leaf2);
+				assertNotChained(branch2, leaf1);
+			});
```
eyelidlessness, line 338: When I first wrote up this PR, I noticed I had forgotten to test that we don't append multiple `deprecatedID` elements. As I went to test that, I realized that what was really lacking was this pair of spec semantics:

- sequenced edits produces a linked list
- concurrent edits produces a tree

I'm pretty sure this was discussed in a team meeting, around the time we started queueing up edit feature work. I hope these two tests clarify those cases!

packages/xforms-engine/src/instance/children/collectChildInputs.ts
eyelidlessness: Yeah, this definitely feels like the right way to go. I've pushed a commit which looks more or less as I described above. A minor difference is that I decided to use `DescendantNodeInitOptions` as a (temporary) name, to disambiguate from a current interface already named `DescendantNodeOptions` (local to `DescendantNode.ts`).

After I made the changes I took a quick sample of the current node constructor signatures, and I wasn't surprised to see that the object looks almost exactly like what those nodes already take, only they're currently positional. The main exception is repeats, which take N `instanceNodes` rather than one optional `instanceNode`. It wouldn't be too big a stretch to unify the signatures on a shared options parameter now, while keeping the other improvements for a later scope.

But I'm already much happier with the naming changes, and some other improvements to code clarity along the way. So I'd also be happy to leave it pretty much as it stands now, if that's where we land in the next review pass!

packages/common/src/constants/regex.ts
latin-panda: FYI note: I’ve extracted the regex for [date time](https://github.com/getodk/web-forms/pull/328) into its own file, since it’s substantial enough to stand alone rather than being combined with other regex patterns in a single file

packages/xforms-engine/src/instance/children/normalizeChildInitOptions.ts
```diff
@@ -0,0 +1,329 @@
...
+ * {@link XFormDOM}, this is a good place to start for logic equivalent to the
+ * hacky normalization for `instanceID`.
+ */
+const populateDeprecatedID = (
+	subroot: EditModeMetaSubrootInitOptions
+): EditModeMetaSubrootInitOptions => {
+	const value = getInstanceIDValue(subroot);
+
+	if (value == null) {
+		return subroot;
```
latin-panda, line 293: Suggestion: 

Handle the rare case where `deprecatedID` exists but `instanceID` is missing by copying `deprecatedID`’s value to `instanceID`. ([Related Slack convo](https://getodk.slack.com/archives/C051L6AKG4F/p1742843180340489?thread_ts=1742842551.382939&cid=C051L6AKG4F))

I couldn’t find this handled elsewhere in the codebase (quick search—might’ve missed it?). It’s edge-case-y and could be deferred, but since we’re already in `normalizeChildInitOptions.ts` tweaking `meta` logic, it’s low-hanging fruit.

eyelidlessness: > Handle the rare case where deprecatedID exists but instanceID is missing by copying deprecatedID’s value to instanceID.

@lognaturel can correct me if I'm wrong, but I think this behavior would be incorrect. The value of `instanceID` is used to populate `deprecatedID`, but not vice versa. The behavior as it is now does account for missing `instanceID`: it does not populate a `deprecatedID` with a value, because it has no value to populate it with.

eyelidlessness: To clarify, the behavior as it exists:

1. If there's no `instanceID` _in the form_, one is created and a `bind preload="uid"` created for it. This happens in `XFormDOM`, which is where form definition normalization is currently handled.
2. If there's no `instanceID` _in the instance_, one will be created from the form definition (regardless of how it's defined by the form or not) as part of resolving missing instance nodes _generally_. This was introduced in #345, as part of how we resolve omission of non-relevant nodes. Its value will not be transferred to `deprecatedID` here, because it does not have a value to transfer. If the form defines a `preload="uid"` for the node (or if we normalize one into existence per point 1), a new `instanceID` will be computed after this point. That is ordering is intended per spec (and tested as such).

lognaturel: I agree that the existing behavior seems like our best bet. That said, I think it's a weird enough case that it doesn't need to have a test or a guaranteed behavior.

latin-panda (approved):
Ready to merge!!

latin-panda (approved):
Ready!

PR #354, opened 2025-03-24
feat(#295): adds overlay for rank initial state

### LINKED ISSUE #295: Differentiate between a user skipping the rank question and consciously choosing the default order
_Note: This is different from [issue #294](https://github.com/getodk/web-forms/issues/294). This issue is about the question’s value (user input) held internally in the form._

**Description**
The [current rank implementation](https://github.com/getodk/web-forms/pull/288) is incomplete because there is no way to differentiate between a user skipping the question and consciously accepting the default order.

**Solution**
- If a rank question is required, and the user has not reordered the options from their default order, there should be an empty state with an overlay that requires the user to interact with it. If they don't interact, it will be treated like a missed question. 
- The overlay is initially displayed regardless of whether it's a required question type or not. 

Image

[Figma](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=1716-19784&t=0VPQJwIRMQGpP3LL-1)

**Additional Details**
Both Collect and Enketo start with an empty answer and only populate the response when the user takes explicit action.
In line with this behavior, Enketo allows clearing a rank question’s value, resetting it to the initial disabled state.

**To do:**
- [x] Research and determine priority
- [x] Design a consistent user interface and experience for this empty state.
- [ ] Automate tests to verify the UIUX

latin-panda:
Closes #295

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [ ] Safari (iOS) - I don't have a device, but it should work fine
- [ ] Not applicable

Testing evidence :

<details><summary> Initial state</summary> 

Mobile
Desktop

</details>

<details><summary> Required state</summary> 

Error and enabled state:

Desktop

Error and disabled state:

Desktop

</details>

<details><summary> Disabled state</summary> 

Mobile
Desktop

</details>

<details><summary>Video: User experience</summary> 

In the first half of the video, the rank question is not required for the form. Submitting the form without tapping the "Rank items" button leaves the question unanswered, and the form is submitted.

In the second half of the video, the rank question is required. Submitting the form without ranking will highlight the rank question in red and prevent submission. The user can then tap "Rank items", which sets the default order of items as the value for the question, and then the form can be submitted. 

https://github.com/user-attachments/assets/bef9a2a7-0a0e-4881-a071-1c1fc1300108

</details>

The [vue-draggable-plus](https://vue-draggable-plus.pages.dev/) library doesn't have this feature. So, building the overlay that reuses the existing button styling and logic to set rank's options. 

- Adds an empty state with an overlay that requires the user to interact with it. If they don't interact, it will be treated like a missed question.

No

## What's changed

- It adds an overlay with a button when the user hasn't interacted with the question type (required or not required question). 
- It updates the button's disabled shades to match geopoint and image upload question types
- Prevents icons from shrinking when text is too long


alyblenkin:
@latin-panda I think the interaction is very clear and looks great! 

The only thing I'm not completely happy with on my end is the colour contrast between the error state and empty state. I contemplated the white opacity instead of the grey for the empty state, but was concerned it might still look interactive. This could be something we tweak later on, but here is a mockup of what I'm thinking! 

Screenshot 2025-03-25 at 7 43 52 AM

alyblenkin:
We discussed as a team that we should make the overlay darker for the disabled state when there is an error (likely pretty rare, but we want to ensure the contrast is high enough). The white opacity isn't strong enough in the examples above and still looks interactive. 


Screenshot 2025-03-25 at 12 22 18 PM

packages/web-forms/src/components/controls/RankControl.vue
```diff
@@ -101,58 +107,70 @@ const swapItems = (index: number, newPosition: number) => {
 <template>
 	<ControlText :question="question" />
 
-	<VueDraggable
```
latin-panda, line 104: Anything inside the `VueDraggable` element is the same as before; I just moved it inside a `rank-control-container`. The structure is:

|- rank-control-container
   |- rank overlay 
   |- VueDraggable

The rank overlay isn't inside `VueDraggable` because it's not good that it inherits "draggable" styles or events when it's not a draggable element.

packages/web-forms/src/components/controls/RankControl.vue
```diff
@@ -43,14 +43,20 @@ const values = computed<string[]>({
...
-		return props.question.currentState.valueOptions.map((option) => option.value);
+		return getRankItems();
 	},
 	set: (orderedValues) => {
 		touched.value = true;
 		props.question.setValues(orderedValues);
 	},
 });
 
+const getRankItems = () => props.question.currentState.valueOptions.map((option) => option.value);
```
latin-panda, line 54: The `props.question.currentState.valueOptions` contains the original order of the rank's items. The order doesn't get modified. 

The ordered rank's items are assigned to `value` (`props.question.currentState.value`).

When the user taps on the `Rank items` button from the overlay, it takes the rank's items from `props.question.currentState.valueOptions` (original order) and sets it to `value`

packages/web-forms/src/components/controls/RankControl.vue
```diff
@@ -164,18 +182,25 @@ const swapItems = (index: number, newPosition: number) => {
 @import 'primeflex/core/_variables.scss';
 
 // Variable definition to root element
-.rank-control {
+.range-control-container {
```
latin-panda, line 185: This section defines CSS variables specific to this component, helping us identify patterns across question types that can later be extracted at a higher level for reuse. We'll get to this when adjusting the styles of Web Forms once PrimeVue is upgraded.

packages/web-forms/src/components/controls/RankControl.vue
```diff
@@ -205,6 +231,10 @@ const swapItems = (index: number, newPosition: number) => {
 		align-items: center;
 		gap: var(--rankSpacing);
 	}
+
+	.rank-label svg {
+		flex-shrink: 0;
```
latin-panda, line 236: Prevents icons from shrinking when text is long and needs to wrap to the next line

lognaturel (approved):
Thanks for the notes!

PR #355, opened 2025-03-24
Initial support for editing submissions, resetting form state after submission

eyelidlessness:
All of #247 except UI/UX revisions!

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [x] Not applicable

I have verified this with tests. We'd need to simulate a host app integration to verify it in browsers. I still don't think anything here would be browser specific.

Testing:

- A quick validation that loading a form with `editInstance` input results in a form with that instance's state
- Much more validation that the solution for creating multiple instances doesn't introduce regressions (i.e. "everything is optional")

Some of this involved cleaning up the states possible in `OdkWebForm` (currently the primary host app interface). It wasn't strictly necessary, but the necessary features required additive changes to possible state which I found **really hard** to reason about without making it more explicit.

As best as I could, I exercised the regression risks I know about in tests.

N/A

## What's changed

**Instance editing:** exposes a host app interface which is mostly a passthrough to the engine's interface of same. One difference, based on feedback from the Central team, is a slightly different representation of instance attachments.

**Creating multiple instances:** expands the emit signatures for `submit` and `submitChunked` to pass a callback, which a host app can invoke to signal:

1. That it is handling the submission (`Promise` == in progress, may also be synchronous, or may be ignored; **everything about this is optional**)
2. That it has finished handling submission (`Promise` resolution == success or failure)
3. That it wants Web Forms to create a new instance (by passing us a sentinel value which we export)

If we receive that sentinel value, we create a new instance. If any other thing happens, the current behavior is unchanged: pass an `InstancePayload` to host app, do nothing afterward.

### DIFF
--- a/packages/web-forms/src/lib/init/updateSubmittedFormState.ts
+++ b/packages/web-forms/src/lib/init/updateSubmittedFormState.ts
@@ -1,0 +1,33 @@
+import { POST_SUBMIT__NEW_INSTANCE } from '../constants/control-flow.ts';
+import type { OptionalHostSubmissionResult } from '../submission/HostSubmissionResultCallback.ts';
+import { ENGINE_FORM_INSTANCE_CONFIG } from './engine-config.ts';
+import type { FormStateSuccessResult } from './FormState.ts';
+
+/**
+ * @todo Clean up {@link currentState}'s {@link FormStateSuccessResult.instance}
+ * before creating a new one. (Requires engine support, but will need to be
+ * invoked here!)
+ */
+const resetInstanceState = (currentState: FormStateSuccessResult): FormStateSuccessResult => {
+	const { form } = currentState;
+	const instance = form.createInstance(ENGINE_FORM_INSTANCE_CONFIG);
+
+	return {
+		status: 'FORM_STATE_SUCCESS',
+		error: null,
+		form,
+		instance,
+		root: instance.root,
+	};
+};
+
+export const updateSubmittedFormState = (
+	submissionResult: OptionalHostSubmissionResult,
+	currentState: FormStateSuccessResult
+): FormStateSuccessResult => {
+	if (submissionResult?.next === POST_SUBMIT__NEW_INSTANCE) {
+		return resetInstanceState(currentState);
+	}
+
+	return currentState;
+};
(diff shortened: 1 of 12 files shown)


packages/web-forms/src/lib/init/updateSubmittedFormState.ts
```diff
@@ -0,0 +1,33 @@
+import { POST_SUBMIT__NEW_INSTANCE } from '../constants/control-flow.ts';
+import type { OptionalHostSubmissionResult } from '../submission/HostSubmissionResultCallback.ts';
+import { ENGINE_FORM_INSTANCE_CONFIG } from './engine-config.ts';
+import type { FormStateSuccessResult } from './FormState.ts';
+
+/**
+ * @todo Clean up {@link currentState}'s {@link FormStateSuccessResult.instance}
+ * before creating a new one. (Requires engine support, but will need to be
+ * invoked here!)
+ */
```
eyelidlessness, line 10: Tracked in #356

packages/web-forms/tests/components/OdkWebForm.test.ts
```diff
@@ -151,4 +173,274 @@ describe('OdkWebForm', () => {
...
+			},
+
+			// EVERYTHING is optional. This case ensures that introducing the callback
+			// as a second parameter doesn't introduce regressions in a host
+			// integration which only uses the first parameter.
+			{
+				description: 'does not reset form state by default (no callback)',
+				hostSubmissonHandler: null,
+				expectedPostSubmissionValue: firstSubmissionInputValue,
+			},
```
eyelidlessness, line 408: One thing not tested here, and not currently handled: we should probably expire the callback. I did a very quick spike on having it expire after the current event loop tick (in other words: a host can call the callback immediately with a `Promise`, but the side effect isn't triggered if a host `await`s a `Promise` and then calls the callback after resolution). This is probably solvable, but I wasn't able to do it as trivially as I thought I could. JS event loop scheduling is a lot of things, but it's definitely not _trivial_ (especially a couple weeks into mental fatigue)! 🙃

latin-panda (approved):
Ready to merge! 🚀

PR #359, opened 2025-03-26
chore: updates dependencies

latin-panda:

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [ ] Safari (iOS) - I don't have device
- [ ] Not applicable

Manual smoke test and CI checks passed successfully 

N/A

N/A

I don't think so 

## What's changed

- Uplift of dependencies with minor or patch updates. 
- Small lint fixes to accommodate the newer version of eslint. 

Notes:
- Playwright's newer versions (>v1.50.0) make the CI stuck when running the tests on Firefox. This doesn't happen when running the tests (headless or not) locally. This is related to the Firefox version in Playwright and how the test runner closes/terminates the browser after a test and starts it again for a new test case, which is an issue on the Playwright or Vitest. We can upgrade the  Playwright and Vitest at the same time and see if the problem persists.


latin-panda:
Hi @lognaturel, this is a tiny PR that is ready for review. It uplifts all the easy dependencies.

I will start the uplift of these other ones (screenshot below) in a separate PR since I think there will be some fixes for breaking changes. 

Screenshot 2025-03-26 at 2 30 14 PM

Thank you for your time!

lognaturel (approved):
No changeset for this one because they're mostly updates to build tools or very unlikely to have user-facing consequences, is that right? But then next round might have a changeset?

latin-panda:
@lognaturel good catch, I've added the changeset

PR #369, opened 2025-04-10
Skip CI if-and-only-if run from a non-fork PR

eyelidlessness:
Fixes failure to produce a build artifact on merge to `main`. (Or at least, I think it fixes that!)

Closes # (do we have an issue for this?)

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [x] Not applicable
- [x] GitHub Actions

After the first failed attempt in #368, I did a bunch of trial and error in a private personal repo. Note: I haven't validated the fork-specific logic (because no fork of private personal repo)! But I have validated that jobs predicated on that condition run as expected (and don't run for a negative predicate result).

It probably isn't the best possible solution!

- I briefly considered investigating other CI solutions but we don't want that.
- We discussed making the build -> upload artifact job more explicit. I think we should do that but don't want to block a pragmatic fix over it.

I could imagine a few hundred ways this breaks assumptions we have about CI. But the real risk is the ways I can't imagine. Such is the nature of GitHub Actions!

N/A

## What's changed

CI will now be skipped if triggered by a `pull_request` event, except if the PR is from a fork... where "from a fork" is defined as `repository_owner` being some value other than `'getodk'`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,31 +24,21 @@ env:
   LOCALE_ID: 'en-US'
 
 jobs:
-  check-run:
+  ci-conditions:
+    name: 'CI conditions'
     runs-on: 'ubuntu-latest'
     outputs:
-      should-skip: ${{ steps.check.outputs.should-skip }}
+      IS_FORK: ${{ github.repository_owner != 'getodk' }}
+      IS_PR: ${{ github.event_name == 'pull_request' }}
+
     steps:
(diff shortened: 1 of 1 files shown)


latin-panda (approved):
It ends up like:
Non-forked runs `push`
Forked runs `pull_request` and `push`

LGTM! Ready to merge

PR #373, opened 2025-04-15
(WIP) Engine/client API design: representation of RichText, itext media, outputs in itext, [limited but thoughtful] groundwork for Markdown

eyelidlessness:
Opening as draft for visibility before team meeting.

<!--

Closes #

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Not applicable

### DIFF
--- a/packages/xforms-engine/src/client/media/MediaResource.ts
+++ b/packages/xforms-engine/src/client/media/MediaResource.ts
@@ -1,0 +1,23 @@
+import type { AudioMediaResource } from './AudioMediaResource.ts';
+import type { MediaResourceType } from './BaseMediaResource.ts';
+import type { ImageMediaResource } from './ImageMediaResource.ts';
+import type { UnknownMediaResource } from './UnknownMediaResource.ts';
+import type { VideoMediaResource } from './VideoMediaResource.ts';
+
+interface MediaResourceByTypeSuffix {
+	readonly AUDIO: AudioMediaResource;
+	readonly IMAGE: ImageMediaResource;
+	readonly VIDEO: VideoMediaResource;
+	readonly UNKNOWN: UnknownMediaResource;
+}
+
+// prettier-ignore
+type MediaResourceTypeSuffix<ResourceType extends MediaResourceType> =
+	ResourceType extends `MEDIA_RESOURCE_TYPE_${infer Suffix}`
+		? Suffix
+		: never;
+
+// prettier-ignore
+export type MediaResource<
+	ResourceType extends MediaResourceType = MediaResourceType,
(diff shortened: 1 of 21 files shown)


packages/xforms-engine/src/client/media/MediaResource.ts
```diff
@@ -0,0 +1,17 @@
...
+	readonly MEDIA_RESOURCE_TYPE_AUDIO: AudioMediaResource;
+	readonly MEDIA_RESOURCE_TYPE_IMAGE: ImageMediaResource;
+	readonly MEDIA_RESOURCE_TYPE_UNKNOWN: UnknownMediaResource;
+	readonly MEDIA_RESOURCE_TYPE_VIDEO: VideoMediaResource;
+}
+
+// prettier-ignore
+export type MediaResource<
+	ResourceType extends MediaResourceType = MediaResourceType,
+> = MediaResourceByType[ResourceType];
```
eyelidlessness, line 17: Not really specific to this design proposal, but I want to call out that I learned something new about TypeScript while working on designing all of the interfaces here.

**THIS** is apparently a technique for unifying:

- Discriminated unions
- Generics
- Generics with a default which is the same discriminated union

It could help us simplify a **lot of types** (hint: most, maybe all, cases where we've defined a generic type like `Foo` and a corresponding union type like `AnyFoo`).

packages/xforms-engine/src/client/text/RichTextUnit.ts
```diff
@@ -0,0 +1,112 @@
...
+	 *   `<output>`, inline text), and may be recomputed more frequently than its
+	 *   {@link children | structured counterpart}.
+	 *
+	 *
+	 * @todo [Design review detail, remove before implementation.]
+	 *
+	 * Do we actually want to expose this? The most likely use case would be
+	 * testing. The risk of exposing it is that it could encourage conflicting or
+	 * divergent implementations of ODK's flavor of Markdown (diverging in terms
+	 * of parsing, semantics, capabilities, and so on).
```
eyelidlessness, line 109: _Actually_, if we lean more toward an approach where **clients determine** which `RichText` cases have what capabilities, access to `rawText` may be important for that! For example, it would allow a client receiving the same `RichText` values for `label` and `hint` to render:

- `<label>**foo** bar</label>` -> `<label><strong>foo</strong bar</label>`
- `<hint>**foo** bar</hint>` -> `<p>**foo** bar</p>`

With the same parsing logic, and without access to `rawText`, those different renderings would probably not be possible.

PR #377, opened 2025-04-17
Optimization: `@getodk/xpath` is faster without `Iterable`/`Iterator`

eyelidlessness:
Closes # (N/A)

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [x] Safari (iOS)
- [ ] Not applicable

As a refactor with no change to functionality, this leans heavily on type check and existing tests.

It's kind of a shame, actually! `Iterator`s align really well with XPath semantics. Unfortunately, they have some drawbacks. One significant drawback is that they're poorly optimized (and to some extent that may be inherent to their semantics).

It's possible some portion of this refactor is incorrect in a way that isn't exercised by tests. That seems pretty unlikely to me, given my familiarity with the coverage and the areas affected. But it's always a possibility with any refactor.

The fixtures which I used most to evaluate performance improvements:

- `bench9-alt.xml`
- `nigeria_wards_external.xml`

The slowest in testing, which also show significant improvements:

- `child_vaccination_VOL_tool_v12-alt.xml` (and this improvement may reduce, maybe eliminate, recent CI timeouts)
- `whova_form.xml`

## What's changed

Everything[^1] in `@getodk/xpath` which uses `Iterator`s and/or has an `Iterable` interface has been refactored to use more eager equivalents. This also includes use of generator functions, which have been refactored to regular functions.

In many cases the eager equivalent chosen is `Set`, which has semantics consistent with XPath `node-set`, which in turn is where the XPath implementation has generally used `Iterable`/`Iterator` interfaces and semantics.

Perhaps counterintuitively, eager turns out to be much faster than lazy, pretty much[^2] across the board! This is noticeable on slow forms across all browsers, but may be especially pronounced in Firefox (which has more pronounced perf penalties for `Iterable`/`Iterator` APIs, and especially generator functions).

I'd be happy to include specific before/after benchmarks if requested.

[^1]: Except where these interfaces are essential.
[^2]: There are probably some very specific cases where it might be slower-in-principle, e.g. deep LocationPath expressions where a single node is targeted and many ancestors have many matching nodes. It would be interesting to fabricate those cases for benchmarking. My hunch is even they may perform faster, due entirely to the nature of the APIs in use.


latin-panda (approved):
This is great! Thank you so much! 
Merging now

PR #385, opened 2025-04-28
chore: fixes lint warnings in error messages

latin-panda:

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [x] Not applicable

N/A

N/A

N/A

N/A

### What's changed

- Adds description to error messages 
- Throws Error object instead of string

### DIFF
--- a/.changeset/legal-chairs-fold.md
+++ b/.changeset/legal-chairs-fold.md
@@ -1,0 +1,6 @@
+---
+'@getodk/scenario': patch
+'@getodk/xpath': patch
+---
+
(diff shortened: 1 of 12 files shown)


lognaturel (approved):
We eventually want to rethink errors across the board and come up with a way to manage localized messages and have more predictable control flow. This change makes for more useful error messages for users in the mean time, silences lint, and means we can search for `new Error` when considering a blanket change.

PR #394, opened 2025-05-02
Flatten TextChunkExpression

lognaturel:
A `TextChunkExpression` represents either some literal text or an expression to get text for a label, hint, constraint message.... This PR goes from using subclasses to represent expressions from different sources to a single concrete class with multiple static factory methods. The subclasses weren't used polymorphically. It also removes several subtypes that were only used internally or weren't used.

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [x] Not applicable

I mostly used my understanding of the intent to guide this and I feel confident I can explain everything the new concrete class does. 

I think it's a pragmatic change that reveals the intended concepts. It's possible I cut concepts we'll want to bring back later but this gives us an understandable starting point to build on.

This is really well covered by scenario tests so I think the risk is low. The biggest risk I can think of is that I may have lost some subtle behavior when going from the constructors to the static factories. In particular, there was a difference in how translation expressions were validated for messages vs. labels/hints that felt accidental to me but that could have been meaningful in some way.

No.


latin-panda (approved):
Sparks ✨

Thank you!

PR #399, opened 2025-05-09
feat(#347): adds support for select with images - engine part

### LINKED ISSUE #347: Add selects with images
### Description

Add selects with images

---

### Checklist
- [x] **Does it need UI/UX design?**
  - If yes, describe the UI/UX requirements or attach a Figma link in the "Design" section.
- [x] **Does it need API design?**
  - If yes, specify the endpoints, methods, or data structures needed.
- [ ] **Does it need design around state flow?**
  - If yes, outline the state flow requirements.
- [x] **Does it need a test plan?**
  - If yes, add the test plan as a comment in this task or attach the document.
---

### User Story 

- **As a** form designer,
  **I want** add selects with images with different appearances  
  **so that** so that data collectors can have a visual reference when selecting choices.

### Appearances for select one/multiple

- **Use columns appearance when no appearance is specified** 
  - display the options in columns depending on the size of the images and device 
  - fixing the height to 300px and width flexible (same as the image question)
  - set breakpoints for different number of columns (based on rules from the text selects)
  - columns all have the same width (determines the max width of each image)
  - if the image is smaller than 300x300px, we add a grey background 
[Figma link](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=1774-7835&t=i9w4aZyF1IVTQaDF-1)

- **Grid: columns appearance**
Same as the new default for no appearance above
[Figma link](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=1774-7835&t=i9w4aZyF1IVTQaDF-1)

- **Grid: columns-pack appearance**
Try to accommodate as many options as possible in a single line as possible per device and scales down accordingly per screen size. For users who want to conserve space to prevent the user from scrolling to see more options. Not ideal for accessibility. 
[Figma link](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=1842-9808&t=i9w4aZyF1IVTQaDF-1) 

- **Grid: [columns-n appearance](https://docs.getodk.org/form-question-types/#select-widget-with-columns-n-appearance)**
Users can decide the number of columns and enforce that number (2, 3, 3+), but it caps at 10.
[Figma link](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=1774-7547&t=i9w4aZyF1IVTQaDF-1)

- **Images, no buttons** 
No formal button in the label area, but the option is highlighted with a tick in the corner. The user can tap anywhere on the image. 
[Figma link](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=1774-7833&t=2mqnH3FESYUxfFX4-1)

- **Long label**
We don't want to truncate information so the box will keep expanding. Other options should also expand so everything in that row should be the same height.
[Figma link](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=1776-10205&t=i9w4aZyF1IVTQaDF-1)

- **Disabled state**
Images should still be partially visible with 60% opacity for the images with disabled text styles.
[Figma link](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=1794-11765&t=Gk1uI68qwB5gjO69-1)

- **Loading images**
When images take longer to load, the user will see a visual placeholder of the content, so they know something is happening. The label will still be visible.
[Figma link](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=2335-2487&t=teJgorYh3L4aO0zh-1)

- **Broken image**
If the image is unable to load, the user sees a image icon with a slash through it.
[Figma link](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=1881-7559&t=teJgorYh3L4aO0zh-1)

---

### Dependencies

- [dependency]
- [dependency]

---

### Additional Notes
- We discussed lots of different options and the [tradeoffs of each here](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=1774-7836&t=i9w4aZyF1IVTQaDF-1), final designs above.
- We want to give users tools to design more accessible and intuitive forms (just because it's possible to layer on lots of different appearances doesn't mean it's a good user experience). That could be a simple visual guide with do, don't, and caution in the docs.
- We also spoke about giving folks an icon library to make their forms more visual. 
- Documented the [experiments and assumptions here](https://docs.google.com/spreadsheets/d/1nD9VJeQriQ4mpSUN6N4zV0625-6GCkp9TM5dcMbYauk/edit?gid=0#gid=0) (e.g., use columns for the default) 

### Related Issues

- [issue link]
- [epic issue link, if applicable]

latin-panda:
Closes #347

## WIP - pending:
- Resolve ToDo
- Polish code
- Add more unit tests for Primary Instance class 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable


packages/xpath/test/xforms/jr-itext.test.ts
```diff
@@ -240,17 +240,38 @@ describe('jr:itext(id)', () => {
 			{ expression: 'jr:itext("three")', language: 'en', expected: 'Three' },
 			{ expression: 'jr:itext("one")', language: 'fr', expected: 'Un' },
 			{ expression: 'jr:itext("two")', language: 'fr', expected: 'Deux' },
-			{ expression: 'jr:itext("three")', language: 'fr', expected: 'Trois' },
+			{ expression: 'jr:itext("three")', language: 'fr', expected: 'Ignored' },
```
latin-panda, line 243: When requesting a `string` value and there are many `value` nodes, XPath returns the text of the first node; in this case, the text is `Ignored`. 

I'm not sure if the expectation is: 
1. Return the text of the first node; then, this is correct.
2. Return the node's text without the `form` attribute; then, we must make some changes to validate each node.

lognaturel (commented):
I've only scrubbed through quickly but here are a few things that caught my eye!

packages/xpath/test/xforms/jr-itext.test.ts
lognaturel: Eventually it should return the value without a `form` attribute. I think we could file an issue for that.

latin-panda: > it should return the value without a form attribute

This means returning the "default" value of an itext. 

```xml
<itext>
  <translation>
    <value form="<short|image|etc..>"> this is an alternative text/media</value>
    <value> this is the default text </value>
 </translation>
</itext>
```

The expectation in this case is to get: `this is the default text`

packages/xpath/src/xforms/XFormsItextTranslations.ts
```diff
@@ -156,51 +162,43 @@ export class XFormsItextTranslations<T extends XPathNode> implements XFormsItext
...
 		if (textElement == null) {
-			return '';
-		}
-
-		const { domProvider } = this;
-
-		for (const valueElement of domProvider.getChildrenByLocalName(textElement, 'value')) {
-			//
-			if (!domProvider.hasLocalNamedAttribute(valueElement, 'form')) {
-				return domProvider.getNodeValue(valueElement);
```
latin-panda, line 199: Fact: This is where the actual XPath query happens; it returned the first `value` nodes that didn't have the `form` attribute. 
Now, it returns all value nodes, and they are evaluated in the engine (see `createTextRange` file)

packages/xpath/test/xforms/jr-itext.test.ts
lognaturel: Exactly, that's the intent for the user-facing `jr:itext` function. But as we've discussed, it's quite rare to use it in form design and if someone does use it in form design they generally won't specify alternative forms.

packages/xpath/src/functions/javarosa/index.ts
```diff
@@ -1,5 +1,5 @@
 import { JAVAROSA_NAMESPACE_URI } from '@getodk/common/constants/xmlns.ts';
 import { FunctionLibrary } from '../../evaluator/functions/FunctionLibrary';
-import * as string from './string.ts';
+import * as nodeSet from './node-set.ts';
 
-export const jr = new FunctionLibrary(JAVAROSA_NAMESPACE_URI, [...Object.values(string)]);
+export const jr = new FunctionLibrary(JAVAROSA_NAMESPACE_URI, [...Object.values(nodeSet)]);
```
latin-panda, line 5: Fact: Currently, there can only be one function for `itext`. I tried leaving string and node-set, but it just keeps one. Knowing that `itext` will always return one data type from the XPath layer could be a good thing.

packages/xforms-engine/src/parse/model/SecondaryInstance/sources/ExternalSecondaryInstanceResource.ts
```diff
@@ -1,19 +1,13 @@
...
 import { ErrorProductionDesignPendingError } from '../../../../error/ErrorProductionDesignPendingError.ts';
 import { getResponseContentType } from '../../../../lib/resource-helpers.ts';
-import { FormAttachmentResource } from '../../../attachments/FormAttachmentResource.ts';
+import {
+	FormAttachmentResource,
+	type FormAttachmentResourceOptions,
+} from '../../../attachments/FormAttachmentResource.ts';
 import type { ExternalSecondaryInstanceSourceFormat } from './SecondaryInstanceSource.ts';
 
-const assertResponseSuccess = (resourceURL: JRResourceURL, response: FetchResourceResponse) => {
```
latin-panda, line 9: Extracting code to the abstract class so it can be reused when fetching media.

packages/xforms-engine/src/parse/expression/TextChunkExpression.ts
```diff
@@ -15,19 +14,20 @@ const isOutputElement = (element: Element): element is OutputElement => {
 	return element.localName === 'output' && element.hasAttribute('value');
 };
 
-export class TextChunkExpression extends DependentExpression<'string'> {
+export class TextChunkExpression<T extends 'nodes' | 'string'> extends DependentExpression<T> {
```
latin-panda, line 17: Once we add all the support for output, markdown, HTML styling, etc., this could naturally drop the `string` and only be `TextChunkExpression<'nodes'> extend DependentExpression {`. I need to experiment with those scenarios to confirm.

packages/xforms-engine/src/instance/PrimaryInstance.ts
```diff
@@ -279,4 +297,14 @@ export class PrimaryInstance<
 
 		return Promise.resolve(result);
 	}
+
+	loadMediaResources(resourceURL: string): Promise<MediaResource> {
```
latin-panda, line 301: This class provides simple functions to access high-level information, like the active language, and could be extended to retrieve media files, such as images or videos. That's why I added the `loadMediaResources` function here. 

To optimize performance, it avoids querying all media files during the initial `LoadForm` phase, as finding media URLs requires traversing a complex node tree. Instead, it leverages the engine’s XML processing phase when it creates an instance, where it already navigates the node tree to prepare data for all question types. By collecting media information during this process, the engine avoids a second, redundant tree traversal.

Later, we can revisit and find a way to retrieve all media URLs at the initial `LoadForm` phase.

packages/xforms-engine/src/lib/reactivity/text/createTextRange.ts
```diff
@@ -74,10 +138,18 @@ export const createTextRange = <Role extends TextRole>(
 	definition: TextRangeDefinition<Role>
 ): ComputedFormTextRange<Role> => {
 	return context.scope.runTask(() => {
-		const getTextChunks = createTextChunks(context, definition.chunks);
+		const textChunks = createTextChunks(context, definition.chunks);
+		const evaluator = isEngineXPathEvaluator(context.evaluator) ? context.evaluator : null;
```
latin-panda, line 142: The type static-check enforces this validation. I tested several forms (all question types form, all note form, all select form, etc). When creating a Text Range, the evaluator always has a root node that is a Primary Instance. 
But, to play safe, we are leaving this validation.

packages/xforms-engine/src/lib/reactivity/validation/createValidation.ts
```diff
@@ -34,7 +35,8 @@ const engineViolationMessage = <Role extends ValidationTextRole>(
 ): Accessor<EngineViolationMessage<Role>> => {
 	const messageText = VALIDATION_TEXT[role];
 	const chunk = new TextChunk(context, 'literal', messageText);
-	const message = new TextRange('engine', role, [chunk]);
+	const evaluator = isEngineXPathEvaluator(context.evaluator) ? context.evaluator : null;
```
latin-panda, line 38: @lognaturel Can a message display media? I think this message is from question type validations, contrains, ...

packages/xforms-engine/src/lib/reactivity/createItemCollection.ts
```diff
@@ -25,8 +25,8 @@ type DerivedItemLabel = ClientTextRange<'item-label', 'form-derived'>;
 
 const derivedItemLabel = (context: TranslationContext, value: string): DerivedItemLabel => {
 	const chunk = new TextChunk(context, 'literal', value);
-
-	return new TextRange('form-derived', 'item-label', [chunk]);
+	//TODO: work this scenario
+	return new TextRange(null, 'form-derived', 'item-label', [chunk], null);
```
latin-panda, line 29: @lognaturel Does `derived label` ring a bell? A label can be missing (L38), and then this derived label is created. I wonder if it can contain media.

packages/xforms-engine/src/instance/text/TextRange.ts
```diff
@@ -21,9 +33,19 @@ export class TextRange<Role extends TextRole, Origin extends TextOrigin>
 		return this.chunks.map((chunk) => chunk.asString).join('');
 	}
 
+	get image(): Promise<MediaResource> | null {
+		if (this.media?.image == null || !isEngineXPathEvaluator(this.evaluator)) {
+			return null;
+		}
+
+		return (this.evaluator.rootNode as PrimaryInstance).loadMediaResources(this.media.image);
```
latin-panda, line 41: The type guard `isEngineXPathEvaluator` verifies that evaluator is `EngineXPathEvaluator`, which `rootNode` is already a `PrimaryInstance`. However, the type static check is confused because XPath defines `rootNode` as some other general type and does not explicitly define a `PrimaryInstance`. This cast is safe.

packages/xforms-engine/src/client/TextRange.ts
```diff
@@ -149,11 +150,13 @@ export type TextOrigin =
...
 export interface TextRange<Role extends TextRole, Origin extends TextOrigin = TextOrigin> {
+	readonly evaluator: EngineXPathEvaluator | null; // TODO: can it be null?
 	readonly origin: Origin;
 	readonly role: Role;
 
 	[Symbol.iterator](): Iterable<TextChunk>;
 
 	get asString(): string;
 	get formatted(): unknown;
+	get image(): Promise<MediaResource> | null;
```
latin-panda, line 161: Ultimately, it seems simple and easy to understand, so I didn't add extra classes or interfaces for media.

packages/scenario/test/calculate.test.ts
```diff
@@ -164,3 +167,62 @@ describe('MultiplePredicateTest.java', () => {
...
+						mainInstance(
+							t('data id="dynamic-choices-predicates"', t('country'), t('city'), t('city_name'))
+						),
+
+						t(
+							'itext',
+							t(
+								'translation lang="default"',
+								t('text id="static_instance-cities-0"', t('value', 'Montréal')),
+								t('text id="static_instance-cities-1"', t('value', 'Grenoble'))
```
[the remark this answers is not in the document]
latin-panda: Yes, I can add more tests for that. 

> How did you decide to put it here? It's certainly logical. Are there other functions tested here?

I thought, "Can the calculate work with itext?" Then I would go to the `calculate` test suite to check. 

In L76, there's something about accessing text, but no other functions inside `calculate`
```
	bind('/data/calc')
								.type('string')
								.calculate("instance('instance')/root/item[value = 'A'][count = /data/input]/id"),
```

But I can create a new test suite if it feels too weird here.

packages/common/src/fixtures/xforms.ts
```diff
@@ -102,7 +102,7 @@ export class XFormResource<Type extends XFormResourceType> {
 				const service = new JRResourceService();
 				const parentPath = localPath.replace(/\/[^/]+$/, '');
 
-				service.activateFixtures(parentPath, ['file', 'file-csv']);
+				service.activateFixtures(parentPath, ['file', 'file-csv', 'images']);
```
latin-panda, line 105: This is used to load files (forms and attachments) from the common package on the Preview Page when running the project in the local environment. I'm unsure yet what it needs to load attachments on the Preview Page when deployed in central.

packages/common/src/fixtures/xform-attachments.ts
```diff
@@ -23,6 +23,9 @@ const xformAttachmentFileExtensions = [
 	'.xml',
 	'.xml.example',
 	'.xlsx',
+	'.png',
```
latin-panda, line 26: @lognaturel I checked the Collect spec's Select with images, but didn't see the supported types for images. 
Is it expected gif and svg?

packages/common/src/fixtures/import-glob-helper.ts
```diff
@@ -46,7 +52,11 @@ const globFixtureLoader = (globURL: string): GlobFixtureLoader => {
 	return async () => {
```
latin-panda, line 52: This import helper is used for the Preview Page and the scenario tests.

latin-panda:
@lognaturel You can have a look, this is still a draft since I still need to:
- Resolve some ToDos based on the questions I tagged you 
- Add more scenario tests and double-check that existing labels, hints, etc., are working fine.

packages/xpath/test/xforms/jr-itext.test.ts
latin-panda: Issue: https://github.com/getodk/web-forms/issues/400 

These test cases are marked as "failing"

packages/scenario/test/calculate.test.ts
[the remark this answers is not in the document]
latin-panda: > I would have added another language and verified that when you switch the language the /data/city_name value updates.

I've added an additional test case for that here.

packages/xpath/test/xforms/jr-itext.test.ts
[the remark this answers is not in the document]
latin-panda: Done

packages/common/src/fixtures/xform-attachments.ts
latin-panda: I've added GIF and SVG extensions to this list

packages/xforms-engine/src/client/TextRange.ts
lognaturel: Can we avoid the null? Maybe do a rejected promise instead?

packages/xforms-engine/src/lib/reactivity/validation/createValidation.ts
latin-panda: I'm leaving validation messages without images because the spec doesn't mention media support for messages and it's out of scope (do we display red background? or red border?, etc.)

packages/xforms-engine/src/lib/reactivity/text/createTextRange.ts
```diff
@@ -1,67 +1,131 @@
...
+};
+
+/**
+ * The function temporarily supports a <value> node with multiple nested nodes.
+ * TODO: Build support for <output> nodes.
+ *  A child might be a node that needs XPath to compute its value. For that, the engine
+ *  should create a {@link: TextElementDefinition}, so that createTextChunks function can
+ *  request the computed value to XPath and create the TextChunk.
+ */
+const processValueNodeChildren = (item: EngineXPathNode) => {
```
lognaturel, line 66: H todo: understand this better

lognaturel (commented):
Can you help me understand when image requests happens a little bit better? Here's what I think happens:
- When a client gets to a node that has an image and accesses it, the engine will request the image
- Currently with our Vue client, that means all the requests will happen in parallel when the form is first rendered
- If the user switches languages, images for the new language will be requested
- The behavior when an image is missing depends on the "missing attachment" behavior defined. If it's set to be BLANK the image will just be blank. Otherwise a 404 will be thrown and the form won't work.

I made some comments about `null` that I think are important. `null` propagates everywhere and makes reasoning about values much harder. Also, I think `undefined` is the more idiomatic way to go in JS/TS. I've made some suggestions for getting rid of some `null`s, I think it should be possible to do away with all, hopefully! Maybe we can make that an explicit code style goal -- avoid `null` at (almost) all cost.

packages/common/src/fixtures/xform-attachments.ts
lognaturel: Why not just let the browser support whatever it can? It's less precise than defining the set of supported image types but could offer some appreciated flexibility. I don't feel very strongly either way but philosophically I find it ok to be a little more loose for things like this, especially if it matches what Collect and Enketo do.

packages/xpath/src/xforms/XFormsXPathEvaluator.ts
```diff
@@ -103,13 +103,13 @@ export class XFormsXPathEvaluator<T extends XPathNode>
 		return context.evaluator.secondaryInstancesById.get(id) ?? null;
 	}
 
-	static getDefaultTranslationText<T extends XPathNode>(
+	static getTranslationTextValues<T extends XPathNode>(
 		context: EvaluationContext<T>,
 		itextID: string
-	): string {
+	) {
```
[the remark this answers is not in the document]
latin-panda: Fixed

packages/xpath/src/xforms/XFormsItextTranslations.ts
```diff
@@ -156,51 +162,43 @@ export class XFormsItextTranslations<T extends XPathNode> implements XFormsItext
...
+	 *   /translation[@lang = xpath3-fn:environment-variable('activeLanguage')]
+	 *     /text[@id = $itextID]
+	 *       /value
 	 * ```
-	 *
-	 * @todo The above really feels like it adds some helpful clarity to how `jr:itext()` is designed to work, and the kinds of structures, state and input involved. Since there's already some discomfort around that API as specified, it's worth considering
+	 * Returns an array of `XFormsItextTranslationValueElement<T>` or `null` if no matching `text` element is found.
 	 */
-	getDefaultTranslationText(itextID: string): string {
+	getTranslationValues(itextID: string): Array<XFormsItextTranslationValueElement<T>> | null {
```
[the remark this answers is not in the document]
latin-panda: Fixed

packages/xpath/src/xforms/XFormsXPathEvaluator.ts
```diff
@@ -103,13 +103,13 @@ export class XFormsXPathEvaluator<T extends XPathNode>
 		return context.evaluator.secondaryInstancesById.get(id) ?? null;
 	}
 
-	static getDefaultTranslationText<T extends XPathNode>(
+	static getTranslationTextValues<T extends XPathNode>(
```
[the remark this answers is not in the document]
latin-panda: Fixed

packages/xforms-engine/src/parse/expression/TextChunkExpression.ts
```diff
@@ -15,19 +14,20 @@ const isOutputElement = (element: Element): element is OutputElement => {
...
 	constructor(
 		context: AnyTextRangeDefinition,
+		resultType: T,
 		expression: string,
 		source: TextChunkSource,
 		options: TextChunkExpressionOptions = {},
 		literalValue = ''
 	) {
-		super(context, 'string', expression, {
+		super(context, resultType, expression, {
```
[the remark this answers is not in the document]
latin-panda: No changes for this line (explained in previous comment)

packages/common/src/fixtures/xform-attachments.ts
```diff
@@ -86,6 +89,18 @@ export class XFormAttachmentFixture {
 				this.mimeType = 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet';
 				break;
 
+			case '.png':
+				this.mimeType = 'image/png';
+				break;
+
+			case '.jpg':
+				this.mimeType = 'image/jpg';
```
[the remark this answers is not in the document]
latin-panda: It's already passing Blob type around, and the Vue client receives a Blob.

This mime type is used to set the [content-type in the request](https://github.com/getodk/web-forms/blob/da4b22020c5bc651f1cda8f4a83a67a8a696badd/packages/common/src/jr-resources/JRResourceService.ts#L44), but I think the main intention here is to have high control over the attachments Web Form can support. 

If we aren't interested in having complete control over supported attachment types. In that case, we can remove the mimeType and switch block completely. I'd open a separate PR for the record that can be merged before this one, as the change spreads over some files.

What do you think?

packages/common/src/fixtures/xform-attachments.ts
latin-panda: [I left a comment about this here](https://github.com/getodk/web-forms/pull/399#discussion_r2085767466)

packages/xforms-engine/src/lib/reactivity/text/createTextRange.ts
latin-panda: The evaluator, in this case, is only used for retrieving media attachments. I've added it as part of the media context, so its role is clearer in the code.

packages/common/src/fixtures/xform-attachments.ts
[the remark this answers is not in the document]
lognaturel: Got it! Let’s leave as-is for now and we can revisit if we find that requirements broaden or that we don’t do anything with the more limited/controlled types.

packages/xforms-engine/src/parse/expression/TextChunkExpression.ts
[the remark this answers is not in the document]
lognaturel: Love it, I don’t see a lot of benefit to generics in this context.

packages/xforms-engine/src/parse/model/SecondaryInstance/sources/ExternalSecondaryInstanceResource.ts
```diff
@@ -183,11 +151,23 @@ export class ExternalSecondaryInstanceResource<
...
-					`Failed to load blank external secndary instance ${resourceURL.href}`
+					`Failed to load blank external secondary instance ${resourceURL.href}`
 				);
 			}
 		} else {
 			this.isBlank = false;
 		}
 	}
+
+	isCSVResource(): this is ExternalSecondaryInstanceResource<'csv'> {
```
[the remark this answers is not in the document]
latin-panda: This is a TypeScript type guard, similar to `isEngineXPathEvaluator(context.evaluator)` in `createTextRange.ts`. It ensures TypeScript recognizes that when `format` is `csv`, the instance is an `ExternalSecondaryInstanceResource<'csv'>`.

After extracting code from `ExternalSecondaryInstanceResource.ts` to `FormAttachmentResource`, TypeScript struggled to infer the type of the `resource` object in the switch statement in `SecondaryInstancesDefinition.ts`.

The switch statement is safe, as it validates `resource.format` before creating the corresponding object. However, TypeScript fails to narrow the type correctly, resulting in the following error:

```
src/parse/model/SecondaryInstance/SecondaryInstancesDefinition.ts:97:65 - error TS2345: Argument of type 'ExternalSecondaryInstanceResource<ExternalSecondaryInstanceSourceFormat>' is not assignable to parameter of type 'ExternalSecondaryInstanceResource<"csv">'.
  Type 'ExternalSecondaryInstanceSourceFormat' is not assignable to type '"csv"'.
    Type '"xml"' is not assignable to type '"csv"'.

97       return new CSVExternalSecondaryInstanceSource(domElement, resource);
```

I've removed the type guard and use direct casting in the switch statement:

```js
switch (resource.format) {
  case 'csv':
    return new CSVExternalSecondaryInstanceSource(domElement, resource as ExternalSecondaryInstanceResource<'csv'>);
 etc....
```

latin-panda:
> Can you help me understand when image requests happens a little bit better? Here's what I think happens:

Sure! @lognaturel thanks for the feedback! Let me know if you have any other suggestion

My comments below:

> When a client gets to a node that has an image and accesses it, the engine will request the image

Yes, a Vue component can make the request, for example, for each select's option: `options.forEach(option => await option.label.image)`

Another idea I'm playing around with is letting the engine's interface do the requests, so the Vue component calls one function that retrieves all images: `const images: Array<Blob> = await question.getAllImages()`

I'm making this as part of the front-end work.

> Currently with our Vue client, that means all the requests for the active language will happen in parallel when the form is first rendered

Correct, the requests are in parallel. Going back to the previous question: 
- If the Vue component handles the request, it will render the images as soon as they are available. 
- If the engine's client interface handles the request (`await question.getAllImages()`), it will wait for all images of a question to be available before passing the array of images to the Vue component. 

Usually, the first one is preferred, as it displays content as quickly as possible. In the select with images context, it will still be better the first option, so the user sees that it is loading (in slow connections) instead of waiting with an empty state. What do you think?

> If the user switches languages, images for the new language will be requested

Yes, it only requests images from the active language.  When a new language is chosen, it requests the images for that language. 

> The behavior when an image is missing depends on the "missing attachment" behavior defined. If it's set to be BLANK the image will just be blank. Otherwise a 404 will be thrown and the form won't work.

Yes, we have a [placeholder](https://github.com/getodk/web-forms/issues/347#issuecomment-2847454141) for when the image can't be loaded.

packages/xforms-engine/src/parse/attachments/FormAttachmentResource.ts
```diff
@@ -30,11 +32,59 @@ type FormAttachmentData<DataType extends FormAttachmentDataType> =
...
+
+export interface LoadResult<T extends FormAttachmentDataType> {
+	response: FetchResourceResponse;
+	data: FormAttachmentData<T>;
+	isBlank: boolean;
+}
+
 export abstract class FormAttachmentResource<DataType extends FormAttachmentDataType> {
 	protected constructor(
 		readonly dataType: DataType,
```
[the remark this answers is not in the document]
lognaturel: I don't think this ends up applying in this case but keeping the comment to jog my own memory and related to https://github.com/getodk/web-forms/pull/399#discussion_r2087273054

packages/xforms-engine/src/parse/model/SecondaryInstance/sources/ExternalSecondaryInstanceResource.ts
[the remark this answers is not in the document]
lognaturel: Here's my alternate proposal: https://github.com/getodk/web-forms/commit/3f2556b6b698cfb0ae2b4fdd29d90734bd4b0b3e

The first commit is more conservative -- it removes indirection and some generics that I don't find useful. Then the cast is not necessary.

The second commit is more aggressive and opinionated -- it removes speculative types and all generics. We end up with an abstract `FormAttachmentResource` that's just for limited sharing of behavior between media attachments and secondary instance attachments. It's not for polymorphic use because there is no polymorphic use anyway. It feels a bit weird that `data` ends up defined on each of the concrete implementations but I think it's clear and maintains type safety where it's helpful (a secondary instance resource has text data, a media resource has Blob data). 

`fetch` ends up being very much duplicated between the two classes. I think that's ok because we don't expect more subclasses and we don't expect polymorphic use. My hunch is that the two will evolve separately because the different form attachments have different download needs, different behavior when blank, etc. I find this pattern easier to follow than having an extra parameter to represent the type of resource and branching on it in the abstract class.

packages/xforms-engine/src/instance/SelectControl.ts
```diff
@@ -225,4 +226,13 @@ export class SelectControl
 
 		return this.root;
 	}
+
+	loadImage(item: SelectItem): Promise<MediaResource> {
```
latin-panda, line 230: Vue component:

```
// Watch for changes in valueOptions (like new language), then: 
  for (const option of props.question.currentState.valueOptions) {
    try {
      const result = await props.question.loadImage(option);
      if (!result.data instanceof Blob) {
         // cannot load, use placeholder
	continue;
      }

      const blobUrl = URL.createObjectURL(result.data);
      images.value.set(option.value, blobUrl);

    } catch (error) {
         // cannot load, use placeholder
    }


/// template
<img
  v-if="imageUrls.get(option.value)"
  :src="images.get(option.value)"
  :@error="handleImageError"
>
``

File: packages/xforms-engine/src/client/SelectNode.ts [hunk 25]
```diff
@@ -19,6 +19,8 @@ export interface SelectItem {
 export type SelectValueOptions = readonly SelectItem[];
 
 export interface SelectNodeState extends BaseValueNodeState<readonly string[]> {
+	get isSelectWithImages(): boolean;
```
latin-panda, line 22: I think this will go away, and not changes in this file

packages/xforms-engine/src/instance/SelectControl.ts
```diff
@@ -51,6 +51,7 @@ interface SelectControlStateSpec extends ValueNodeStateSpec<readonly string[]> {
 	readonly label: Accessor<TextRange<'label'> | null>;
 	readonly hint: Accessor<TextRange<'hint'> | null>;
 	readonly valueOptions: Accessor<SelectValueOptions>;
+	readonly isSelectWithImages: Accessor<boolean>;
```
latin-panda, line 54: All these changes might be reverted and handled in the front-end side

lognaturel:
We've made a high-level decision to pass JR URLs through to the client for media. This is a departure from how external secondary instances and submission media attachments are handled but it's a deliberate choice based on the current lack of need to have the engine do any media processing. This could change in the future but is consistent with the JavaRosa/Collect split. JR URLs are exposed through the `TextRange` client API. That naming doesn't feel as appropriate anymore but we want to learn more about the evolution needed around these concepts before renaming things.

There's some kind of API test missing here: we should test that on language changes, the JR URLs on `TextRange` change appropriately. We'll defer that since we know we have more to do in this area. There aren't really any edge cases here -- it either works or doesn't -- so we can verify this in the Vue client for now.

We've gone with a minimally-invasive approach. The `jr:itext` function is now a nodeset function. When the language changes, the `jr:itext` function call that defines a form element's text is made again, getting the new language's itext block. The itext block is split into its different "form"s (default, image, video, audio) at evaluation time.

There are some things we know are uncomfortable:
- itext block parsing feels like it should be done at parse time
- the `string | nodes` return type on `TextChunkExpression` suggests there's more than one concept there. Indeed, a translation expression is very different from a reference expression or literal text
- `TextRange` feels like the wrong name for something that can contain media

We'll continue to work towards addressing those in future work.

lognaturel (approved):
I put some notes on how I'm thinking about this work at https://github.com/getodk/web-forms/pull/399#issuecomment-2905510287

I haven't tried this with a frontend and there are no API tests so I think it's important to hook things up before merging to do a quick sanity check, @latin-panda 

Other than that, I'm comfortable with merging!

lognaturel:
I thought the changes from [11fbc00](https://github.com/getodk/web-forms/pull/399/commits/11fbc00e14cae3ddc587d69b5377df54f7ed0bbc) would be confusing once we squashed so I reverted them for now. We could apply them as a separate set of changes or wait until we are in that part of the code again. The latter would be my preference because I think we'll understand a lot more by then and may choose a different direction.

packages/xforms-engine/src/lib/reactivity/text/createTextRange.ts
```diff
@@ -1,128 +1,72 @@
...
-};
-
-/**
- * The function temporarily supports a <value> node with multiple nested nodes.
- * TODO: Build support for <output> nodes.
- *  A child might be a node that needs XPath to compute its value. For that, the engine
- *  should create a {@link: TextElementDefinition}, so that createTextChunks function can
- *  request the computed value to XPath and create the TextChunk.
- */
-const processValueNodeChildren = (item: EngineXPathNode) => {
```
lognaturel: If I recall correctly, this was introduced to re-establish passable behavior for itext default forms with outputs in them. I don't think it's necessary for this PR or if we do want to do it here we should introduce a test related to it.

PR #404, opened 2025-05-13
Prepare release with support for rows attribute on text

lognaturel:
Prepares release

I added the `rows` parameter and in doing so noticed that the `label` and `value` parameters were marked as incomplete. I verified that they are fully supported with the following form: 
[external-data(1).csv](https://github.com/user-attachments/files/20197341/external-data.1.csv)
[select-one-file(1).xlsx](https://github.com/user-attachments/files/20197342/select-one-file.1.xlsx)

### DIFF
--- a/.changeset/icy-llamas-fail.md
+++ b/.changeset/icy-llamas-fail.md
@@ -1,4 +1,0 @@
----
-'@getodk/web-forms': minor
----
-Add support for a multiline text question type.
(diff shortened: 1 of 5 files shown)


lognaturel:
Looks like the Github action conditions don't work from a fork! I think we can look at that later -- there are no substantive changes here.

latin-panda (approved):
Thank you!

Issue #76, opened 2024-04-12
Expose as custom element


sadiqkhoja:
Or have some solution to address the needs of non-Vue host application

jkuester:
Super excited to see a web-component solution for this issue in the works!  :star_struck:

lognaturel:
We've decided to deprioritize this for the moment because there are some changes coming to both Vue and PrimeVue that would make a less hacky solution possible.

In the short term, we want to focus on https://github.com/getodk/web-forms/issues/151 so that anyone can easily try out their XLSForms and evaluate progress relative to their needs.

If someone is really eager to embed Web Forms in a non-Vue application, https://github.com/getodk/web-forms/pull/106 does have a working custom element solution.

latin-panda:
## ODK Web Forms Custom Component Implementation

**Context**: Implementing the `OdkWebForm` component as a Vue Custom Component (i.e., Web Component) in the current stack (Vue 3.5.x, PrimeVue 4.3.x, Vite 6.3.0) achieves style isolation via Shadow DOM and enables non-Vue apps to use Web Forms. The goal is to encapsulate the component and its dependencies (e.g., PrimeVue, Vue-draggable-plus) for consumer apps like the Central front-end.

We aim to use Vue’s `defineCustomElement` for Shadow DOM isolation but are currently facing issues with PrimeVue integration and form initialization. While consumer apps expect `<OdkWebForm>` as a Vue component, we’ve tested `<odk-web-form>` as a custom element.

### Implementation
(Expand the sections to see the code)

<details><summary>1. Declare Custom Component in 'index.ts'. 
Define 'OdkWebForm' as a Web Component with 'defineCustomElement' to enable Shadow DOM and register it as &lt;odk-web-form&gt;. </summary>

**Code** ([index.ts](https://github.com/getodk/web-forms/blob/main/packages/web-forms/src/index.ts)):

```javascript
import { defineCustomElement } from 'vue';
import { webFormsPlugin } from './WebFormsPlugin';
import OdkWebFormComponent from './components/OdkWebForm.vue';
import { POST_SUBMIT__NEW_INSTANCE } from './lib/constants/control-flow.ts';

import '@fontsource/roboto/300.css';
import '@fontsource/roboto/400.css';
import '@fontsource/roboto/500.css';
import 'primeflex/primeflex.css';
import './assets/css/style.scss';

const OdkWebForm = defineCustomElement(OdkWebFormComponent, {
[...]

PR #409, opened 2025-05-19
fix: normalize web forms styles

### LINKED ISSUE #395: Rank: when an item is being moved, the whole page content shifts down by 1px
https://github.com/user-attachments/assets/b045d73b-d686-472f-bf9e-85dc35a53fc5

### LINKED ISSUE #401: The rank question overlay can be offset from the choices
From [this form](https://docs.google.com/spreadsheets/d/1GVNFQtWn4Q74l7NSWv6S_WcAhsfkkcFnIDSYePZp-jc/edit?gid=1623666754#gid=1623666754):
Image

I've confirmed it doesn't happen with [the all questions form on the preview site](https://getodk.org/web-forms-preview/app/index.html#/form?url=https://getodk.org/web-forms-preview/app/assets/all-question-types-v2024091201-LqEndp46.xml)

### LINKED ISSUE #407: Tweaking the rank options
- Lighten the disabled button so it's not so distracting. 

- [Edit] the new line height solves this problem 
~~Make the height on mobile the same as desktop. The selectable options are smaller on mobile and with the reduced spacing between each item it's hard to select and drag.~~ 

Image

### LINKED ISSUE #408: Center the logo and version number
We want to center the logo and version number at the bottom of the page

[Figma](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=1456-16775&t=R6mmMIGbzajxBesK-1)

latin-panda:
Depends on: https://github.com/getodk/central-frontend/pull/1269

Closes #408, Closes #407, Closes #401 Closes #395

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [x] Safari (iOS)
- [x] Chrome for Android
- [ ] Not applicable

I have set up Central, running with the latest local WF, and have been testing the minimalistic reset stylesheet and the scoping by.odk-form To see if anything is wrong.

> all: revert 

It breaks the drag-and-drop functionality of the third-party library WF has. It likely relies on properties such as touch-action, transform, and user-select, among others. This can't be applied to all elements in the DOM. It's added just to `body` element. 

> Vue custom component for encapsulation 

Central can utilize them when props are simple and no complex external dependencies are involved. However, when making WF a custom component, there are some difficulties in incorporating PrimeVue; I[ have added more details here ](https://github.com/getodk/web-forms/issues/76#issuecomment-2888083108)from my experience trying it out.

>  https://github.com/necolas/normalize.css

It hasn't been updated in a considerable amount of time. I found https://github.com/sindresorhus/modern-normalize, which is more updated with recent browser features. Not much effect on Central's style and Central's Bootstrap, so just taking a couple things from it but not using the library.  

I included some basic normalization rules and made a reset stylesheet that is imported as the first thing in WF.  It's not scoped by global classes yet (.odk-forms) because it overrides PrimeVue (even though it is imported after the reset.css)

> .odk-form on the root element can be used

WF is scoping CSS on .odk-form more

Manual testing and comparing with current preview mode of WF

### What's changed

- Styles and removing some leftover code


packages/web-forms/src/WebFormsPlugin.ts
```diff
@@ -1,9 +1,14 @@
 import { odkThemePreset } from '@/odkThemePreset.ts';
 import { type App } from 'vue';
 import PrimeVue from 'primevue/config';
+import resetStyles from './assets/css/reset.scss?raw';
 
 export const webFormsPlugin = {
 	install(app: App) {
+		const styleElement = document.createElement('style');
```
latin-panda, line 8: Adds reset style-sheet as the first thing in the HTML's header element. 

But Central still overpowers WF's reset style

packages/web-forms/src/assets/css/reset.scss
```diff
@@ -0,0 +1,97 @@
+*,
+::before,
+::after {
+	box-sizing: inherit;
+}
+
+html {
+	font-size: 16px !important;
```
latin-panda, line 8: Not a fan of using `!important`, but that's how it's overriding Central's and Bootstrap's styles.

Not wrapping all this in a Web Form's super class, because it overrides PrimeVue classes :(

packages/web-forms/src/components/controls/Input/Geopoint/GeolocationRequestDialog.vue
```diff
@@ -123,7 +123,7 @@ onBeforeUnmount(cleanup);
 		<template #header>
 			<div class="geo-dialog-header">
 				<div class="geo-dialog-header-title">
-					<ProgressSpinner class="spinner" stroke-width="4" />
+					<ProgressSpinner class="geo-spinner" stroke-width="4" />
```
latin-panda, line 126: Conflicts with Central's class names.

This code is loaded in a dialog. And the dialogs are at the end of the DOM, outside the Web Forms super class. 
This is the fastest fix. We need to scope that later.

packages/web-forms/vite.config.ts
```diff
@@ -144,34 +144,9 @@ export default defineConfig(({ mode }) => {
 			preprocessorOptions: {
 				scss: {
 					api: 'modern',
+					quietDeps: true, // Suppress warnings from node_modules
 				},
 			},
-			postcss: {
```
latin-panda, line 149: I forgot to remove this from the PrimeVue upgrade to v4. 

This code was needed for the deprecated library: primevue-sass-theme
And the primevue in this project, doesn't work with CSS layers anymore

latin-panda:
@lognaturel @sadiqkhoja I'm going to do Android tests and another pass on desktop tonight. You can start reviewing :)

lognaturel (approved):
I've verified this thoroughly on iOS and Android and am satisfied that it fixes our immediate issues.

The reset approach looks reasonable to me. I think we will need to continue discussing and exploring our vision for custom styling both from form managers and host apps. But for now, saying that Web Forms resets styles seems reasonable.

I feel like between what @latin-panda and I have done, we have enough verification for now. Longer term, we should think about how we test the Central integration.

PR #429, opened 2025-06-06
feat(#425): adds support for multiline appearance in input text

### LINKED ISSUE #425: Select and multiselect are not disabled when set to readonly
### Description

Selects with dropdowns are editable when set to 'readonly'. We must check other select types.  

Note: In this video, I have added text to display the read-only value on screen; it's for test purposes only. 

https://github.com/user-attachments/assets/b3bd7d9a-c64f-4e85-b662-47eed60446dc

Upload this form: [readonly-selects.xlsx](https://github.com/user-attachments/files/20513977/Copy.of.all-question-types-v2024091201-2.1.xlsx)

### Expected
1. The "Select one search" question should be disabled when the "Select one" question has a selected value
2. The "Multi select autocomplete" question should be disabled when the "Multi select" question has a selected value

latin-panda:
Closes #425

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

Test videos

- Regular input text without appearance and input text with `multiline` appearance

https://github.com/user-attachments/assets/462db7d7-38dc-42c8-bfd5-3065d5462fb6

- Input text with `rows` property 

https://github.com/user-attachments/assets/d015f8ef-8c31-4245-bf10-2e1676efb444

### What's changed

- Adds support for `multiline` appearance

### DIFF
--- a/.changeset/afraid-dogs-march.md
+++ b/.changeset/afraid-dogs-march.md
@@ -1,0 +1,5 @@
+---
+'@getodk/web-forms': patch
+---
+
+Adds support for multiline appearance in input text
--- a/packages/web-forms/src/components/controls/Input/InputText.vue
+++ b/packages/web-forms/src/components/controls/Input/InputText.vue
@@ -4,6 +4,8 @@ import InputText from 'primevue/inputtext';
 import Textarea from 'primevue/textarea';
 import { computed, inject } from 'vue';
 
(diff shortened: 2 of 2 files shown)


alyblenkin:
Also seems to work the way I would expect.

lognaturel (approved):

PR #419, opened 2025-05-22
docs: adds guidelines for contributors

latin-panda:
### I have verified this PR works in these browsers (latest versions):

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [x] Not applicable

### What else has been done to verify that this works as intended?

NA

### Why is this the best possible solution? Were any other approaches considered?

NA

### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?

NA

### Do we need any specific form for testing your changes? If so, please attach one.

NA

### What's changed

- Adds guidelines for contributors
- Adds lock file to /scripts
- Adds engine info to package.json in /scripts 
- Adds data encryption to feature matrix

PR #431, opened 2025-06-09
Move script dependencies to top-level package.json

### LINKED ISSUE #430: Dependency installation for scripts/ dir is not documented
Either I've missed something, or the `scripts/` or either:

1. setup for `./scripts` is not documented, or
2. `scripts/package.json` could just be merged with top-level `package.json`

What's the correct procedure for initialising dependencies for `./scripts`?

alxndrsn:
Closes #430 

Maybe?

### DIFF
--- a/scripts/package.json
+++ b/scripts/package.json
@@ -5,9 +5,5 @@
   "type": "module",
   "scripts": {
     "test": "true"
-  },
-  "devDependencies": {
-    "globby": "^14.0.2",
-    "mustache": "4.2.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -67,6 +67,10 @@
     "vue": "3.5.13",
     "vue-tsc": "^2.2.8"
   },
+  "devDependencies": {
+    "globby": "^14.0.2",
+    "mustache": "4.2.0"
+  },
   "resolutions": {
     "**/tree-sitter": "0.22.1"
   }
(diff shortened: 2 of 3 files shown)


scripts/package.json
```diff
@@ -5,9 +5,5 @@
   "type": "module",
   "scripts": {
     "test": "true"
-  },
-  "devDependencies": {
```
latin-panda, line 9: The intention is to keep the root's package.json file minimal. In my opinion, it's okay to keep `globby` and `mustache` dependencies here because they are not used anywhere else, and these scripts aren't vital for WF functioning. 

What do you think?

alxndrsn: > these scripts aren't vital for WF functioning

I think that's the purpose of `devDependencies`.

What are the benefits of a minimal root `package.json`?

If these dependencies are kept in `scripts/package.json`, would you be open to a PR to add `scripts/yarn.lock`?

I think additionally, `scripts/package.json` would then also replicate some of the stuff in `/package.json` which standardises tools, specifically:

```json
{
  "engines": {
    "node": "^18.20.5 || ^20.18.1 || ^22.12.0",
    "yarn": "1.22.22"
  },
  "volta": {
    "node": "22.12.0",
    "yarn": "1.22.22"
  },
  "packageManager": "yarn@1.22.22"
}
```

alxndrsn: Alternatively, maybe `scripts/package.json` can be removed completely.

latin-panda: I like the idea of adding `engines`, `volta`, and a lock file to /scripts, which will ensure consistency and maintain isolation and separation from the source. I can prepare a PR with those changes :)

latin-panda:
We discussed this during the team meeting and decided to keep the /script with its dependencies. We have added the lock file and node engine definition to the package.json to align with the root. Addressed [here](https://github.com/getodk/web-forms/pull/419). cc: @lognaturel

PR #453, opened 2025-07-21
feat(#390): Improves error dialog style and message

### LINKED ISSUE #390: Show better error message when one or more form attachments are missing
#### Problem description
Unable to fill submission using Web Forms when a geojson attachment is missing (404).
I also noticed this when trying to use a form with last saved option in WFs – json file (but I don’t know if last saved is supposed to work well in WFs – languageTest form).
#### URL of the page
https://staging.getodk.cloud/projects/24/forms/all-question-types/submissions/new

https://staging.getodk.cloud/projects/146/forms/languageTest/submissions/new

#### Steps to reproduce the problem
1. Upload a form which needs a geojson attachment (e.g. All question types form).
2. Don’t upload the geojson attachment.
3. Publish the form.
4. Change the settings to use WF.
5. In Submissions tab click +New to open the submission page.
#### Screenshot

Image

#### Expected behavior
I'm not sure what is the expected bahavior- a user-friendly error message or being able to send the submission even without the attachment.
#### Central version shown in version.txt
https://staging.getodk.cloud/
versions:
64a785e60c911469ac60ade9635ccdc5a1bfa47d (v2024.3.2-35-g64a785e)
+50683f824573ca88d28f3ce72c08876aa100e540 client (v2024.3.1-91-g50683f82)
+90859b3173c05c6ea101ded5211eb6b75d4069cc server (v2024.3.1-84-g90859b31)

#### Browser
Chrome and Firefox

latin-panda:
Closes #390

Before the improvements:

- Printing a long message in red
- Hard to scroll to the technical details

https://github.com/user-attachments/assets/51d3a768-79d8-4dce-81ad-d2810252a695

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

Video showing the new user experience: 
- Moves technical details to a collapsible section, keeping the message for the user simple.
- The dialog is less wide
- The font isn't bold

https://github.com/user-attachments/assets/a4fc8d8a-02a4-4db4-b1d1-3ce048778491

### What's changed
- Updated dialog styles as per [designer notes](https://getodk.slack.com/archives/C051L6AKG4F/p1752011473893869). 
- Removed the raw data from the message, which was printing the entire XML, resulting in a lengthy wall of red text.
- Moved the error message to the technical section and part of the stack info.

### DIFF
--- a/.changeset/stupid-clocks-spend.md
+++ b/.changeset/stupid-clocks-spend.md
@@ -1,0 +1,6 @@
+---
+'@getodk/xforms-engine': patch
+'@getodk/web-forms': patch
+---
+
+Improves error dialog message and style.
--- a/packages/web-forms/src/components/FormLoadFailureDialog.vue
+++ b/packages/web-forms/src/components/FormLoadFailureDialog.vue
@@ -47,42 +47,55 @@ const detail = computed((): FormLoadErrorDetail | null => {
 		:keep-in-viewport="true"
 	>
 		<div class="content">
-			<Message severity="error" class="message" :closable="false">
+			<Message severity="error" class="message" :closable="false" :unstyled="true">
 				{{ error.message }}
 			</Message>
 
 			<details v-if="detail != null" class="initialize-form-failure-details">
 				<summary>{{ FORM_LOAD_ERROR_TEXT.DETAILS_SUMMARY_LABEL }}</summary>
 
-				<pre v-if="detail.unknownCauseDetail != null">{{
-				detail.unknownCauseDetail
-				}}</pre>
+				<pre v-if="detail.unknownCauseDetail != null">{{ detail.unknownCauseDetail }}</pre>
 
-				<pre v-if="detail.stack != null">{{
-				detail.stack
-				}}</pre>
+				<pre v-if="detail.stack != null">{{ detail.stack }}</pre>
 			</details>
 		</div>
 	</Dialog>
 </template>
 
-<style scoped>
-.form-load-failure-dialog .content {
-	width: calc(100dvw - 6rem) !important;
-}
-
+<style lang="scss" scoped>
 .form-load-failure-dialog .message {
-	margin-top: 0;
+	margin: 10px 0 30px 0;
+	color: var(--odk-error-text-color);
+	font-size: var(--odk-base-font-size);
 }
 
 .initialize-form-failure-details {
 	display: block;
 	position: relative;
 	max-width: 100%;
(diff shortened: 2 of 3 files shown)


latin-panda:
Hi @lognaturel, this is ready for review :)

lognaturel:
Thanks for those details!

What is the literal string "Raw string data" and where is it coming from? Can we remove it?

My ideal would be to remove the literal string "Failed to load form resource" because it's redundant with the title and instead pull just the error message above the technical details. For example, "Failed to load resource: jr://file/towns.xml: resource is missing (status: 404)" or "Failed to evaluate node for expression". Those are helpful, actionable messages that feel buried if they're only under technical details.

Ideally if there are two missing attachments or a missing attachment and a bad expression both messages would show and both stack traces would be in the details. It's not a big deal if we only show the first one for now.

latin-panda:
> What is the literal string "Raw string data" and where is it coming from? Can we remove it?

It's coming from the same file [LoadFormFailureError](https://github.com/getodk/web-forms/pull/453/files#diff-de99f975272ba58fc4879a9b597acb34501b43312e5b9b9d23e04562bb825a3c). If the resource is a blob, then it displays the name; if it's a URL, then it displays the URL, or defaults to `Raw string data`. It's not particularly useful, but still okay to display the blob name and URL, then remove the default `Raw string data`.

I'm applying the rest of your suggestion, thanks!

latin-panda:
Continuing the conversation 

The blob is not an attachment, but the XML form sent by the front-end client to the engine. The xform-engine interface [supports blobs, URLs, and strings](https://github.com/getodk/web-forms/blob/f8000cf70cf9ec2b245d42b105ad11303ba93f89/packages/xforms-engine/src/error/LoadFormFailureError.ts#L55). However, our Vue client only sends strings, so this support is intended for other clients that want to send different formats. 

Example of a bad XML, as a blob, by the Vue client to the engine: 
```js
const init = async () => {
	state.value = await loadFormState(
            new File([new Blob([`<form><field name="test">value</field></form>`], { type: "application/xml" })], "my-form.xml", { type: "application/xml" }), 
           {
		form: formOptions,
		editInstance: props.editInstance ?? null,
	    }
        );
};
```

The error message: 
Screenshot 2025-07-23 at 1 25 29 AM

latin-panda:
@lognaturel, this is ready for review. If approved, I'll start the release tomorrow 🤩

PR #454, opened 2025-07-21
fix(#421): XML declaration in submission leads to crash on edits

### LINKED ISSUE #421: XML declaration in submission leads to crash on edits
Some submissions start with `<?xml version='1.0' encoding='UTF-8' ?>`. On editing that submission, Web Forms throws "Invalid instance input. Expected instance id to be "xyz", got: "null".

latin-panda:
Closes #421

- [x] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

- There is a new scenario test confirming this case.
- Ran Central, Collect, and Web Forms with this fix, following the steps below, and verified they are working correctly.

Removing the XML declaration before processing because:
- Some submissions don't have an XML declaration, so including it is not mandatory.
- The XML engine currently supports UTF-8, and the declaration does not provide additional data required by the engine. We could add validation messages to inform form designers if the encoding or version specified in the XML declaration is incorrect, but this seems unnecessary at the moment. They can also test their forms to verify all is working.

N/A

This is tricky to test and requires running Central and Collect:

- Build Web Forms.
- Point Central's Web Forms dependency to your Web Forms code and install it.
- Expose Central using a tunneling tool like ngrok. Add the tunnel's URL to the vite.config.ts server configuration section.
- Run Central.
- Install Collect.
- Create a project in Central with a form and grant a user access.
- Access the form in Collect using that new user.
- Create a submission and let it sync the information.
- Go to Central and edit the submission.
- It should not throw an error, and editing should be possible.

The other option is to temporarily hard-code the submission on WF's demo page for testing: 

Submit using Collect, then go to Central and preview the new submission by adding `v1` before the "projects" and `.xml` at the end of the URL, for example: 

> https://test.getodk.cloud/v1/projects/617/forms/date/submissions/uuid:<the-submission-uuid>.xml

Use the browser's developer tools to see the full definition, and copy the entire XML. 
Go to `packages/web-forms/src/demo/FormPreview.vue` and add the code:

```js
const WithPrefix =
	'<?xml version=\'1.0\' encoding=\'UTF-8\' ?><data id="date" version="2025020402" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:orx="http://openrosa.org/xforms" xmlns:odk="http://www.opendatakit.org/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa"><dates><survey_date>2025-07-16</survey_date><date_of_birth>2025-07-30</date_of_birth><fruits_date>2025-07-19</fruits_date><vegetables_date>2025-07-31</vegetables_date></dates><meta><instanceID>uuid:08ea72ce-745b-4158-a095-1b7d8e48f31c</instanceID></meta></data>';

const editInstanceOptions = {
	resolveInstance: () => Promise.resolve(WithPrefix),
	attachmentFileNames: [],
	resolveAttachment: Promise.resolve(null),
};
```

Send the `editInstanceOptions` to `OdkWebForm` as input (all in FormPreview.vue) 
Build the project, upload the form using the demo page, and then preview the form. The submission should load successfully. 

### What's changed
- Removes XML declaration in submission
- Adds scenario test to assert it doesn't fail when the submission has the XML declaration.

### DIFF
--- a/.changeset/dull-papers-yawn.md
+++ b/.changeset/dull-papers-yawn.md
@@ -1,0 +1,6 @@
+---
+'@getodk/xforms-engine': minor
+'@getodk/scenario': patch
+---
+
+Fixes an issue where a submission with an XML declaration causes a crash during edits.


latin-panda:
@lognaturel, when you have time, can you please have a look and share your thoughts?

lognaturel (approved):
Nice to see this getting addressed! The test looks good.

I had hoped it could be ignored either during parsing or while going through the parsed input. Using replace with a regex means iterating over the full instance string and building another string which could be big in the case of submissions to large forms and especially forms with repeats. The fundamental issue is that the declaration is being considered an element node so is being treated as the root, right? Is that because of the wrapping for namespace support? I would have expected the declaration not to be treated as an element node.

I took a quick look around. I see that https://github.com/getodk/web-forms/blob/6ea87574faeb1706f198652076679d74116d69bd/packages/xforms-engine/src/parse/shared/parseStaticDocumentFromDOMSubtree.ts#L54 assumes there's only one top-level element and maybe that's where things go wrong. But again, that doesn't feel like the right place to address the issue, I'd expect the initial parse from string to treat the declaration as something special and not leave it in the representation we work with. I think the perf implications aren't significant enough to spend much more time digging around. I see that the wrapping for default namespace also builds a new string.

Unless the ideas above inspire you to take another look, let's go with the regex approach for now!

Issue #468, opened 2025-08-04
Possible to enter a dot and a comma in an integer widget on a phone


dbemke:
#### Problem description
It's possible to enter a comma or a dot in an integer widget when the submission is made on a phone. To enter the comma and the dot a user needs to tap the button twice. The dot stays in the field while the comma disappears.
The issue was verified on Android 10 (Redmi 9t) and Android 16 (Pixel 7a). English (US) was set in the devices. The public link was used to make the submission.
#### Screenshot

https://github.com/user-attachments/assets/a8787a79-7200-40d1-8b61-854ddcbc4400
#### Steps to reproduce
1. Create a public link of All widgets/All question types form (set WF).
2. Copy the link to a phone.
3. Open the link.
4. Go to integer widgets and try to enter `.` and `,` by double-clicking.

#### Expected behavior
Only integer values should be allowed. 
#### Web Forms version

vb0ae2c0
#### Browser version
Chrome

latin-panda:
Find the implementation notes for the solution in this [PR description](https://github.com/getodk/web-forms/pull/470).

In summary:
> - This was fixed by setting the min and max decimals to zero for the PrimeVue's InputNumber component.
> - Note that the user can still type the comma or dot, but once the field loses focus, the user moves to another question, or taps outside the field, these characters are removed. They are never passed to our components and are discarded by PrimeVue.

srujner:
**Tested with Success!**

Issue #467, opened 2025-08-04
Decimal widget- double-clicking to enter a dot


dbemke:
#### Problem description
A user needs to double-clicking to enter a dot in decimal fields when a submission is made on a phone. Tapping the dot three times removes the dot.
The issue was verified on Android 10 (Redmi 9t) and Android 16 (Pixel 7a) with a public link of All widgets form on Chrome.
#### Screenshot

https://github.com/user-attachments/assets/f38cc70e-e899-46f6-9900-e98897b53970

#### Steps to reproduce
1. Upload All widgets/ All question types form to Central.
2. Create a public link.
3. Copy the link so that you can use it on the phone.
4. In the phone open the link and go to the decimal widget.
5. Try to enter e.g. "2.5".

#### Expected behavior
The dot should appear after tapping once.
#### Web Forms version

vb0ae2c0

#### Browser version
Chrome

Issue #416, opened 2025-05-20
Attempting to type a 10th digit in an integer field turns the number to 9 9s


lognaturel:
Web Forms limits the number of digits to 9 to match Collect. When an additional digit is entered, the field value changes to 999999999.

This was a deliberate decision to go up to the largest number that can be represented with 9 digits but it feels like a bug: https://forum.getodk.org/t/odk-web-forms-beta-integer-question-shows-999999999-after-attempting-to-type-10-digits/54955

Let's match Collect. Collect prevents typing a 10th digit. That behavior is clear to users -- they can tell they've reached some limit.

I think the 999999999 behavior was particularly designed for pasting in values when truncation may not be as obvious. We could maintain that for pasting but I think the truncation is clear enough.

Possible later extensions:
- introduce a system validation message that's shown when a 10th digit is ignored

latin-panda:
Find the implementation notes for the solution in this [PR description](https://github.com/getodk/web-forms/pull/470).

Issue #415, opened 2025-05-20
After entering 16 digits in a string question with numbers appearance, subsequent digits are 0


lognaturel:
There should be no limit to the number of digits entered because it should be stored as a string.

PR #470, opened 2025-08-11
fix: adds input restrictions for numeric fields

latin-panda:
Refactors string with appearance number question type

Closes #

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

### DIFF
--- a/packages/web-forms/src/components/form-elements/input/InputNumeric.vue
+++ b/packages/web-forms/src/components/form-elements/input/InputNumeric.vue
@@ -26,6 +26,7 @@ interface InputNumericProps {
 	readonly isDecimal?: boolean;
 	readonly min?: number;
 	readonly max?: number;
+	readonly maxDigits: number;
 }
 
 const props = defineProps<InputNumericProps>();
@@ -47,7 +48,7 @@ let inputMode: NumericInputMode;
 if (props.isDecimal) {
 	fractionalDigits = {
 		min: 0,
-		max: 18,
+		max: 13,
 	};
 	inputMode = 'decimal';
 } else {
@@ -90,11 +91,11 @@ const modelValue = customRef<number | null>(() => {
 	};
 
 	return {
-		get: () => {
-			return internalValue.value;
-		},
+		get: () => internalValue.value,
 		set: (assignedValue) => {
-			let newValue = assignedValue;
+			const currentValue = internalValue.value;
+			const stringValue = assignedValue != null ? assignedValue.toString() : '';
+			let newValue = stringValue.length <= props.maxDigits ? assignedValue : currentValue;
 
 			if (newValue != null) {
 				const { min = newValue, max = newValue } = props;
@@ -108,9 +109,7 @@ const modelValue = customRef<number | null>(() => {
 				props.setNumericValue(newValue);
 				setValidValue(assignedValue, newValue);
 			} catch {
-				const previousValue = internalValue.value;
-
-				setValidValue(newValue, previousValue);
+				setValidValue(newValue, currentValue);
 			}
 		},
 	};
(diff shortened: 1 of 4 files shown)


packages/web-forms/src/components/form-elements/input/InputNumeric.vue
```diff
@@ -45,56 +52,29 @@ type NumericInputMode = 'decimal' | 'numeric';
...
 if (props.isDecimal) {
-	fractionalDigits = {
-		min: 0,
-		max: 18,
-	};
+	fractionalDigits = { min: 0, max: 13 };
 	inputMode = 'decimal';
 } else {
-	fractionalDigits = {};
+	fractionalDigits = { min: 0, max: 0 };
```
latin-panda, line 58: This resolves https://github.com/getodk/web-forms/issues/468

lognaturel:
It will be very nice to have this behavior aligned!

> Note that the user can still type the comma or dot, but once the field loses focus, the user moves to another question, or taps outside the field, these characters are removed. They are never passed to our components and are discarded by PrimeVue.

This one makes me uneasy. Is it possible to specify an input filter/restriction so that only digits can be typed at all? A user may not notice the change when leaving the field and going from "12.56" to "1256" is not great. I feel like these kinds of manipulations are acceptable when pasting because the user is looking at the field they're pasting into but it feels less ok on focus change.

latin-panda:
>  Is it possible to specify an input filter/restriction so that only digits can be typed at all?

It happens when double-tapping those keys. This is a related issue I haven't been able to fix: https://github.com/getodk/web-forms/issues/467

I want to try upgrading to PrimeVue first; otherwise, I'll attempt to work around it with keypress events.

lognaturel:
Sounds good, we can leave that issue open and dig deeper separately.

srujner:
@latin-panda

**Overall:** 

-Verified All Question Types, Thousands Separator and String Widgets form; 
- Tested if String type has a digit limitation; 
- Tested if Decimal type has a limit to 15 digits
- Tested if Integer type has a limit to 9 digits
- Tested if only numeric values can be entered and non-numeric characters are not accepted
- Tested very large numbers, negative numbers
- Tested with leading zeros to ensure that the field handle it properly (“0” is replaced by the next entered digit) 
- Issues:  https://github.com/getodk/web-forms/issues/263 https://github.com/getodk/web-forms/issues/415  https://github.com/getodk/web-forms/issues/416 https://github.com/getodk/web-forms/issues/468 
Are no longer reproducing
- I didn’t found any UI/UX issues on Iphone

**Iphone:**

- Looks like Holding slider button does not work (On web a can hold it and the number will increase faster, on mobile it is not supported) 
- Keyboard is displayed by default for number fields
- Keyboard on Iphone does not allow non-numeric inputs in
- Rotating the screen, blocking the phone does not introduced any issues


**Other:**

- In String Widget form thousand separator does not work for StringNumberWidget but I’m not sure if this is not a form issue
- On Iphone the slider button stays selected after increasing/decreasing number (it's hardly visible on screenshot) 
IMG_8353
- The screen "jumps" when the limit in the number field is reached and you click on the increase button
https://drive.google.com/file/d/1ExnGtEezIFZR96fDFF4GplnSFsuAHEo2/view?usp=sharing

latin-panda:
@srujner thanks for verifying and providing all those details, it's very useful! 

> In String Widget form thousand separator does not work for StringNumberWidget but I’m not sure if this is not a form issue

I didn't add support for that appearance, I'm working on that right now 🛠️

latin-panda:
@lognaturel @srujner I've added support for the "thousands-sep" in string type with "numbers" appearance.

I've updated the [test server](https://getodk.slack.com/archives/C051L6AKG4F/p1755104131688959) with the latest code.   

This PR is ready.

lognaturel (commented):
Wonderful to have all these addressed! Couple of tiny suggested changes inline.

Did y'all get a chance to try things in a locale where the thousands separator is a space, @srujner? French, for example. Edit: I see evidence of that in the screenshot at https://github.com/getodk/web-forms/pull/470#issuecomment-3188437886 so seems like yes

latin-panda:
> Did y'all get a chance to try things in a locale where the thousands separator is a space, @srujner?

Yes, here is a screenshot:
Screenshot 2025-08-20 at 12 47 08 AM

lognaturel:
Thanks!

Are the e2e tests currently running in CI? I'm not able to quickly confirm that they are.

latin-panda:
> Are the e2e tests currently running in CI? I'm not able to quickly confirm that they are.

Yes, and we added 10 new test cases 🤩, here for example:

Screenshot 2025-08-20 at 1 13 42 AM

PR #472, opened 2025-08-18
fix(#272): round decimals before constructing bigint

### LINKED ISSUE #272: Randomization with seed error: number cannot be converted to BigInt, it's not an integer
#### Web Forms version and package (e.g. are you using `web-forms` or only the engine?)

Web Forms v0.5.0, `main` branch at [4b31432](https://github.com/getodk/web-forms/commit/4b314322c4737d7242dadee3c8c947a206bec774)

#### Browser and version

Chrome v131.0.6778.265

#### Device used

MacOs - Desktop

#### Problem description

The following error pops up when refreshing a freshly uploaded form with a `seed` defined like this `once(substr(decimal-date-time(now()), 8))`.

#### Steps to reproduce the problem

- Upload a form that uses the `randomization` feature with a `seed`. For example: 
[sample form](https://github.com/user-attachments/files/18403273/odk_new_widgets_seed.xlsx)
- Click preview 
- See how form loads without error
- Reload the browser a couple of times
- See popup with error 

Video:
https://github.com/user-attachments/assets/92848dc5-3b6d-4528-aaad-79daf1508ef7

#### Expected behavior
It should not throw an error when reloading the form, and the randomization should work with a seed like `once(substr(decimal-date-time(now()), 8))`.

#### Other information

The same form works fine in the Enketo implementation: https://getodk.org/xlsform/

garethbowen:
Closes #272 

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

Added a unit test and confirmed the test suite passes

I think a better solution would be to avoid the division and multiplication entirely, but because we don't know the exact type of the parameter (just that it's a number) this risks polluting the function with date specific code. One potential benefit of this approach is it protects other code paths too, guaranteeing that the BigInt construction will never throw. One potential downside is there will be a loss of precision, but I think in most cases this loss is the intended behaviour (as in the test case). Another downside is this might silently handle expressions that SHOULD throw, so form builders can find mistakes in their forms, but I think this is unlikely.

I'm not yet familiar with this codebase or its uses so it's possible that there is a regression on a particular expression, but I think this change only impacts forms that previously would have thrown an exception.

The form attached to #272 is sufficient to test this.

### What's changed

Instead of passing a "number" to the BigInt constructor this change rounds it to ensure the constructor is only ever called with an integer.

### DIFF
--- a/packages/xpath/test/xforms/decimal-date-time.test.ts
+++ b/packages/xpath/test/xforms/decimal-date-time.test.ts
@@ -55,7 +55,8 @@ describe('#decimal-date-time()', () => {
 
 	// Tests this bug: https://github.com/getodk/web-forms/issues/272
 	it('should handle milliseconds losing precision during conversion', () => {
-		testContext.assertNumberValue('decimal-date-time(date("2025-08-18T02:03:23.958Z"))', 20318.085693958332);
+		const expression = 'decimal-date-time(date("2025-08-18T02:03:23.958Z"))';
+		testContext.assertNumberValue(expression, 20318.085693958332);
 	});
 
 	it.fails('with invalid args, throws an error', () => {


garethbowen:
Thanks! Still finding my way around :)

Can you approve the run, please?

PR #473, opened 2025-08-18
Handle http status codes

garethbowen:
Closes #

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

I'm not sure about outputting the error code but because it's a demo app I'm assuming it's for more technical users who would appreciate this.

None. Provides more info about errors.

None.

### What's changed

Before this change if the request to the XLSForm Online service failed the status code was ignored and the server attempted to parse the JSON response which doesn't work. This shows a misleading error message which implied (to me) that there was something wrong with the input rather than checking that the service was running.

Before:

Screenshot From 2025-08-19 10-51-57

After:

Screenshot From 2025-08-19 10-55-26

I also changed the code from a mix of async/await and promises, to be completely async/await which I think is easier to read, but I did this in a separate commit so I can revert that easily if you just want the bug fix and nothing else.

### DIFF
--- a/packages/web-forms/src/demo/composables/xlsform-online.ts
+++ b/packages/web-forms/src/demo/composables/xlsform-online.ts
@@ -28,11 +28,11 @@ export const useXlsFormOnline = (baseUrl: Ref<string | undefined>) => {
 				body: formData,
 			});
 			if (res.ok) {
-				data = await res.json();
+				data = (await res.json()) as XlsOnlineResponse;
 			} else {
 				error = `Error connecting to xlsform-online server, status: ${res.status}. If problem persists, please report it on ODK Forum.`;
 			}
-		} catch(fetchError: unknown) {
+		} catch (fetchError: unknown) {
 			error = String(fetchError);
 		}
 


latin-panda (commented):
This looks great, thanks! The demo is automatically deployed at https://getodk.org/web-forms-preview/ once the changes are merged.

There was a bug in the CI that prevented it from running for pull requests from forks. Could you rebase to include the latest fix? Then we will see the CI checks passing, and I'll approve the PR.

By the way, how did you encounter that error? I attempted to stop the XLSForm Online service, disable the network, and restart it with incorrect configurations, but it always ends up in the catch block.

garethbowen:
I've rebased - awaiting approval.

I got the error while I was figuring out how to get CORS working. In one misguided attempt I changed the port that the demo was running on to 8000 which was the port that the xlsform service was configured to run on. Because the ports were the same the CORS issue went away, but the request went to the demo which didn't have the requested endpoint and returned a 404. While this is probably an edge case not worth worrying about, I expect it would be possible to get the same result with any response from the xlsform service with invalid JSON (eg: empty response).

You will still end up in the catch block because the `res.json()` call throws an Error, the problem is the error message isn't very helpful (see the before screenshot).

latin-panda (approved):
Thank you! ✨

There are a couple of lint errors. The format error is usually fixed by running `yarn format` in the root folder. Since this project is quite strict with types, we might want to define something to describe the returned object. That would make it ready to squash and merge.

garethbowen:
Same here - should pass now!

PR #478, opened 2025-08-25
feat(#305):Select one for map

### LINKED ISSUE #305: Add select one from map question type
User-facing documentation: https://docs.getodk.org/form-question-types/#select-one-from-map-widget

latin-panda:
Closes #305 

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable


packages/xforms-engine/src/lib/reactivity/createItemCollection.ts
```diff
@@ -20,8 +18,11 @@ import type { ReactiveScope } from './scope.ts';
 import { createTextRange } from './text/createTextRange.ts';
 
 export type ItemCollectionControl = RankControl | SelectControl;
-type Item = RankItem | SelectItem;
 type DerivedItemLabel = ClientTextRange<'item-label', 'form-derived'>;
+export interface BaseItem {
```
latin-panda, line 22: Start in this file

This is the simple base item for Rank and Select. 

Select's item will include `geometry` and `metadata`. The latter contains everything defined in the `choice` sheet from an XLS form, except for the value's column (usually called "name", but it can be another column), itextId, and geometry, to prevent duplicate data in Select's item.

packages/xforms-engine/src/lib/reactivity/createItemCollection.ts
```diff
@@ -101,6 +102,8 @@ const createItemsetItemLabel = (
 interface ItemsetItem {
```
latin-panda, line 102: If rank and `<item>` require geometry and metadata, then we can use only this interface and remove BaseItem.

I think `<item>` needs geometry and metadata, but rank does not.

packages/xforms-engine/src/lib/reactivity/createItemCollection.ts
```diff
@@ -120,11 +123,30 @@ const createItemsetItems = (
...
 						defaultValue: '',
 					});
+
 					const label = createItemsetItemLabel(context, itemset, value);
 
+					const geometry = createComputedExpression(context, itemset.geometry, {
+						defaultValue: '',
+					});
+
+					const nodeElements = itemNode
```
latin-panda, line 133: The itemNode already includes everything from the `choice` sheet. Here, we follow the same approach as with "value" (L123).

packages/xforms-engine/src/parse/body/control/ItemsetDefinition.ts
```diff
@@ -49,7 +54,19 @@ export class ItemsetDefinition extends BodyElementDefinition<'itemset'> {
 			throw new Error(`<itemset> has no <value>`);
 		}
 
-		this.value = new ItemsetValueExpression(this, valueExpression);
 		this.label = ItemsetLabelDefinition.from(form, this);
+		this.value = new ItemsetValueExpression(this, valueExpression);
+		this.geometry = new ItemsetGeometryExpression(this);
+
+		this.metadataExclusions = ['itextId', this.value.expression, this.geometry.expression];
```
latin-panda, line 61: As noted earlier, I aim to exclude anything already exposed as a class attribute to prevent data duplication and simplify the Vue client’s looping and rendering work.

packages/xforms-engine/src/client/SelectNode.ts
```diff
@@ -6,14 +7,13 @@ import type { LeafNodeDefinition } from '../parse/model/LeafNodeDefinition.ts';
...
 import type { ValueType } from './ValueType.ts';
 import type { GeneralParentNode } from './hierarchy.ts';
 import type { LeafNodeValidationState } from './validation.ts';
 
-export interface SelectItem {
-	get label(): TextRange<'item-label'>;
-	get value(): string;
+export interface SelectItem extends BaseItem {
+	geometry?(): string;
+	metadata?: Array<{ label: string; value(): string }>;
```
lognaturel, line 16: `properties` or `additionalProperties`?

`geometry` could also simply be an `additionalProperty`.

But that makes me think -- in what format is the `geometry` value right now when items are from a geoJSON file? Ideally the engine exposes a consistent format so the client doesn't have to do any conversion.

If `geometry` is made into its own concept, we should use a type to describe/validate/represent the structure rather than a raw string.

packages/xforms-engine/src/lib/reactivity/createItemCollection.ts
```diff
@@ -120,11 +123,30 @@ const createItemsetItems = (
...
+						.getXPathChildNodes()
+						.filter((node) => node.nodeType === 'static-element');
+					const metadata = itemset.getMetadataExpressions(nodeElements).map((meta) => {
+						return {
+							label: meta.elementName,
```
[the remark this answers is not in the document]
lognaturel: These are the values of additional properties, right? Currently these are not translatable -- each property/column is independent from the others.

lognaturel (commented):
I haven't had a chance to think about this in detail quite yet. I think the key question is what to expose to the client. Some of the big questions I see:
* Is `geometry` a first-class concept? In JavaRosa/Collect, it's not something JR knows about, it's just passed as part of an "additional properties" list and Collect pulls it out.
* What format should the geometry value be exposed in? In Collect/JR, it's always in ODK format, even when from geoJSON

latin-panda:
Closing since we're going to implement another solution for this feature.

PR #487, opened 2025-09-01
feat(#22): support output in translation text

### LINKED ISSUE #22: Allow output in translation text
- Should render `<output>` value correctly   
- Dynamic repeat labels with `iText` and `<output>` should work, see related comment: https://github.com/getodk/web-forms/issues/101#issuecomment-3132702638

### LINKED ISSUE #418: Support `output` in validation messages
`MessageDefinition` does not identify and process `output`s.

Test form: [ref_in_message.xlsx](https://github.com/user-attachments/files/20380293/ref_in_message.xlsx)

garethbowen:
Closes #22
Closes #418

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

An additional scenario test has been enabled and passes.

I tested multiple forms with translation text in repeats, labels, constraints, required, and value.

I tried updating the value, which worked but it's designed to be immutable so may have had unintended consequences. It also required reducing the results that could be cached which would have been worse for performance than this solution.

I also looked at updating the xpath package instead but that was spreading the translation code across packages. This way the code is co-located in the engine.

The change may have affected images in select options.

Form which includes output in constraint and required message: 
[ref_in_message.xlsx](https://github.com/user-attachments/files/22069748/ref_in_message.xlsx)

### What's changed

Now when building the text instead of just assuming it's a single element and evaluating it, this change checks for children and evaluates them as well, with the result as the concatenation of the strings of all children in order.

### DIFF
--- a/.changeset/old-shoes-wonder.md
+++ b/.changeset/old-shoes-wonder.md
@@ -1,0 +1,5 @@
+---
+'@getodk/xforms-engine': minor
+---
+
+Support output in translation text
--- a/packages/scenario/test/repeat-output.test.ts
+++ b/packages/scenario/test/repeat-output.test.ts
@@ -316,72 +316,69 @@ describe('Interaction between `<repeat>` and `<output>`', () => {
 					).toBe('Position: 2');
 				});
 
-				it.fails(
-					'produces the output of an expression with a relative reference (alternate #2)',
-					async () => {
-						const scenario = await Scenario.init(
-							'<output> with relative ref in translation',
-							html(
-								head(
-									title('output with relative ref in translation'),
-									model(
+				it('produces the output of an expression with a relative reference (alternate #2)', async () => {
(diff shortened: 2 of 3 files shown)


latin-panda (changes requested):
Looks good! I also verified with some forms that contain select with images and other dynamic text. 

I found [a couple of output tests](https://github.com/getodk/web-forms/blob/6015ab6b3cb9ccc5b79ad4a453e95c6cae6c79f3/packages/scenario/test/label-hint-text.test.ts#L86) that could be enabled. What do you think?

garethbowen:
@latin-panda One more question I had about this... this solution only supports one level of children, (ie: it's not recursive). I think that's ok - I can't see how you could nest more than that in an itext. What do you think?

Issue #459, opened 2025-07-23
Upgrade tree-sitter dependencies


latin-panda:
## Description
Upgrade all dependencies related to `tree-sitter` in the project to their latest compatible versions. Ensure the following requirements are met:

1. Address any conflicts that arise during the upgrade process.
2. Update the [expected version variable](https://github.com/getodk/web-forms/blob/b0ae2c0f98e4ca7cdba352b98e47fdcd4720ed81/packages/tree-sitter-xpath/scripts/build/pre.mjs#L17) to reflect the new `tree-sitter` version.
3. Ensure the bundle size does not increase, as `tree-sitter` already contributes significantly to the bundle size.
4. Verify that no new unnecessary build artifacts are introduced (reference: [build script](https://github.com/getodk/web-forms/blob/b0ae2c0f98e4ca7cdba352b98e47fdcd4720ed81/packages/tree-sitter-xpath/scripts/build/pre.mjs#L7)).
5. Confirm that all scenario tests pass successfully.
6. Verify that performance is unchanged or improved compared to the current implementation.

<img width="1551" height="276" alt="Image" src="https://github.com/user-attachments/assets/6526f069-c793-4009-8cd8-58afe4adcad0" />

Issue #8, opened 2023-10-25
Large bundle size (already!)


eyelidlessness:
This is something I’ve intentionally deferred because it intersects with tooling which has already eaten up a lot of time and energy. But it should be tracked, and some initial observations are worth dropping here for a starting point.

- The `crypto-js` library is quite large, and probably not the ideal solution for `digest` anyway. Any solution should likely also be optional, as it is in ORXE. I’ve looked at other options, but that’s best for another issue.
- The `Temporal` polyfill is gargantuan. It’s also poorly set up for bundle size optimization like tree shaking (this can be improved by importing the parts in use directly from source, but it’s worth reconsidering the polyfill in general depending on the standard’s progress)
- The `web-tree-sitter` WASM size is large-ish, and `tree-sitter-xpath` isn’t tiny either. These would benefit from aggressive caching.
- In general the best tooling-side approach to bundle size optimization is an open question. Besides deferring for pragmatic progress reasons, it also makes sense to hold until other stories are more clear (like offline, PWA if we pursue that, etc)

latin-panda:
I spent some time trying a few things. Moving demo assets out of core components provided the only significant reduction to the main bundle size.

| Metric | Initial Build | Final Build (After Asset Move) |
| :--- | :--- | :--- |
| **Main Bundle (OdkWebForm)** | 3,612.88 kB | **2,205.49 kB** |
| **Gzip Size** | 1,532.40 kB | **524.93 kB** |

The remaining bundle size is primarily driven by the engine and UI dependencies:

* **xforms-engine:** 1.46 MB. This contains the core logic and is the expected for this package (it has tree-sitter and other dependencies).
* **UI Library (node_modules):** ~1.02 MB. This is dominated by PrimeVue dependencies, including @primeuix, @primevue/core, and themes.

Some additional notes from my investigation:

* For some reason, the CSS is currently being injected into the JS files. Extracting CSS into a separate index.css (1,283.71 kB) reduced the index.js chunk size but failed to significantly improve the size of the main OdkWebForm chunk.
* Trying to use Vite chunking for the demo page did not have significant improvements for the main bundle.

By physically moving the demo images to the demo page directory, the main bundle size was reduced by approximately 1.4 MB ([Pull Request](https://github.com/getodk/web-forms/pull/665))

PR #488, opened 2025-09-01
feat(#242): add pulldata function

### LINKED ISSUE #242: Add the pulldata function
Here is a user interested in using it: https://forum.getodk.org/t/select-multiple-choices-then-rank-the-selected-choices/16444/16?u=ln

garethbowen:
Closes #242

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

Ported scenario tests from javarosa.

This matches the other similar function implementations.

The changes are be isolated to the previously unsupported function call so it should not impact any existing forms.

This is the one I used... 
[242-pulldata.xml](https://github.com/user-attachments/files/22070379/242-pulldata.xml)

Typing in "south" returns "Texas" which is the first result that matches.
Typing in "north" returns "Alaska" which is the only result that matches.
Typing in anything else will return empty string.

### What's changed

A brand new function `pulldata` has been implemented compliant with the ODK spec: https://getodk.github.io/xforms-spec/#fn:pulldata


packages/scenario/test/xpath/functions/pulldata.test.ts
```diff
@@ -0,0 +1,78 @@
...
+			{
+				testName: 'returns empty string when uses not existing property',
+				property: '@doesnotexist',
+				inputValue: 'nothing',
+				expectedOutput: '',
+			},
+			{
+				testName: 'returns empty string when uses not existing property with no value',
+				property: '@doesnotexist',
+				inputValue: '',
```
latin-panda, line 47: Is it worth adding a couple of cases where the `inputValue` has single quotes or double quotes? 
For example: `O'Reilly` and `Hello "world"`
Maybe related post: https://forum.getodk.org/t/different-behaviour-between-enketo-and-odk-collect-with-straight-quotes-in-select-from-file-questions/49973/7

latin-panda: Tricky! I don't have experience with the Tree Sitter code, and I share your concern.

I'm unsure if any of the Tree Sitter is autogenerated or manually ported ([commit](https://github.com/getodk/web-forms/commit/5dc146e7df4a7ddeb9bd3847e3b1af8fd8f7f2ff)). I want to dig more into the history tomorrow (docs, slack, etc)

>It's possible I've now enabled forms to have escaped quotes in all sorts of places, or it's possible I've broken something somewhere... 

+1. How can we increase confidence? We have a solid set of scenario tests. Do you think it's worth adding more where things might break, like calculations? 

-----

Side note:


Tree Sitter is something I've been curious about. I'm wondering if it's really the best tool for us, but we haven't had the time to reevaluate it ([ticket](https://github.com/getodk/web-forms/issues/41)) (for example: Is there a smaller library available ([bundle size ticket](https://github.com/getodk/web-forms/issues/8)) that doesn't require the Docker daemon to run and is easier to upgrade ([ticket](https://github.com/getodk/web-forms/issues/459))?)

garethbowen: Interesting. So openrosa-xpath-evaluator doesn't appear to handle any escaping either which made me wonder why web-forms/xpath had to. I think the cause is that the openrosa-xpath-evaluator implementation doesn't strip the quotes out, so the form builder had to pick the right quotes to not have an escaping issue.

Note the difference between:

https://github.com/enketo/enketo/blob/5ce794321a85d1c0ffb90284581b2edffa8b3933/packages/enketo-core/src/js/form-model.js#L1405

and the version in this PR:

```
const expr = `instance('${instanceId}')/root/item[${queryElement}='${escapedQuery}']/${desiredElement}`;
```

I think this is because of the way the string literals are passes. In openrosa-xpath-evaluator the parameters still have the quotes, and must be manually stripped, whereas in web-forms the quotes are automatically stripped:

https://github.com/getodk/web-forms/blob/6015ab6b3cb9ccc5b79ad4a453e95c6cae6c79f3/packages/xpath/src/evaluator/expression/StringLiteralExpressionEvaluator.ts#L11

Therefore in my implementation I add them back, but I don't know whether to use `'` or `"` so sometimes it's wrong.

Options:
1. Leave it the way it is in this PR, which may have unintended consequences, both good and bad.
2. Update the StringLiteralExpressionEvaluator to not strip quotes, or at least expose the original string, so I can use that (compatible with openrosa-xpath-evaluator).
3. (mentioned above) Use double quotes if the value has single quotes, and vice versa. However this wouldn't work if a string had both.

2 and 3 would not require any changes to webforms/tree-sitter-xpath which feels safer, and more compliant with openrosa.

latin-panda: I'm more inclined to Option 2. It is slightly faster due to less string manipulation and a lower risk of breaking something. We can document that quote-handling behavior in the pulldata documentation.

lognaturel: Good call on thinking about quotes, @latin-panda. I think there are two cases to consider here:
* `query` is a literal value with a quote and/or apostrophe -- this is likely to be quite rare
* `query` is a reference to another field (often a select) and that field can contain a value with a quote and/or apostrophe -- this is somewhat likely and is what [the forum post](https://forum.getodk.org/t/different-behaviour-between-enketo-and-odk-collect-with-straight-quotes-in-select-from-file-questions/49973) illustrates

I'm not totally sure whether these would be handled differently but want to point out that the second one is most important to support.

I agree with the discomfort with the current approach. 3 feels too hacky to me. I'm having trouble with thinking through the implications for 2. Would it also have the problem that Enketo seems to have as highlighted by the forum post linked above?

If so, my preference would be to remove anything special for apostrophe/quote handling for now and file it as a separate issue since it is an edge case.

lognaturel: I do think the literal vs reference cases are different, thinking about it more. In the literal case, the form designer can wrap the value with quotes but in the reference case the values are whatever is in that field. Though I'm not sure what the escaping story is there currently!

lognaturel: Another thing to note is that more typical `pulldata` usage would be pulling from a CSV rather than from an internal instance. I also don't think `pulldata` can get attribute values in Collect nor Enketo so filtering by the `value` child would be more realistic. I also imagine values with quotes/apostrophe from CSV might have special handling so it would be good to have a scenario test on that end-to-end pattern. I'm more and more inclined to do quote/apostrophe handling as its own mini project.

garethbowen: Okie dokie...

Upon another look at the Enketo code, it's not as clever as I suggested it was - it literally wraps the evaluation of the query with single quotes just like my naive first attempt did. So, I've removed the escaping code and raised #492 to cover it more thoroughly, and commented out the scenario test that was asserting quotes worked.

I'm not terribly happy with where this ended up but I think it's completely compatible with Enketo. I'll leave it up to those with more experience to decide whether we want to do better than Enketo in the first pass.

> filtering by the value child would be more realistic

Thanks - I've added a test for `value` so that's covered, not just attributes.

garethbowen:
@latin-panda I think this is ready for a review again. It still doesn't handle apostrophes which is covered by #492

packages/scenario/test/xpath/functions/pulldata.test.ts
lognaturel: I’m happy to match Enketo for now and have the issue but @latin-panda you can have the final say!

PR #496, opened 2025-09-19
feat(#117): add support for jr:choice-name function

### LINKED ISSUE #117: Add support for jr:choice-name
Plus test that is a recent addition to JR with a slight behavior change: https://github.com/getodk/javarosa/pull/758

garethbowen:
Closes #117 

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

Scenario tests enabled and passed.

Discussed in the issue.

New feature so should not impact users.

The [choice-name fixture](https://github.com/getodk/web-forms/blob/main/packages/common/src/fixtures/test-javarosa/resources/org/javarosa/core/model/jr-choice-name.xml) form tests a range of different implementations.

### DIFF
--- a/packages/common/src/fixtures/test-javarosa/resources/org/javarosa/core/model/jr-choice-name.xml
+++ b/packages/common/src/fixtures/test-javarosa/resources/org/javarosa/core/model/jr-choice-name.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <!-- see binds for explanations of each case -->
-<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa">
+<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:odk="http://www.opendatakit.org/xforms/">
     <h:head>
         <h:title>jr-choice-name</h:title>
         <model>
@@ -128,6 +128,9 @@
                         <select_one_name/>
                     </my-repeat>
 
+                    <rank/>
+                    <rank_name/>
+
                     <meta>
                         <instanceID/>
                     </meta>
@@ -192,6 +195,10 @@
             <bind nodeset="/jr-choice-name/my-repeat/select_one" type="string"/>
             <bind calculate="jr:choice-name( ../select_one , ' ../select_one ')" nodeset="/jr-choice-name/my-repeat/select_one_name" type="string"/>
 
+            <!-- works with ranks too -->
+            <bind nodeset="/jr-choice-name/rank" type="rank"/>
+            <bind calculate="jr:choice-name( selected-at(/jr-choice-name/rank, 0) , ' /jr-choice-name/rank ')" nodeset="/jr-choice-name/rank_name" type="string"/>
+
             <bind jr:preload="uid" nodeset="/jr-choice-name/meta/instanceID" readonly="true()" type="string"/>
         </model>
     </h:head>
@@ -211,6 +218,10 @@
 
         <input ref="/jr-choice-name/my-repeat/select_one_name">
             <label>both parameters to jr:choice-name can be relative XPath paths</label>
+        </input>
+
+        <input ref="/jr-choice-name/rank_name">
+            <label>works with ranks too</label>
         </input>
 
         <select1 ref="/jr-choice-name/select_one_outside">
@@ -279,5 +290,13 @@
                 </select1>
             </repeat>
         </group>
+
+        <odk:rank ref="/jr-choice-name/rank">
+            <label>Rank</label>
+            <itemset nodeset="instance('cocoteros')/root/item">
+                <value ref="name"/>
+                <label ref="label"/>
+            </itemset>
+        </odk:rank>
     </h:body>
 </h:html>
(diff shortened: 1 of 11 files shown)


packages/scenario/test/choice-name.test.ts
```diff
@@ -38,10 +39,13 @@ describe('JavaRosa ports: ChoiceNameTest.java', () => {
 	});
 
 	describe('choice name call in repeat with static choices', () => {
-		it.fails('gets choice name', async () => {
+		it('gets choice name', async () => {
 			const scenario = await Scenario.init(r('jr-choice-name.xml'));
+
 			scenario.answer('/jr-choice-name/my-repeat[1]/select_one', 'choice4');
+			scenario.createNewRepeat('/jr-choice-name/my-repeat');
```
garethbowen, line 46: I needed to add this to get the tests to pass, but I think it's correct. Before this it was trying to update the second repeat item but there was only one child.

garethbowen:
It would be good to test this with output in the translation text when this is resolved: https://github.com/getodk/web-forms/issues/22

packages/xpath/src/functions/javarosa/select.ts
```diff
@@ -0,0 +1,43 @@
...
+		}
+
+		const firstNode = nodes?.[0];
+		if (!firstNode) {
+			return '';
+		}
+		if (!('getChoiceName' in firstNode)) {
+			throw new Error(`Evaluating 'jr:choice-name' on element without choices`);
+		}
+		return (firstNode as XPathChoiceNode).getChoiceName(node) ?? '';
```
latin-panda, line 41: So, this is the one defined in `packages/xforms-engine/src/instance/SelectControl.ts`, right? And, it's not exposed to the Vue front-end because it's not specified in `packages/xforms-engine/src/client/SelectNode.ts`. Then it should be fine; we want to expose only the necessary. 

It will require a bit of detective work for anyone new to understand why SelectControl.ts has a `getChoiceName` method or why they should implement `getChoiceName` in a control to enable `jr:choice-name` in a question type.

We can clarify it further, for instance, by expanding the exception message from L39 (`Evaluating 'jr:choice-name' on element without choices. Check element's control definition.`) or by adding code comments that connect both things.

latin-panda (approved):
Looks great, I left some minor comments for consideration. I looked for common issues in the forum and tested those; it works well.

packages/common/src/fixtures/test-javarosa/resources/org/javarosa/core/model/jr-choice-name.xml
```diff
@@ -196,6 +196,23 @@
         </model>
     </h:head>
     <h:body>
+
+        <input ref="/jr-choice-name/literal_choice_name">
```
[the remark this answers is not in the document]
garethbowen: Good idea. Added now.

packages/xforms-engine/src/instance/SelectControl.ts
```diff
@@ -231,4 +232,9 @@ export class SelectControl
 
 		return this.root;
 	}
+
+	getChoiceName(value: string): string | null {
```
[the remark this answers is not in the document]
garethbowen: Yes it does.

packages/xpath/src/functions/javarosa/select.ts
garethbowen: > So, this is the one defined in packages/xforms-engine/src/instance/SelectControl.ts, right?

It's implemented there, if that's what you mean? The interface definition is in `xpath` so we can use it, it's implemented in the `xforms-engine`, and you're right, the `web-forms` package has no idea it exists.

If anyone else wants to use the `xpath` package as a standalone lib they can choose to implement this interface or not, but it's completely hidden to anyone depending on the `xforms-engine` package. It seems unlikely that anyone would want to use the `xpath` package only but maybe they will one day?

> We can clarify it further, for instance, by expanding the exception message from L39

Yes. It's tricky because it could either be shown to a developer implementing their own controls, or by someone building a form and applying the function to an invalid element. Either way the instructions are different. I think the compromise is to tailor the Error message for the form builder, and include a code comment for the developer. I've updated the documentation to call this out.

latin-panda: Sounds good!

latin-panda:
@garethbowen thanks! It's ready to merge

packages/scenario/test/choice-name.test.ts
```diff
@@ -74,14 +78,11 @@ describe('JavaRosa ports: ChoiceNameTest.java', () => {
 			// TODO (web-forms): per @lognaturel,
 			// https://github.com/getodk/javarosa/issues/737 is pertinent.
 			//
-			// TODO (JR): why does test fail if value is not set to choice3 again?
-			// Does changing language not trigger recomputation?
```
lognaturel, line 78: Changing language does not trigger recomputation in JR, indeed.

packages/xpath/src/functions/javarosa/select.ts
```diff
@@ -0,0 +1,43 @@
...
+			nodes = context.evaluator.evaluateNodes(value);
+		}
+
+		const firstNode = nodes?.[0];
+		if (!firstNode) {
+			return '';
+		}
+		if (!('getChoiceName' in firstNode)) {
+			throw new Error(`Evaluating 'jr:choice-name' on element without choices`);
+		}
```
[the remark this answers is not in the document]
lognaturel: We haven’t historically been very strict about process for follow-up work like this. I think a PR with a good description can stand on its own.

PR #499, opened 2025-09-22
fix(#479): render groups with no ref attribute

### LINKED ISSUE #479: Groups without `ref` are not rendering
### Description
Groups without `ref` are not rendering in Web Forms, it's being interpreted as a note type `subtree`

Image

After adding the `ref` it works fine (from: `<group appearance="field-list">` To: `<group appearance="field-list" ref="/data/group">`)

Fix the xforms-engine (or xpath) to resolve the form correctly. 

<details><summary>XML Form</summary>

```xml
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa">
  <h:head>
    <h:title>requiredOnePage</h:title>
    <model>
      <instance>
        <data id="build_requiredOnePage_1518518934">
          <meta>
            <instanceID/>
          </meta>
          <group>
            <text/>
            <number/>
            <date/>
            <time/>
            <location/>
            <barcode/>
          </group>
        </data>
      </instance>
      <itext>
        <translation lang="English">
          <text id="/data/group:label">
            <value>group</value>
          </text>
          <text id="/data/group/text:label">
            <value>text</value>
          </text>
          <text id="/data/group/number:label">
            <value>number</value>
          </text>
          <text id="/data/group/number:hint">
            <value>hint number</value>
          </text>
          <text id="/data/group/time:hint">
            <value>hint time</value>
          </text>
          <text id="/data/group/location:requiredMsg">
            <value>requiredLocation</value>
          </text>
          <text id="/data/group/barcode:label">
            <value>barcode</value>
          </text>
          <text id="/data/group/barcode:requiredMsg">
            <value>requiredBarcode</value>
          </text>
        </translation>
      </itext>
      <bind nodeset="/data/meta/instanceID" type="string" readonly="true()" calculate="concat('uuid:', uuid())"/>
      <bind nodeset="/data/group/text" type="string" required="true()"/>
      <bind nodeset="/data/group/number" type="int" required="true()"/>
      <bind nodeset="/data/group/date" type="date" required="true()"/>
      <bind nodeset="/data/group/time" type="time" required="true()"/>
      <bind nodeset="/data/group/location" type="geopoint" required="true()" jr:requiredMsg="jr:itext('/data/group/location:requiredMsg')"/>
      <bind nodeset="/data/group/barcode" type="barcode" required="true()" jr:requiredMsg="jr:itext('/data/group/barcode:requiredMsg')"/>
    </model>
  </h:head>
  <h:body>
    <group appearance="field-list">
      <label ref="jr:itext('/data/group:label')"/>
      <input ref="/data/group/text">
        <label ref="jr:itext('/data/group/text:label')"/>
      </input>
      <input ref="/data/group/number">
        <label ref="jr:itext('/data/group/number:label')"/>
        <hint ref="jr:itext('/data/group/number:hint')"/>
      </input>
      <input ref="/data/group/date">
      </input>
      <input ref="/data/group/time">
        <hint ref="jr:itext('/data/group/time:hint')"/>
      </input>
      <input ref="/data/group/location">
      </input>
      <input ref="/data/group/barcode">
        <label ref="jr:itext('/data/group/barcode:label')"/>
      </input>
    </group>
  </h:body>
</h:html>


```

</details>

### Related Issues

- Spec: https://getodk.github.io/xforms-spec/#groups

garethbowen:
Closes #479

Also includes a significant refactor to reduce the Group types from 3 (structural, presentation, logical) to 1. There was some inconsistency here with a) naming internally, and b) the spec. Another option would have been to retain two types (presentation and non-presentation) but as this is not useful right now I decided to go back to 1 type and we can add subtypes back when it becomes an issue.

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

Using the form provided in the issue.

1. Making a simpler change which just checked for either Group or Subtree, but in the end the flattening was preferred to clean up the inconsistencies.

This is a breaking change in terms of the engine API, but front end users will not notice.

The form attached in the issue is useful.

### What's changed

Group types have been reconciled so the web-forms UI handles groups with no ref in the same way as groups with a ref.


packages/web-forms/src/components/form-layout/FormGroup.vue
```diff
@@ -1,10 +1,10 @@
 <script setup lang="ts">
-import type { GroupNode } from '@getodk/xforms-engine';
+import type { GroupNode, SubtreeNode } from '@getodk/xforms-engine';
 import { computed } from 'vue';
 import FormPanel from './FormPanel.vue';
 import QuestionList from './QuestionList.vue';
 
-const props = defineProps<{ node: GroupNode }>();
+const props = defineProps<{ node: GroupNode | SubtreeNode }>();
```
latin-panda, line 7: To cross-post the conversation from before. 

A Subtree always appears as either a logical group or a presentation group. It might be possible to simplify this by having a single Node representation for both, allowing the client to handle small rendering differences (such as when appearances is null) within the Group Vue component.

It was a bit confusing at first: What is a Subtree? What else can be a Subtree? Why are repeats not Subtree? How does "Subtree" align with the terminology found in ODK's spec documentation?
It took longer to feel confident about my interpretation being correct, and confidence about side effects if I were to modify SubTree in the future.

latin-panda (commented):
I tested other presentation and logical group forms, as well as repeats. It works fine! 

I left a suggestion below. Let me know what you think!

packages/common/src/fixtures/groups/presentation-group.xml
```diff
@@ -0,0 +1,79 @@
...
+      <bind nodeset="/data/group/text" type="string" required="true()"/>
+      <bind nodeset="/data/group/number" type="int" required="true()"/>
+      <bind nodeset="/data/group/date" type="date" required="true()"/>
+      <bind nodeset="/data/group/time" type="time" required="true()"/>
+      <bind nodeset="/data/group/location" type="geopoint" required="true()" jr:requiredMsg="jr:itext('/data/group/location:requiredMsg')"/>
+      <bind nodeset="/data/group/barcode" type="barcode" required="true()" jr:requiredMsg="jr:itext('/data/group/barcode:requiredMsg')"/>
+    </model>
+  </h:head>
+  <h:body>
+    <group appearance="field-list">
```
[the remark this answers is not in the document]
latin-panda: Good question, yes, leaving just one should be fine. To render one means the FormQuestion component has been reached, and then any other question type can render as well.

packages/web-forms/src/components/form-layout/FormGroup.vue
garethbowen: Phew... well... this thread just keeps coming. Firstly some useful comments in the code:

https://github.com/getodk/web-forms/blob/30997b55d582f41945b90c214d6ddb770f440b20/packages/xforms-engine/src/parse/body/group/BaseGroupDefinition.ts#L14-L37

There are various comments sprinkled through the code like:

> // TODO: obviously there is a naming inconsistency emerging here.

1. I don't see any value in the "subtree" nomenclature. It's not used anywhere else that I can see. Can we find and replace "subtree" -> "group"?
2. web-forms talks about logical, presentation, and structural groups as being distinct, but the spec doesn't mention structural groups at all. The only difference between structural and presentation is the latter includes a label. I don't think this distinction is useful in web forms and we should reduce the choices to "logical" and "presentation" groups, where presentation groups may or may not have a label.
3. I do think there's value in the type system reinforcing the differences between logical and presentation groups. However it would be useful to have views deal with any group where the logic is the same (or similar). I propose introducing a parent type for the client api, ie: LogicalGroupNode and PresentationGroupNode would both extend  GroupNode. We should also update the internal engine types at the same time to be consistent with this naming.

garethbowen: I don't think there's much difference in the UI aside from appearances. A few notes from [the spec](https://getodk.github.io/xforms-spec/#groups)...

- A logical group’s ref is used as the context node for the relative ref paths of its descendants.
- a logical group can also contain a relevant attribute on its <bind> element

But that's about it.

garethbowen: A summary of our conclusion from the call just now...

1. Replace the term Subtree with Group as it's more obvious what it is.
2. Define types for PresentationGroup (a group which has a label child element) and NonPresentationGroup (is there a better name for this?) because there _should_ be a visual distinction between the two.
3. Make no type distinction for groups with or without a ref - the latter Node will have `appearances = null`.

@lognaturel @latin-panda Is this what you took out of the meeting too?

latin-panda: Sounds good to me

Naming options for NonPresentationGroup:
- StructuralGroup: seemed like a natural choice to me
- LogicalGroup: aligns with specs — initially, someone might question what "logical" means here.

garethbowen:
@latin-panda As discussed I've refactored the *Group to greatly simplify it and removing naming inconsistencies. After some investigation it was clear there is very little distinction between any of the three group types either in the UI or the engine so I decided to flatten down to just one. I removed most of the references to "subtree" where "group" was more descriptive but left references where it's referring to a subset of the document.

Let me know what you think!

packages/web-forms/src/components/form-layout/FormGroup.vue
lognaturel: A logical group is any group that has a `ref` attribute so it would not be accurate to use here! Super confusing, I know. `NonPresentationGroup` seems like the most clear to me.

lognaturel:
Haven't looked in detail but the simplification sounds good to me. I would expect tests at https://github.com/getodk/web-forms/blob/6cfff8b4c5a2cf6a23a71ef6d4308343bccd2436/packages/odk-web-forms/src/lib/xform/model/ModelDefinition.test.ts#L479 and https://github.com/getodk/web-forms/blob/6cfff8b4c5a2cf6a23a71ef6d4308343bccd2436/packages/odk-web-forms/src/lib/xform/model/ModelDefinition.test.ts#L576 to be addressed in some way, either in this PR or in follow-up. I haven't dug deeply but I think the first can be modified to pass and the second can be removed because there isn't a need for the structural group concept I think.

.changeset/great-deer-march.md
```diff
@@ -0,0 +1,5 @@
+---
+'@getodk/web-forms': minor
+---
+
+Render groups without a ref attribute
```
lognaturel, line 5: There's now also a change to the engine that would be helpful to document as a minor change (because we're pre-1.0)?

.changeset/great-deer-march.md
```diff
@@ -0,0 +1,5 @@
+---
+'@getodk/web-forms': minor
```
garethbowen, line 2:
```suggestion
'@getodk/xforms-engine': minor
'@getodk/web-forms': minor
```

lognaturel:
Appreciate you! Nice to have a bug fix and a nice simplification. 🎉

PR #502, opened 2025-09-25
fix(#117): throw an error when called with a ref that does not exist

garethbowen:

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

New tests added to catch the error.

As far as I can tell this situation indicates a form that can never be valid, so failing loudly and early is preferable to find the issue ASAP. 

Discussion: https://github.com/getodk/web-forms/pull/496#discussion_r2377274153

Instead of invalid forms swallowing the error, it will now be thrown.

### DIFF
--- a/.changeset/shy-carpets-add.md
+++ b/.changeset/shy-carpets-add.md
@@ -4,4 +4,4 @@
 '@getodk/web-forms': patch
 ---
 
-Throw an error when jr:choice_name is called with an element reference that does exist
+Throw an error when jr:choice_name is called with a reference that does not exist
--- a/.changeset/shy-carpets-add.md
+++ b/.changeset/shy-carpets-add.md
@@ -1,0 +1,7 @@
+---
+'@getodk/scenario': patch
+'@getodk/xpath': patch
+'@getodk/web-forms': patch
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/shy-carpets-add.md
```diff
@@ -0,0 +1,7 @@
+---
+'@getodk/scenario': patch
+'@getodk/xpath': patch
+'@getodk/web-forms': patch
+---
+
+Throw an error when jr:choice_name is called with an element reference that does exist
```
lognaturel, line 7: There’s a missing “not”! Small nit since I’m here: how about “reference” rather than “element reference”?

garethbowen: Thanks! Fixed.

lognaturel:
> [build:tree-sitter ] Unable to find image 'emscripten/emsdk:3.1.64' locally
[build:tree-sitter ] docker: Error response from daemon: unauthorized: authentication required

Surprising! Any idea what that might be about?

garethbowen:
None. I've posted to Slack for help... Maybe it's transient?

lognaturel (approved):

PR #509, opened 2025-10-03
chore: automated tests for select one for map

latin-panda:
Related to https://github.com/getodk/web-forms/issues/305

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [x] Not applicable

Running CI multiple times

The map feature is highly visual, and testing it at the unit level would require extensive mocking of OpenLayers functions. It would significantly reduce the value of the tests since the real rendering logic would be bypassed.

Using E2E tests allows us to verify the actual behavior of the map, including OpenLayers’ rendering and our integration code, without relying on mocks. This provides the most accurate validation of how the map appears and behaves in the real application.

We use snapshots because many interactions can’t be queried with selectors, as the map and its features are rendered on a canvas. This approach allows us to verify the display of features (points, lines, and shapes) and their states (unselected, selected, saved), as well as zooming and panning behavior.

NA

NA

### What's changed
- Adds e2e tests for select one from maps, verifying the main functionality of the map features and handling geolocation access denial.
- Enhances the All Questions form's map attachment by incorporating point, line, and shape features in a simplified city layout, making visual test assessments easier.
- Uses snapshots to verify map states, automatically generated by Playwright ([docs](https://playwright.dev/docs/test-snapshots#generating-screenshots)). 
- Uses the `mouse` event instead of `click` to select features and pan the map, because `click` doesn't work for this element. 
- Uploads Playwright results to facilitate debugging
- Extract code that converts from BaseItem to GeoJson and add a unit test. The logic remains the same; just moved the code to a new file.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,0 +1,39 @@
+---
+name: Verified Bug
+about: Report a confirmed bug in ODK Web Forms
+title: '<High-level descriptive title>'
+type: 'bug'
+assignees: ''
+---
+
+### Expected Behaviour
+
+<!-- A clear and concise description of what you expected to happen. -->
+
+### Observed Behaviour
+
+<!-- A clear and concise description of what actually happened, including any error messages. -->
+
+### Reproduction Steps
+
+<!-- Provide the form and a clear, step-by-step guide to reproduce the bug. -->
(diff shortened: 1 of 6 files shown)


latin-panda:
🎉🎉🎉🎉

PR #528, opened 2025-10-17
chore: add screenshot tests for markdown

garethbowen:


packages/web-forms/e2e/test-cases/markdown.test.ts
```diff
@@ -0,0 +1,76 @@
...
+	}
+	await page.waitForTimeout(ANIMATION_TIME);
+}
+
+async function selectDropdownOption(page: Page, container: Locator, label: string) {
+	const dropdown = container.locator('.dropdown');
+	await dropdown.click();
+	await page.locator(`.p-select-option:has-text("${label}")`).click();
+	await page.waitForTimeout(ANIMATION_TIME);
+}
```
latin-panda: These functions can be part of a [SelectControl](https://github.com/getodk/web-forms/tree/main/packages/web-forms/e2e/page-objects/controls), we'll probably need them in the future

packages/web-forms/e2e/test-cases/markdown.test.ts-snapshots/user-entered-chromium.png
latin-panda: Chrome Linux has a known issue where screenshots are made one pixel bigger failing the test in macOS 🙁. [Before I had to add this fix](https://github.com/getodk/web-forms/blob/45a3d1665e8978ff15df728ef1f2eea21f079067/packages/web-forms/e2e/page-objects/controls/MapControl.ts#L159-L174).

Firefox and Webkit are fine; they produce screenshots of the same size consistently.

packages/web-forms/e2e/test-cases/markdown.test.ts-snapshots/user-entered-webkit.png
latin-panda: This is strange; the font looks very different on macOS (see italics). I wonder if there's an issue in the CSS, since we've only been testing on MacOS and haven't spotted it before 🤔





I'm not sure if this is a quick font fix. 

I've tried to avoid including OS specific screenshots, but maybe we can consider that again? 🥲 at least for this suite. That's Playwright default anyways, this is where [I changed it.](https://github.com/getodk/web-forms/blob/45a3d1665e8978ff15df728ef1f2eea21f079067/packages/web-forms/playwright.config.ts#L33) 

These are the snapshots for MacOS: 
- [markdown.test.ts-snapshots-darwin.zip](https://github.com/user-attachments/files/22970122/markdown.test.ts-snapshots-darwin.zip)
- [all-question-types.test.ts-snapshots-darwin.zip](https://github.com/user-attachments/files/22970359/all-question-types.test.ts-snapshots-darwin.zip)

packages/web-forms/e2e/test-cases/markdown.test.ts
garethbowen: Done.

garethbowen:
I've been thinking about this more over the weekend. If we have to have 2-5% variance then are we going to actually catch any mistakes? Is it going to be possible to get the same os and the same browser to give results that are consistent enough that the tests are useful?

packages/web-forms/e2e/test-cases/markdown.test.ts-snapshots/user-entered-chromium.png
latin-panda: Yes, that's a good idea, let's use clip :)

packages/web-forms/e2e/test-cases/markdown.test.ts-snapshots/user-entered-webkit.png
latin-panda: Yes, that's better, running them in a Linux container and as a separate suite

latin-panda:
> I've been thinking about this more over the weekend. If we have to have 2-5% variance then are we going to actually catch any mistakes? Is it going to be possible to get the same os and the same browser to give results that are consistent enough that the tests are useful?

I saw some examples using a maximum of 100 pixels of variance. For markdown, where small details matter more, I think it's okay to be more strict. Maybe 1 - 1.5%?

When comparing snapshots from different browsers, the differences are minor. Border lines and shapes are the same, but the font appears a bit thicker. It reported the font variation on my Mac, where the italic style renders differently. 
It can help identify if the feature stops working. I can't think of a better way to test this feature's rendering other than Manual Testing

packages/web-forms/e2e/test-cases/markdown.test.ts-snapshots/user-entered-chromium.png
garethbowen: I tried clip but I couldn't figure out how to get it to work. The documentation isn't useful. I think it's only applied when doing `fullPage` screenshots as an alternative to screenshotting the HTML element.

garethbowen:
@latin-panda Here's a different take... apparently the dimension inconsistency only impacts screenshots by HTML element, so I've switched to using a single `fullPage` screenshot instead. This isn't ideal because it's not as clear which part has changed, and a diff ratio of a bigger image allows for more things to change before failing, but I think it takes more consistent screenshots (dimensions should be stable now) so we should be able to dial up the sensitivity. What do you think?

PR #536, opened 2025-10-29
feat(#257): Add geopoint with maps and placement-map appearances

### LINKED ISSUE #257: Form filling: Geopoint with map
## User stories
As a data collector,
I want to see and confirm my location on a map,
so I can be sure the selected point is accurate and correctly placed

Given I am answering a geo question that includes a map,
When I go to the question,
Then I see a disabled map with a "Find my location" button,
So that my location isn't used unless I choose to enable it.

Given I have granted Web Forms permission to access my location,
When I tap "Find my location"
Then a blue pin appears at my current location on the map,
And I can long-press anywhere on the map to set a point.

Given I have long-pressed to select a location,
Then a green glow appears around the pin to confirm the selection,
And when I tap "View details"
Then I see the geopoint coordinates displayed below,
So that I can verify the point is correct or remove the point and try again

## Designs

[Figma](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/Web-Forms?node-id=549-9375&t=Y5S9NbCrHF694ul7-1)

See screenshots in [the test plan](https://github.com/getodk/web-forms/issues/257#issuecomment-3407783362).

## Related 
- https://github.com/getodk/web-forms/issues/250
- https://github.com/getodk/web-forms/issues/258
- https://github.com/getodk/web-forms/issues/308
- https://github.com/getodk/web-forms/issues/309
- https://github.com/getodk/web-forms/issues/321
- https://github.com/getodk/web-forms/issues/543

latin-panda:
Closes #257

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [ ] Safari (iOS)
- [x] Chrome for Android
- [ ] Not applicable

Verified test plan
The e2e tests are passing

Layer 
- Geopoint only needs to display a single feature. That's why we're using a regular VectorLayer, which offers more tools for interactions like dragging a feature than a WebGL Layer. Later, we will want to use this layer for drawing polygons and lines, styling vertices, and dragging vertices. 
- "Select one from the map" will continue using the WebGL layer, as it offers better performance for loading many features and it doesn't require complex interactions (like drawing). 

Modes
- I tried out different methods to represent modes in this code and chose the simplest and lightest solution I found. It's based on a configuration file, detailing what each mode can do and cannot do. This approach works well because each mode has specific and unique interactions, avoiding any mixing between them. 
- The mode configuration has its own file, making testing much easier. 
- Its usage is incorporated into a couple of functions, which helps avoid conditionals everywhere. It reduces cognitive load, the potential for bugs and keeps the code cleaner.

Modules
- The `MapBlock` Vue component's main responsibility is UI. Keeping minimal logic. 
- The `useMapBlock` is the main map manager, in charge of orchestration, and to expose all functions for `MapBlock` Vue component. I will import other helper modules (useMapFeatures, useMapInteractions, useMapViewControl).
- The `useMapFeatures` is in charge of loading, saving and selecting features. 
- The `useMapInteractions` is in charge of dragging, long press, click events (select). 
- The `useMapViewControl` is in charge of centering the view, watch user location.

- Geopoint with no appearance should work the same, no changes.
- "Select one from map" is evolving to maintain consistent user experience across maps, mainly in icons, status bar messages, and current location behavior (see test plan). 

See the test plan and the project pointed out by Szymon for forms. 

### What's changed
- Refactors code to introduce modes and modules 
- Adds long press to create a feature type point
- Adds drag to relocate a feature type point
- Changes to logic for getting user location to a watch and a stop watch on scroll. 
- Fixes alignment of pin icon to be more precise where it points. 
- Creates a new input component for geopoint with map

### DIFF
--- a/packages/web-forms/src/components/common/map/MapStatusBar.vue
+++ b/packages/web-forms/src/components/common/map/MapStatusBar.vue
@@ -33,7 +33,7 @@ const emit = defineEmits(['view-details', 'save', 'discard']);
 				<span>–</span>
 				<!-- TODO: translations -->
 				<span class="mobile-only">Remove</span>
-				<span class="desktop-only">Remove selection</span>
+				<span class="desktop-only">Remove point</span>
 			</Button>
 			<Button v-if="canViewDetails" outlined severity="contrast" @click="emit('view-details')">
 				<!-- TODO: translations -->
@@ -51,7 +51,7 @@ const emit = defineEmits(['view-details', 'save', 'discard']);
 				<IconSVG name="mdiCheckboxMarkedCircleOutline" size="sm" variant="inverted" />
 				<!-- TODO: translations -->
 				<span class="mobile-only">Save</span>
-				<span class="desktop-only">Save selected</span>
+				<span class="desktop-only">Save point</span>
 			</Button>
 		</div>
 	</div>


packages/web-forms/src/components/common/map/MapBlock.vue
```diff
@@ -28,41 +31,36 @@ const showErrorStyle = inject<ComputedRef<boolean>>(
...
-	},
+	(newData) => mapHandler.updateFeatureCollection(newData, props.savedFeatureValue),
 	{ deep: true }
 );
 
-watch(
-	() => props.savedFeatureValue,
-	(newSaved) => mapHandler.setSavedByValueProp(newSaved)
-);
+watch(() => props.savedFeatureValue, mapHandler.setSavedByValueProp);
```
garethbowen: I prefer the previous style because it's clearer that attributes are being passed through from the watch callback to setSavedByValueProp.

This pattern can also lead to really obscure bugs because it's implicitly also passing through other arguments.

packages/web-forms/src/components/common/map/MapBlock.vue
```diff
@@ -28,41 +31,36 @@ const showErrorStyle = inject<ComputedRef<boolean>>(
...
 
-watch(
-	() => props.disabled,
-	(newValue) => mapHandler.toggleClickBinding(!newValue)
-);
+watch(() => props.disabled, mapHandler.setupMapInteractions);
```
garethbowen: Same here.

packages/web-forms/src/components/common/map/useMapInteractions.ts
```diff
@@ -0,0 +1,205 @@
...
+					return;
+				}
+
+				const distanceX = Math.abs(event.pixel[0] - startPixel[0]);
+				const distanceY = Math.abs(event.pixel[1] - startPixel[1]);
+				if (distanceX > pixelTolerance || distanceY > pixelTolerance) {
+					clearTimeout(timer);
+					timer = null;
+					startPixel = null;
+					mapInstance.getTargetElement().style.cursor = '';
```
garethbowen: Extract a function for this and reuse in the handleUpEvent which (I think) should be identical.

packages/web-forms/src/components/common/map/useMapInteractions.ts
```diff
@@ -0,0 +1,205 @@
...
+
+	const setupFeatureDrag = (layer: VectorLayer, onDrag: (feature: Feature) => void) => {
+		if (translateInteraction.value) {
+			return;
+		}
+
+		translateInteraction.value = new Translate({ layers: [layer] });
+
+		translateInteraction.value.on('translating', () => {
+			mapInstance.getTargetElement().style.cursor = 'grab';
```
garethbowen: Extract this line too? It's simple but used everywhere...

packages/web-forms/src/components/common/map/useMapViewControls.ts
```diff
@@ -0,0 +1,159 @@
...
+			onSuccess();
+		};
+
+		const handleError = () => {
+			currentLocationSource.clear(true);
+			onError();
+		};
+
+		if (!navigator.geolocation) {
+			handleError();
```
garethbowen: If you don't return here you'll get an exception on line 88 I think?

```suggestion
			handleError();
			return;
```

packages/web-forms/tests/components/common/map/createFeatureCollectionAndProps.test.ts
```diff
@@ -46,7 +46,7 @@ describe('createFeatureCollectionAndProps', () => {
 			],
 		});
 
-		expect(orderedExtraPropsMap).toEqual(new Map([['point1', []]]));
+		expect(orderedExtraPropsMap).toEqual(new Map());
```
garethbowen: I must've missed something... what changed here?

packages/web-forms/src/components/common/map/useMapInteractions.ts
latin-panda: fixed

packages/web-forms/tests/components/common/map/createFeatureCollectionAndProps.test.ts
latin-panda: It removes properties that have no value, so they won't show up in the properties popup. This means less scrolling, too.

packages/common/src/fixtures/geopoint/geopoint-map.xml
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0"?>
+<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml"
```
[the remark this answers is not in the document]
latin-panda: I'll improve this form tomorrow, I ran out of time today

garethbowen (commented):
Ran the code today on my dev machine (firefox, linux, desktop) and have some suggestions...

1. I clicked the "zoom to current location" button, which was way off (unsurprising for a desktop), so I started scrolling around, but before I could get to my location, the map panned back to the browser provided GPS co-ords. This happens reliably - within a few seconds it always sends me back. My guess is that there's an event firing over and over again.
2. If you click "zoom to current location" it takes some time (as expected), and when it completes, you click "zoom to current location" on the second map, it takes just as long again. I'm assuming the gps is local to the widget and not shared globally. It might be worth having a central service to cache the GPS location with some timeout so services can share this as it can take potentially a long time, and also can use a lot of battery. This is probably out of scope (most forms won't have multiple maps?) so feel free to split this out as a separate issue.
3. The map starts in New York which is a long way away from my current location. That means when my location is found the map has to scroll a long way over map tiles that it hasn't yet downloaded. The effect of this is a rather jarring flashing between blue of the Pacific Ocean and the grey of the not yet loaded tiles. Downloading and rendering these tiles is wasteful, and the visual impact is inelegant (and flashing isn't good for some users!). It might be more useful to do a slight pan (start from 1km away) or a small zoom (start from zoomed out 2x and zoom in to the usual). That way we're downloading tiles that are actually useful, and the visual effect is smoothed. 
4. The spinner and message when detecting location is great, but because you're allowed to interact with the map during the detection period you may do that and then be whisked away to your location without warning. What about using the overlay that is shown for a required field as the "getting location" splash, which disables the UI which it's fetching? You could add a cancel button if the user gets bored and wants to manually find their location. This would work well with number 3 above.

I'll test on mobile (which with an actual GPS unit will be more representative) another day but that's probably enough to be getting on with!

alyblenkin:
Lots of great feedback. I wanted to quickly jump on on these points..

>3. It might be more useful to do a slight pan (start from 1km away) or a small zoom (start from zoomed out 2x and zoom in to the usual). That way we're downloading tiles that are actually useful, and the visual effect is smoothed.

I really like this idea if it isn't too much work! 

> 4. The spinner and message when detecting location is great, but because you're allowed to interact with the map during the detection period you may do that and then be whisked away to your location without warning. What about using the overlay that is shown for a required field as the "getting location" splash, which disables the UI which it's fetching? You could add a cancel button if the user gets bored and wants to manually find their location. This would work well with number 3 above.

I think this is worth testing with users. Our current assumption is that users may want to interact with the map while their location is being detected. I agree it can feel disorienting to be suddenly moved, but we’ve been assuming users prefer to keep the map visible so they can confirm they’re in the right spot as they wait.

Agreed adding the cancel button could be really useful here. Maybe we layer that on later? Not sure how much work that would be.

packages/web-forms/src/components/common/map/useMapViewControls.ts
```diff
@@ -0,0 +1,187 @@
...
+			center: DEFAULT_VIEW_CENTER,
+			zoom: MIN_ZOOM,
+			duration: ANIMATION_TIME,
+		});
+	};
+
+	watch(
+		() => userCurrentLocation.value,
+		(newLocation) => {
+			const canCenterView = !userCurrentLocationFeature.value;
```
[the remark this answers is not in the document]
latin-panda: Yes, I see what you're saying. It was requested to add a watch and update the map. Some users might move around (walking or driving) and want to see their current location on the map. It's up to the user to decide when the location is good enough and save it. Saving the location and scrolling the question out of the viewport will stop the watch, which helps free up some resources.

latin-panda:
Thanks @garethbowen! I've addressed the last feedback; it was just a settings issue that needed to be turned on. It's now ready for review.

I'll be adding the new e2e tests in a fresh PR I’d like to extract the visual tests into their own suite and run them in a container. So, we can avoid those discrepancies that appeared in the snapshots a few weeks ago (macOS vs Linux).

I'm sending this work to QA tomorrow, I'll prepare the server.

PR #542, opened 2025-11-04
feat(#285): bind and serialize attributes

### LINKED ISSUE #285: Support attributes
Like elements, they can be bound, generally to calculations.

Attributes are key to [Entities support](https://getodk.github.io/xforms-spec/entities).

garethbowen:
Closes #285

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

- Scenario tests added
- Tested locally in Central and I can create and edit entities

It's new functionality, and shouldn't change anything that worked before.

In the demo app there's an entities > bind-attributes form which is a start (check the console for the submitted xml)
To fully test you'll need to get this branch running in central then I recommend following this tutorial: https://docs.getodk.org/central-entities/#central-entities-follow-up-forms

### What's changed

- Attributes are now serialized on submission
- Attributes can be bound using `@name` syntax
- This only applies to parent nodes and not leaf nodes currently - the framework is there for leaf nodes but I don't have for this yet

### DIFF
--- a/packages/xforms-engine/src/lib/client-reactivity/instance-state/createParentNodeInstanceState.ts
+++ b/packages/xforms-engine/src/lib/client-reactivity/instance-state/createParentNodeInstanceState.ts
@@ -12,17 +12,11 @@ export const createParentNodeInstanceState = (
 				return '';
 			}
 
-			const serializedChildren = node.currentState.children.map((child) => {
-				return child.instanceState.instanceXML;
-			});
-
-			const attributes = node.currentState.attributes.map((attribute) => {
-				return { serializeAttributeXML: () => attribute.instanceState.instanceXML };
-			});
-
-			return serializeParentElementXML(node.definition.qualifiedName, serializedChildren, {
-				attributes,
-			});
+			return serializeParentElementXML(
+				node.definition.qualifiedName,
+				node.currentState.children,
+				node.currentState.attributes
+			);
 		},
 	};
 };
--- a/packages/xforms-engine/src/lib/xml-serialization.ts
+++ b/packages/xforms-engine/src/lib/xml-serialization.ts
@@ -1,3 +1,5 @@
+import type { GeneralChildNode } from '../client/hierarchy.ts';
+import type { Attribute } from '../instance/Attribute.ts';
 import type { NamespaceDeclarationMap } from './names/NamespaceDeclarationMap.ts';
 import type { QualifiedName } from './names/QualifiedName.ts';
 
@@ -78,15 +80,6 @@ export const escapeXMLText = <Text extends string>(
 		: (out as EscapedXMLText);
 };
 
-interface SerializableElementAttribute {
-	serializeAttributeXML(): string;
-}
-
-interface ElementXMLSerializationOptions {
-	readonly namespaceDeclarations?: NamespaceDeclarationMap;
-	readonly attributes?: readonly SerializableElementAttribute[];
-}
-
 const serializeElementNamespaceDeclarationXML = (
 	namespaceDeclarations?: NamespaceDeclarationMap
 ): string => {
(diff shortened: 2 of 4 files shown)


packages/xforms-engine/src/parse/model/RootAttributeDefinition.ts
```diff
@@ -1,38 +1,59 @@
...
 	readonly value: string;
 }
 
-/**
- * @todo This class is named and typed to emphasize its intentionally narrow
- * usage and purpose. It **intentionally** avoids addressing the much broader
- * set of concerns around modeling attributes in primary instance/submissions.
- */
-export class RootAttributeDefinition implements NamedNodeDefinition {
+export class AttributeDefinition
```
latin-panda: I like the new name `AttributeDefinition`; `root` was too generic. Can this file adopt the name from this class? Not sure if there's any conflict.

packages/xforms-engine/src/parse/model/RootAttributeDefinition.ts
```diff
@@ -1,38 +1,59 @@
...
+	readonly children = null;
+	readonly attributes = null;
+
+	readonly qualifiedName: QualifiedName;
+
+	constructor(
+		root: RootDefinition,
+		source: RootAttributeSource,
+		bind: BindDefinition,
+		readonly template: StaticAttribute
```
latin-panda: Why does this class need this `template` parameter?

packages/xforms-engine/src/lib/names/NamespaceDeclarationMap.ts
```diff
@@ -11,7 +11,7 @@ type NamedNodeDefinitionMap = ReadonlyMap<QualifiedName, NamedNodeDefinition>;
 export interface NamedSubtreeDefinition extends NamedNodeDefinition {
 	readonly namespaceDeclarations: NamespaceDeclarationMap;
 	readonly parent: NamedSubtreeDefinition | null;
-	readonly attributes?: NamedNodeDefinitionMap;
+	readonly attributes: NamedNodeDefinitionMap | null;
```
latin-panda: Not a suggestion, just a comment: 
This is because LeafNode won't have attributes. When I see this, it makes me a bit hesitant to trust `attributes`, but I suppose it's okay since adding other subtrees like Group, Repeat, or LeafNode isn’t very common.

latin-panda (commented):
It's exciting to see this! Thank you for the comments, it helped me understand better. The code looks great!

I ran out of time, but I plan to test it later, I'll prepare my local Central with some scenarios

packages/xforms-engine/src/parse/model/RootAttributeMap.ts
```diff
@@ -29,17 +23,17 @@ const isNonNamespaceAttribute = (attribute: StaticAttribute) => {
  *
  * @see {@link QualifiedName} for more detail.
  */
-export class RootAttributeMap extends Map<QualifiedName, RootAttributeDefinition> {
-	static from(root: RootDefinition, instanceNode: StaticElement) {
+export class AttributeDefinitionMap extends Map<QualifiedName, AttributeDefinition> {
```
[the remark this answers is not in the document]
garethbowen: Yes, sorry, done!

packages/xforms-engine/src/lib/xml-serialization.ts
```diff
@@ -146,6 +146,14 @@ const serializeElementXML = (
 	return `${prefix}>${children}</${nodeName}>`;
 };
 
+export const serializeAttributeXML = (
+	qualifiedName: QualifiedName,
+	xmlValue: EscapedXMLText
+): string => {
+	const nodeName = qualifiedName.getPrefixedName();
+	return ` ${nodeName}="${xmlValue.normalize()}"`;
```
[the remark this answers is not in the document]
garethbowen: Yes that's needed. It's used in this line:

```
const prefix = `<${nodeName}${namespaceDeclarations}${attributes}`;
```

Without the space there would be no whitespace between the nodeName and the attributes. This is implemented in a slightly different place for namespaceDeclarations in NamespaceDeclaration.ts

packages/xforms-engine/src/lib/names/NamespaceDeclarationMap.ts
latin-panda: Ah, I missed that `?` 🥲

packages/xforms-engine/src/lib/client-reactivity/instance-state/createParentNodeInstanceState.ts
```diff
@@ -16,7 +16,13 @@ export const createParentNodeInstanceState = (
...
 			});
 
-			return serializeParentElementXML(node.definition.qualifiedName, serializedChildren);
+			const attributes = node.currentState.attributes.map((attribute) => {
+				return { serializeAttributeXML: () => attribute.instanceState.instanceXML };
+			});
+
+			return serializeParentElementXML(node.definition.qualifiedName, serializedChildren, {
+				attributes,
+			});
```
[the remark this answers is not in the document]
latin-panda: I like that idea!

PR #553, opened 2025-11-13
feat(#547): allow attributes on leaf nodes

### LINKED ISSUE #547: Expand support for attributes to leaf nodes
### Description

Build on the work in #542 to support attributes on leaf nodes.

This was left out of the initial implementation due to complexity around the types requiring significant rework to support attributes having non-parent owners.

Currently there's no way for forms generated with pyxform to have attributes on leaf nodes, but the spec allows it so forms generated in other ways may do. Leaving this issue here as documentation for the current lack of support.

garethbowen:
Closes #547

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

### What's changed

Attributes are now less like regular Elements because sharing too much of the stack was requiring all sorts of properties that aren't relevant.

### DIFF
--- a/packages/xforms-engine/src/instance/Attribute.ts
+++ b/packages/xforms-engine/src/instance/Attribute.ts
@@ -19,12 +19,14 @@ import {
 	createSharedNodeState,
 	type SharedNodeState,
 } from '../lib/reactivity/node-state/createSharedNodeState.ts';
+import type { ReactiveScope } from '../lib/reactivity/scope.ts';
 import type { SimpleAtomicState } from '../lib/reactivity/types.ts';
 import type { AttributeDefinition } from '../parse/model/AttributeDefinition.ts';
 import type { AnyChildNode, AnyNode } from './hierarchy.ts';
 import type { AttributeContext } from './internal-api/AttributeContext.ts';
 import type { DecodeInstanceValue } from './internal-api/InstanceValueContext.ts';
 import type { ClientReactiveSerializableAttributeNode } from './internal-api/serialization/ClientReactiveSerializableAttributeNode.ts';
+import type { PrimaryInstance } from './PrimaryInstance.ts';
 import type { Root } from './Root.ts';
 
 export interface AttributeStateSpec {
@@ -60,9 +62,9 @@ export class Attribute
 	protected readonly setValueState: RuntimeValueSetter<RuntimeInputValue<'string'>>;
 	readonly evaluator: EngineXPathEvaluator;
 	readonly getActiveLanguage: Accessor<ActiveLanguage>;
-	readonly contextNode;
-	readonly scope;
-	readonly rootDocument;
+	readonly contextNode: AnyNode;
+	readonly scope: ReactiveScope;
+	readonly rootDocument: PrimaryInstance;
 
 	readonly root: Root;
 


packages/xforms-engine/src/instance/attachments/buildAttributes.ts
```diff
@@ -1,8 +1,15 @@
...
+import type { ModelValue } from '../ModelValue.ts';
+import type { Note } from '../Note.ts';
+import type { RangeControl } from '../RangeControl.ts';
 
-export function buildAttributes(parent: AnyParentNode): Attribute[] {
-	return Array.from(parent.definition.attributes.values()).map((attributeDefinition) => {
-		return new Attribute(parent, attributeDefinition, attributeDefinition.template);
+export function buildAttributes(
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	owner: AnyNode | InputControl<any> | ModelValue<any> | Note<any> | RangeControl<any>
```
garethbowen: This is ugly, but `AnyNode` alone doesn't work when called with generics. Any ideas?

latin-panda:
```
export function buildAttributes(
	owner: AnyNode
): Attribute[] {
	return Array.from(owner.definition.attributes.values()).map((attributeDefinition) => {
		return new Attribute(owner, attributeDefinition, attributeDefinition.template);
	});
}
```
Like this, it works fine when running `test:types` and `lint`. It seems you fixed it when simplifying the higher-level classes.

garethbowen: When I do that I get TS compilation errors...

```
yarn workspace @getodk/xforms-engine build:js

src/instance/RangeControl.ts:108:48 - error TS2345: Argument of type 'this' is not assignable to parameter of type 'AnyNode'.
  Type 'RangeControl<V>' is not assignable to type 'AnyNode'.
    Type 'RangeControl<V>' is not assignable to type 'RangeControl<"decimal"> | RangeControl<"int">'.
      Type 'RangeControl<V>' is not assignable to type 'RangeControl<"decimal">'.
        Type at position 0 in source is not compatible with type at position 0 in target.
          The types of 'state.spec.value' are incompatible between these types.
            Type 'Accessor<RangeValue<V>>' is not assignable to type 'Accessor<DecimalRuntimeValue>'.
              Type 'RangeValue<V>' is not assignable to type 'DecimalRuntimeValue'.
                Type 'number | bigint | null' is not assignable to type 'DecimalRuntimeValue'.
                  Type 'bigint' is not assignable to type 'DecimalRuntimeValue'.
                    Type 'this' is not assignable to type 'PrimaryInstance<FormInstanceInitializationMode>'.
                      Type 'RangeControl<V>' is missing the following properties from type 'PrimaryInstance<FormInstanceInitializationMode>': initializationMode, model, attachments, setActiveLanguage, and 4 more.

108   attributeState.setAttributes(buildAttributes(this));
```

I'll stick with the ugly but compiler friendly option for now to keep this rolling, but if you figure it out, let me know!

PR #564, opened 2025-11-21
chore: provide TZ parameter so 'yarn test' passes

garethbowen:
Fixing a minor annoyance...


alxndrsn (approved):
Making dev more accessible & intuitive 👍

PR #566, opened 2025-11-25
chore: upgrade dependencies

latin-panda:

- [x] Chrome
- [ ] Firefox
- [x] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

- I tested with All Questions form and didn't find anything wrong
- CI passes

### What's changed
After the upgrade, most of the errors fell into two categories:
- Types are redundant when using union
    - In these cases, some of the types were actually making the code clearer, so I chose to disable the type check in line. 
- Possible undefined variables
    - Some of these were false positives where it couldn’t trace or infer the types of dependencies that were actually defined. For those, I also decided to skip the type check.

After upgrading Playwright, the snapshots for WebKit were 1px solid. This PR updates the snapshots.


packages/web-forms/src/components/common/map/createFeatureCollectionAndProps.ts
```diff
@@ -48,11 +48,11 @@ const getGeoJSONCoordinates = (geometry: string) => {
 
 const getGeoJSONGeometry = (coords: Coordinates[]): Geometry => {
 	if (coords.length === 1) {
-		return { type: 'Point', coordinates: coords[0] };
+		return { type: 'Point', coordinates: coords[0]! };
```
latin-panda: These here and below have already been validated as not being undefined by `getGeoJSONCoordinates` and `createFeatureCollectionAndProps`. However, after the upgrade, it couldn't infer the type.

packages/xforms-engine/src/client/validation.ts
```diff
@@ -204,4 +204,4 @@ export interface NullValidationState {
 export type NodeValidationState =
 	| AncestorNodeValidationState
 	| LeafNodeValidationState
-	| NullValidationState;
+	| NullValidationState; // eslint-disable-line @typescript-eslint/no-redundant-type-constituents
```
latin-panda: @garethbowen I believe you added this type, but it's considered redundant. I left it because it still adds some clarity to the code. Does it make sense?

latin-panda:
@garethbowen This is ready for review. I didn't upgrade PrimeVue because it breaks the calendar's clear button (known issue)

packages/xpath/src/adapter/defaults.ts
```diff
@@ -20,7 +20,7 @@ import { xpathDOMProvider } from './xpathDOMProvider.ts';
  * types to their WHAT* equivalent. This is for clarity at usage and reference
  * sites, internally but especially at the package boundary.
  */
-type AsDefault<T> = T | (T & { readonly _?: never });
+type AsDefault<T> = T | (T & { readonly _?: never }); // eslint-disable-line @typescript-eslint/no-redundant-type-constituents
```
garethbowen: This is popping up so frequently maybe we need to disable it in the tsconfig?

latin-panda: I've resolved some of these and removed the skip of half of them 🙂

packages/web-forms/src/components/common/map/createFeatureCollectionAndProps.ts
```diff
@@ -43,16 +43,16 @@ const getGeoJSONCoordinates = (geometry: string) => {
...
 
-const getGeoJSONGeometry = (coords: Coordinates[]): Geometry => {
+const getGeoJSONGeometry = (coords: [Coordinates, ...Coordinates[]]): Geometry => {
 	if (coords.length === 1) {
 		return { type: 'Point', coordinates: coords[0] };
 	}
 
 	const [firstLongitude, firstLatitude] = coords[0];
-	const [lastLongitude, lastLatitude] = coords[coords.length - 1];
+	const [lastLongitude, lastLatitude] = coords[coords.length - 1]!; // ! because coords.length > 1
```
latin-panda: I had a different type assertion here but lint suggested this `!` instead

garethbowen:
> I didn't fix the ! in test files

Agreed. If a test hits a null reference then we should know about it.

> By the way, have you had any issues running `test:types`?

Yes! I hit that too. I think it's to do with the yarn upgrade (sorry!). I have a fix planned but it touches the dependencies so I was waiting for this PR to get merged to avoid potential conflicts.

PR #585, opened 2025-12-11
feat: add geoshape and geotrace

### LINKED ISSUE #250: Editing: Geopoint with no appearance
As of right now, geo questions are the only ones that differ form filling experience because we want the project managers to have all the functionality available to edit with more context no matter what the appearance is.

Given I'm editing a submission with a default geo question (without map)
When I edit a point
Then I should see all functionality (e.g. map, search bar, place name, and coordinate text fields) 
So that I can adjust the accuracy with additional context

### LINKED ISSUE #308: Add geotrace support
### Description
Collect has 3 recording modes for creating a trace: manual, automatic, and placement. Web Forms will differ slightly and support placement and manual modes.

When a user opens a geotrace question in Web Forms, they won’t see a location overlay because we are not requesting permissions up front. If they want to manually add a point using their current location, they can tap the “Find location” button to enable it.

The goal is for them to create a trace using placement or manual modes and not have to worry about saving (continuous saving). 

---

### Checklist
  - [] No
  - [x] No
  - [x] Yes. Outline the state flow requirements.
     Need to make a decision on when that case is identified and how it's communicated down (e.g. would manually forcing the placement-map appearance feel appropriate)
  - [ ] No
  - [x] Yes. Add the test plan as a comment in this task or attach the document.
  - [ ] No
---

### User Stories

**Starting a trace**
- As a user, I want to start adding points in placement mode without granting location permissions, as I may not want to map my physical location. 
- As a user, I only want to share my location when needed (e.g., for manual mode), so that I can map my physical location
- As a user, I want to see a clear status such as “No points saved” when I begin, so that I know the trace hasn’t started yet. And it updates once I place points. 

**Adding/editing points**
- As a user, I want to add points by long-pressing on the map, so that I can precisely define the line I want to capture.
- As a user, I want to move a point by long-pressing and dragging, so that I can correct mistakes or refine the trace.
-  As a user, I want points to autosave as I add them, so that my work is never lost.
- As a user, I want the most recently added point to be  selected and visually larger, so that I can easily identify and adjust it.

**Delete or undo**
- As a user, I want to delete the entire shape by selecting the trace (not individual points) and tap the delete button with a confirmation, so that I can start over. 
- As a user, I want to delete/undo a single point by selecting the point and tapping the trash icon, so I can delete easily 
- As a user, I want undo to remove only the most recently added point, so that I can quickly correct mistakes.

---

### Design

- **[Figma](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=3659-14401&t=TVRbd5KZnbscg32x-1)**
- **[Design Docs](https://docs.google.com/document/d/1hkp5XqN7-0ZQg0DeaF4TSIRo7TlQWkwV9FFCnevX-Yg/edit?tab=t.2o484l4qwxto#heading=h.vfqk1ncu7p2o)**

---

### Dependencies

- https://github.com/getodk/web-forms/issues/257

---

### Additional Notes

- [Test plan document](https://docs.google.com/document/d/1hkp5XqN7-0ZQg0DeaF4TSIRo7TlQWkwV9FFCnevX-Yg/edit?tab=t.nooik8hnnok#heading=h.1frinu3q5u61)

---

### Related Issues

- https://github.com/getodk/web-forms/issues/249

### LINKED ISSUE #309: Add geoshape support
### Description
When a user opens a **geoshapes question** in Web Forms, they won’t see a location overlay because we are not requesting permissions up front. If they want to map their physical location, they can tap the “Find location” button to enable it.

Similar to geotrace, the goal is for them to create a shape using placement or manual modes and not have to worry about saving (continuous saving). The shape is also always closed, so they don't need to remember to finish before moving on to the next question.

---

### Checklist
  - [] No
  - [x] No
  - [x] Yes. Outline the state flow requirements.
     Need to make a decision on when that case is identified and how it's communicated down (e.g. would manually forcing the placement-map appearance feel appropriate)
  - [ ] No
  - [x] Yes. Add the test plan as a comment in this task or attach the document.
  - [ ] No
---

### User Stories & designs

**Starting a trace** - [Figma](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=3644-21649&t=tkhhHooXZdufYXHK-1)
- As a user, I want to start adding points in placement mode without granting location permissions, as I may not want to map my physical location. 
- As a user, I only want to share my location when needed (e.g., for manual mode), so that I can map my physical location
- As a user, I want to see a clear status such as “No points saved” when I begin, so that I know the shape hasn’t started yet. And it updates once I place points. 

**Adding/editing points** - [Figma](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=3659-12329&t=tkhhHooXZdufYXHK-1)
- As a user, I want to add points by long-pressing on the map, so that I can precisely define the shape I want to capture.
- As a user, I want to move a point by long-pressing and dragging, so that I can correct mistakes or refine the shape.
-  As a user, I want points to autosave as I add them, so that my work is never lost.
- As a user, I want the most recently added point to be selected and visually larger, so that I can easily identify and adjust it.

**Delete or undo** - [Figma](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=3659-12677&t=tkhhHooXZdufYXHK-1) 
- As a user, I want to delete the entire shape by selecting the shape (not individual points) and tap the delete button with a confirmation, so that I can start over. 
- As a user, I want to delete/undo a single point by selecting the point and tapping the trash icon, so I can delete easily 
- As a user, I want undo to remove only the most recently added point, so that I can quickly correct mistakes.

---

### Dependencies

- https://github.com/getodk/web-forms/issues/257

---

### Additional Notes

- [Test plan document](https://docs.google.com/document/d/1hkp5XqN7-0ZQg0DeaF4TSIRo7TlQWkwV9FFCnevX-Yg/edit?tab=t.nooik8hnnok#heading=h.6hofc76556s)

---

### Related Issues

- https://github.com/getodk/web-forms/issues/249

### LINKED ISSUE #593: Extra circle is visible around "X" button in the Delete dialog
**Expected Behaviour**

There should not be an extra circle around "X" button in the Delete dialog

**Observed Behaviour**

Look at the attached video. After clicking on the dialog the "circle" disappears, but it is visible each time the dialog is opened.

**Reproduction Steps**

1. Open any form with Geoshape or GeoTrace 
2. Create at least 2 points for Geotrace or 3 for Geoshape
3. Select the whole Trace/Shape
4. Click on the Delete button -> Delete dialog should appear

**Screenshots or Videos**

<details> <summary> extra circle </summary>

Image

</details>

**Environment**

    Browser version: Firefox 136.0 (64-bit)
    OS: Ubuntu 20.04.5 LTS
    Web Forms version: v0.18.1

**Additional Context**

Reproduced on Iphone 12 Pro

latin-panda:
Closes #250
Closes #308
Closes #309
Closes #593

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [ ] Safari (iOS)
- [x] Chrome for Android
- [ ] Not applicable

I tested on desktop and 3 Android devices:
  - Samsung      - high-end - Android 16   - Chrome 143.xxx   - works
  - FaibaMobile - 2GB        - Android 13   - Chrome v138.xxx - works
  - Nokia            - 3GB        - Android 9     - Chrome v119.xxx - works
Following the test plan for [Geopoint](https://github.com/getodk/web-forms/issues/257#issuecomment-3407783362), [Select one from map](https://github.com/getodk/web-forms/issues/305#issuecomment-3206808806), [Geoshape](https://docs.google.com/document/d/1hkp5XqN7-0ZQg0DeaF4TSIRo7TlQWkwV9FFCnevX-Yg/edit?tab=t.nooik8hnnok#heading=h.6hofc76556s) and [Geotrace](https://docs.google.com/document/d/1hkp5XqN7-0ZQg0DeaF4TSIRo7TlQWkwV9FFCnevX-Yg/edit?tab=t.nooik8hnnok#heading=h.1frinu3q5u61)

Tests in CI are passing.

I'll send this to QA after the review is finished.

_Pending: test with central - I'll do it on Monday_

- This increases long press time to 1.3s, providing a better experience on mobile and preventing accidental additions when tapping on points. 
- The baked-in interactions from OpenLayers don't support the user experience we want. So I experimented with events for long pressing, hovering, and dragging a lot. The solution in this PR was the closest to providing the desired UX, with the least flakiness and with the best performance in phones with 2GB.
- Normally, OpenLayers handles the placement of new vertices in a segment of a shape or trace, but it doesn't support long-press. I've added a helper function to calculate the correct location in the segment to place the vertex (vertex-geometry.ts)
- It introduces a new mode, `Draw`, and it needs to know somehow the type of feature it intends to draw. This introduces one parameter to the map component `type DrawFeatureType = 'shape' | 'trace'`
- On the engine side, it reuses as much as possible the existing logic for evaluating geopoints by creating classes for geoshape and geotrace and extending from geolocation (formerly named geopoint). 

- Select one from map and Geopoint should work just the same as before. 

- Test server updated and shared in the channel
- This PR adds new forms in common package for testing.

### What's changed
- Adds support of geoshape and geotrace in the xforms-engine. 
- Adds a new (`Draw`) mode, with new interactions (`dragFeatureAndVertex`), and new capabilities (`canUndoLastChange`, etc.0
- Adds new component for control buttons (zoom, my location, fulls screen), confirm dialog for delete action, 
- Updates map status bar to show selected vertex info and the number of points in a shape/trace.
- Adds undo, delete feature and delete vertex actions.  
- Extracts the map control mixing to a shared style file so it can be reused in the corresponding map components. 
- Refactors long press for more stability and better performance (low-end phones).
- Adds a phantom point when hovering lines of the shape or trace.
- Adds minimum support for Geoshape and Geotrace Notes (static readonly fields), reusing the existing component for Geopoint. 
  - We are reevaluating, as part of another ticket, whether Notes is valuable. This work just does the minimum to make the value more readable. 
- Extracts helper functions out of useMapBlock to a new file (map-helpers.ts) to keep the composable targeted and short. 
- Update scenario tests, and increase unit tests and e2e tests coverage. 
- Saves altitude and accuracy when available (ie saving the current location in Geopoint) - in the future, we might need it for shape and trace in ["automation location recording"](https://docs.getodk.org/form-question-types/#geoshape)

### DIFF
--- a/packages/web-forms/src/components/common/map/MapConfirm.vue
+++ b/packages/web-forms/src/components/common/map/MapConfirm.vue
@@ -55,5 +55,10 @@ const emit = defineEmits(['update:visible', 'deleteFeature']);
 	.p-dialog-footer button {
 		font-size: var(--odk-base-font-size);
 	}
+
+	button.p-button-secondary:focus-visible {
+		outline: none;
+		outline-offset: unset;
+	}
 }
 </style>


garethbowen:
Thanks for the heads up. I'm running out of time today but will add this to the todo list :)

packages/web-forms/src/components/common/map/MapStatusBar.vue
```diff
@@ -1,16 +1,91 @@
...
+	const geometry = props.savedFeatureValue?.geometry as LineString | Point | Polygon | undefined;
+	const isShape = props.drawFeatureType === DRAW_FEATURE_TYPES.SHAPE;
+	const coords = geometry?.coordinates ?? [];
+	const count = isShape && Array.isArray(coords[0]) ? coords[0].length : coords.length;
+	if (count === 0) {
+		return null;
+	}
+
+	// TODO: translations
+	const message = `${count} points saved`;
```
garethbowen: This message feels wrong for geoshapes - if I draw a triangle it says "4 points saved" which is internally correct, but from a user perspective totally wrong.

The quick fix would be to subtract one when end === start.

But it might be more useful to the user if we report in terms of shapes for geoshape, and points for geotrace. For example, the message for geoshape could be "no shape saved" or "1 shape saved" rather than "n points saved".

packages/web-forms/src/assets/styles/map-block.scss
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	&:disabled {
+		cursor: not-allowed;
+
+		svg {
+			opacity: 0.5;
+		}
+	}
+}
```
garethbowen: Love it!

packages/web-forms/src/components/common/map/MapConfirm.vue
```diff
@@ -0,0 +1,59 @@
...
+	.p-dialog-header {
+		padding: 15px 20px;
+		font-size: var(--odk-dialog-title-font-size);
+	}
+
+	.p-dialog-content p,
+	.p-dialog-footer button {
+		font-size: var(--odk-base-font-size);
+	}
+}
```
garethbowen: Why is this dialog being styled differently to every other dialog? Can we standardise on one style across web-forms?

packages/web-forms/src/components/common/map/useMapBlock.ts
```diff
@@ -236,55 +310,69 @@ export function useMapBlock(mode: Mode, events: { onFeaturePlacement: () => void
...
+		if (location.altitude != null) {
+			coords.push(location.altitude);
+		}
+
+		if (location.accuracy != null) {
+			coords.push(location.accuracy);
+		}
+
+		const feature = new Feature({
+			geometry: new Point(fromLonLat(coords), 'XYZM'),
```
garethbowen: XYZM is only used twice (outside of tests) but it would save me from Googling what it is if it were a constant with a useful name.

packages/web-forms/src/components/common/map/useMapInteractions.ts
```diff
@@ -76,11 +108,50 @@ export function useMapInteractions(mapInstance: Map): UseMapInteractions {
 		}
 	};
 
+	const onSelectFeatureOrVertex = (
+		event: MapBrowserEvent,
+		onSelect?: (feature: Feature | undefined, selectedVertexIndex: number | undefined) => void
+	): void => {
```
garethbowen: Unless there are sideeffects this call does nothing if the `onSelect` function is not given, right? If so, we can shortcircuit all the work, and also remove the `?` throughout.

```suggestion
	): void => {
	  if (!onSelect) {
	  	return;
	  }
```

packages/web-forms/src/components/common/map/useMapViewControls.ts
```diff
@@ -243,8 +243,13 @@ export function useMapViewControls(mapInstance: Map): UseMapViewControls {
 				return;
 			}
 
-			const parsedCoords = fromLonLat([newLocation.longitude, newLocation.latitude]);
-			userCurrentLocationFeature.value = new Feature({ geometry: new Point(parsedCoords) });
+			const parsedCoords = fromLonLat([
+				newLocation.longitude,
+				newLocation.latitude,
+				newLocation.altitude ?? 0,
```
garethbowen: This feels like a lie - if we don't know the altitude then leaving it undefined makes that clear. Does this cause an issue in OL?

packages/web-forms/src/components/common/map/vertex-geometry.ts
```diff
@@ -0,0 +1,228 @@
...
+	return geometry.getCoordinates()[0] ?? [];
+};
+
+const getDeltaBetweenPoints = (pointA: number[], pointB: number[]) => {
+	const isDefined = (c: number) => c != null;
+	const pairA = pointA.filter(isDefined) as [number, number];
+	const pairB = pointB.filter(isDefined) as [number, number];
+	if (pairA.length < 2 || pairB.length < 2) {
+		return { deltaX: 0, deltaY: 0 };
+	}
```
garethbowen: Filtering this feels strange because if for some reason `x` were null then y would become x and altitude would become y, right?

If so I think it'd be cleaner to leave the coordinates where they are and abort as soon as either a or b is null in index 0 or 1.

eg:


```suggestion
	const isDefined = (point: number[]) => point.length > 1 && point[0] && point[1];
	if (!isDefined(pointA) || !isDefined(pointB)) {
		return { deltaX: 0, deltaY: 0 };
	}
```

packages/web-forms/src/components/common/map/vertex-geometry.ts
```diff
@@ -0,0 +1,228 @@
...
+
+	const [ax, ay] = pairA;
+	const [bx, by] = pairB;
+	return { deltaX: ax - bx, deltaY: ay - by };
+};
+
+const getDistanceBetweenPoints = (deltaX: number, deltaY: number) => {
```
garethbowen: It took me a minute to work out why you were squaring the deltas before realising it's the Pythagorean theorem - could you add a comment?

NB: Very clever, but not clever in the show off kind of way, clever in the right way! Love it.

packages/web-forms/src/components/common/map/vertex-geometry.ts
```diff
@@ -0,0 +1,228 @@
...
+	return { segmentIndex: bestIndex, closest: bestClosest, squaredDist: minSquaredDist };
+};
+
+const isCoordsEqual = (coordA: Coordinate | undefined, coordB: Coordinate | undefined) => {
+	return coordA && coordB && coordA[0] === coordB[0] && coordA[1] === coordB[1];
+};
+
+const isOnLine = (squaredDist: number, resolution: number, hitTolerance: number) => {
+	const tolerance = resolution * hitTolerance;
+	return squaredDist <= tolerance * tolerance;
```
garethbowen:
```suggestion
	return squaredDist <= tolerance ** 2;
```

garethbowen (changes requested):
Massive effort, and great result.

Lots of comments but that's simply a reflection of the size of the change.

Some additional things I noticed when testing (firefox android):

1. Selecting a point on mobile using long press is hard because your finger covers up the point itself so you can't see when the 1.3 seconds is up. The number in the status bar changes but most of the time that is off screen for me, and even if it's not, changing the number is very subtle. I think the ideal solution would be to use a little [haptic feedback](https://developer.mozilla.org/en-US/docs/Web/API/Vibration_API). If we don't like that then we could do a little highlight on the number when it changes, or some other visible animation that it's done.
2. The feature to add a point on a line isn't discoverable on mobile because the mini-dot (phantom?) is only shown on hover. I'm not sure how to resolve this and it's probably not blocking, but we could do something like a single tap on mobile shows the little blue dot, and least hinting that there's something you can do with it.
3. If you select a trace/shape, then select a point, then click delete it tries to delete the trace/shape instead of just the point as I'd expect. Essentially I think selecting a point should deselect the trace/shape.

packages/web-forms/src/components/common/map/MapBlock.vue
```diff
@@ -126,15 +145,36 @@ const discardSavedFeature = () => {
...
+
+				<Message
+					v-if="!disabled && mapHandler.canLongPressAndDrag()"
+					severity="contrast"
+					closable
+					size="small"
+					:class="{ 'map-message': true, 'above-secondary-controls': showSecondaryControls }"
+				>
 					<!-- TODO: translations -->
 					<span v-if="savedFeatureValue">Long press and drag move point</span>
```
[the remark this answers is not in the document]
latin-panda: I've changed it to `Press and drag to move a point`.

packages/web-forms/src/components/common/map/MapConfirm.vue
```diff
@@ -0,0 +1,59 @@
...
+			<strong v-if="drawFeatureType === DRAW_FEATURE_TYPES.SHAPE">Delete entire shape?</strong>
+			<strong v-if="drawFeatureType === DRAW_FEATURE_TYPES.TRACE">Delete entire trace?</strong>
+		</template>
+
+		<template #default>
+			<!-- TODO: translations -->
+			<p v-if="drawFeatureType === DRAW_FEATURE_TYPES.SHAPE">
+				Are you sure you want to delete this entire shape and start over?
+			</p>
+			<p v-if="drawFeatureType === DRAW_FEATURE_TYPES.TRACE">
```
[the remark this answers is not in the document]
latin-panda: Fixed

packages/web-forms/src/components/form-elements/input/InputGeotrace.vue
```diff
@@ -0,0 +1,21 @@
...
+
+<template>
+	<AsyncMap
+		:disabled="question.currentState.readonly"
+		draw-feature-type="trace"
+		:mode="MODES.DRAW"
+		:saved-feature-value="question.currentState.instanceValue"
+		@save="(value) => question.setValue(value ?? '')"
+	/>
+</template>
```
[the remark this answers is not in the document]
latin-panda: I have been uncertain about this since the project includes at least an explicit component for each question type. However, here I don't anticipate adding any additional distinct logic. I've named it InputGeoMultiPoint, open to suggestions.

packages/web-forms/src/components/common/map/vertex-geometry.ts
latin-panda: Fixed

packages/xforms-engine/src/lib/codecs/geolocation/Geotrace.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export class Geotrace extends Geolocation {
+	static parseStringToGeotrace(value: string): GeotraceRuntimeValue {
+		const parts = Geolocation.getSegments(value);
+		if (parts === null) {
+			return null;
+		}
+
+		const points = parts.map((point) => Geolocation.parseString(point)) as LocationPoint[];
+		if (points.some((p) => p === null) || points.length < 2 || Geolocation.isClosedShape(points)) {
```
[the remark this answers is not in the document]
latin-panda: Yes, at least for xforms, that's my understanding from [the specs ](https://getodk.github.io/xforms-spec/#data-type:geotrace)

> Semi-colon-separated list of at least 2 geopoints, where the last geopoint’s latitude and longitude is not equal to the first

packages/web-forms/src/components/common/map/MapStatusBar.vue
alyblenkin: I agree that it's confusing for users, and we should indicate the actual number of visible points rather than what's internally correct. I believe this is how it works with Collect as well. So in this case we would say 3 points saved.

> But it might be more useful to the user if we report in terms of shapes for geoshape, and points for geotrace. For example, the message for geoshape could be "no shape saved" or "1 shape saved" rather than "n points saved".

This is something we want to get feedback on because we display the points added in Collect too. So if it's not helpful we would want to likely remove it on both.

alyblenkin:
> I don't think we need to wait for Collect necessarily, unless you think it needs significant design? I'm not sure the tear drop anchor achieves my goal either because it's also likely to be obscured by my thumb, and it looks like that was designed for moving points rather than adding them.

Yah, I think this would need testing/iterating to figure out how to make it work for both adding and moving, but it's something we've noticed as a pattern with other geo apps clients have shared. 

> I was thinking a short vibration on selection would be simple and easily understood. Doing it in WF also makes it easy to collect user feedback for a potential implementation in Collect.

I assumed adding haptic feedback would take some time, so didn't think we needed it for the first iteration, but I agree it would add value! I also agree testing it WF first is a great.

packages/web-forms/src/components/common/map/vertex-geometry.ts
```diff
@@ -0,0 +1,230 @@
...
+	const [ax, ay] = pointA as [number, number];
+	const [bx, by] = pointB as [number, number];
+	return { deltaX: ax - bx, deltaY: ay - by };
+};
+
+/**
+ * Computes the squared Euclidean distance between two points using the Pythagorean theorem.
+ */
+const getDistanceBetweenPoints = (deltaX: number, deltaY: number) => {
+	// prettier-ignore
```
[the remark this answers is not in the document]
latin-panda: > ESLint: Replace `(deltaX·**·2)·+·(deltaY·**·2)` with `deltaX·**·2·+·deltaY·**·2` (prettier-vue/prettier)

It's the parentheses. Prettier seems opinionated about parentheses. I didn't find a setting to turn that off, and the settings for that in ESLint don't override Prettier :( 
In this case, I prefer to keep the parentheses for readability. I'll leave the prettier-ignore for now.

PR #587, opened 2025-12-14
fix(#586): defer parsing of translation elements

### LINKED ISSUE #586: Defer parsing of translations until needed
### Description

During form load we do some parsing of translation elements which can be deferred until its needed. This is particularly beneficial if the form has multiple languages because we should only do work on the language that's being used currently.

More information on Drive: https://docs.google.com/document/d/1lTDaNY2Qg9kbn1jf0zBKxy_tqEUrgSrrl3P_bZ4fxw0/edit

---

### Checklist
  - [x] No
  - [x] No
  - [x] No
  - [x] No
---

garethbowen:
Closes #586

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

- CI
- Performance testing: https://docs.google.com/document/d/1lTDaNY2Qg9kbn1jf0zBKxy_tqEUrgSrrl3P_bZ4fxw0/edit

Shouldn't

### DIFF
--- a/.changeset/thin-dogs-poke.md
+++ b/.changeset/thin-dogs-poke.md
@@ -1,0 +1,6 @@
+---
+'@getodk/xforms-engine': patch
+'@getodk/web-forms': patch
+---
+
+Defer evaluation of translation elements to improve performance


packages/xforms-engine/src/instance/text/markdownFormat.ts
```diff
@@ -208,7 +208,8 @@ function toOdkMarkdown(str: string): MarkdownNode[] {
 }
 
 export function format(chunks: readonly TextChunk[]): MarkdownNode[] {
-	outputStrings = new Map<string, string>();
 	const str = escapeEditableChunks(chunks);
-	return toOdkMarkdown(str);
+	const result = toOdkMarkdown(str);
+	outputStrings.clear();
```
garethbowen: By clearing after the fact the output strings aren't kept in memory longer than necessary.

Also it saves on creating a new map.

packages/xforms-engine/src/lib/reactivity/text/createTextRange.ts
```diff
@@ -17,57 +20,84 @@ interface ChunksAndMedia {
...
+		// only translations have 'nodes' chunks
+		return definition.chunks as Array<TextChunkExpression<'string'>>;
+	}
+	const itextId = context.evaluator.evaluateString(definition.chunks[0].toString()!, {
+		contextNode: context.contextNode,
+	});
+	const lang = context.getActiveLanguage();
+	const elem = definition.form.model.getItextElement(lang, itextId);
+	return elem ? generateChunksForTranslation(elem) : [];
+};
```
garethbowen: This code has been moved here from the parsing step. By only chunking the translations for the active language we save a bunch of work for languages that may never be loaded.

packages/xforms-engine/src/parse/model/TranslationDefinitionMap.ts
```diff
@@ -0,0 +1,23 @@
...
+export class TranslationDefinitionMap extends Map<string, Map<string, Element>> {
+	constructor(translationElements: readonly DOMItextTranslationElement[]) {
+		super(
+			translationElements.map((translationElement) => {
+				const lang = translationElement.getAttribute('lang');
+				return [lang, generateChunksForLanguage(translationElement)] as const;
+			})
+		);
+	}
+}
```
garethbowen: This file was renamed because now it's a lot simpler. I tried removing this altogether but it turned out that it was expensive to look the element up by ID so keeping the map sped up looking up the element which more than made up for the initial load.

PR #592, opened 2025-12-18
fix(#356): clean up caches and dispose of scopes on form reset

### LINKED ISSUE #356: Form instance clean up
Noted as a TODO in #355. The intent is primarily to be conservative about our use of _user device resources_ (CPU and memory, anything with side effects like network or sensor access).

As a heuristic, we want an instance that goes out of scope to be garbage collected.

At minimum to accomplish that, we will need to break stateful cyclic references, many of which we establish indirectly. Top of mind: reactive computations, and scoped ownership of same. This will primarily involve:

- correcting how we establish _hierarchy_ of reactive scopes (we don’t nest them on creation now, despite designing for that; we probably want to revise the signature of `createReactiveScope` to make this mistake less likely to reoccur)
- exposing a mechanism to invoke disposal
- calling that at appropriate points where disposal is desired (and knowing where disposal is desired is crucial to that!)

We may also want to tear down instance hierarchy (i.e. remove children recursively), and overwrite any other stateful values (stored, not necessarily derived) which reference others (likely candidates are complex `ValueNode` values).

We should be prepared to know how to **measure** whether clean up is effective. This will likely involve use of browser’s profilers, and may also involve use of logging with `FinalizationRegistry`, but that might be a good last resort. We don’t necessarily need to watch this super closely once we’ve addressed many of the most obvious details above, but it will be valuable to know how to diagnose pathological cases when they’re reported by real users.

garethbowen:
Closes #356

- [x] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

Memory profiling as documented here: https://docs.google.com/document/d/1RqP82e90DLBGtltVGygkelnPJ62sz7ZWZ549cu9zdTg/edit?tab=t.0#heading=h.j2f2t9esrqzs

A workaround was to force a reload of the page, but that slows users down because you need to parse the form again.

### DIFF
--- a/packages/web-forms/src/components/common/map/useMapBlock.ts
+++ b/packages/web-forms/src/components/common/map/useMapBlock.ts
@@ -348,6 +348,7 @@ export function useMapBlock(config: MapBlockConfig, events: MapBlockEvents) {
 		mapViewControls = undefined;
 		mapInteractions?.teardownMap();
 		mapInteractions = undefined;
+		mapFeatures = undefined;
 
 		iconImageCache.clear();
 		if (mapInstance) {
--- a/packages/xforms-engine/src/entrypoints/FormResult/BaseInstantiableFormResult.ts
+++ b/packages/xforms-engine/src/entrypoints/FormResult/BaseInstantiableFormResult.ts
@@ -64,18 +64,9 @@ export abstract class BaseInstantiableFormResult<
 		};
 
 		this.resetInstance = (instanceConfig: FormInstanceConfig = {}) => {
-			this.assertInstantiable();
-
 			instanceOptions.scope.dispose();
 			instanceOptions.scope = createPotentiallyClientOwnedReactiveScope();
-			// const resetInstanceOptions: BasePrimaryInstanceOptions = { ...instanceOptions, scope };
-
-			return new FormInstance(this, {
-				mode: 'reset',
-				instanceOptions: instanceOptions,
-				initialState: null,
-				instanceConfig,
-			});
+			return this.createInstance(instanceConfig);
 		};
 
 		this.editInstance = async (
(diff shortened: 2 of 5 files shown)


packages/xforms-engine/src/entrypoints/FormResult/BaseInstantiableFormResult.ts
```diff
@@ -60,6 +63,21 @@ export abstract class BaseInstantiableFormResult<
 			});
 		};
 
+		this.resetInstance = (instanceConfig: FormInstanceConfig = {}) => {
+			this.assertInstantiable();
+
+			instanceOptions.scope.dispose();
+			instanceOptions.scope = createPotentiallyClientOwnedReactiveScope();
+			// const resetInstanceOptions: BasePrimaryInstanceOptions = { ...instanceOptions, scope };
```
latin-panda: Can this comment be removed?

packages/web-forms/src/components/common/map/useMapBlock.ts
```diff
@@ -256,7 +256,33 @@ export function useMapBlock(mode: Mode, events: { onFeaturePlacement: () => void
 
 	const teardownMap = () => {
 		mapViewControls?.stopWatchingCurrentLocation();
+		mapViewControls = undefined;
 		mapInteractions?.teardownMap();
+		mapInteractions = undefined;
+
```
latin-panda: Maybe for consistency, it can also clean useMapFeatures, in case it has something that should be cleared at some point.
```suggestion
useMapFeatures = undefined;
```

latin-panda (approved):
Very interesting! Thank you!
Could you link this to the Central channel? They can benefit from the map teardown logic, since they have their own implementation of maps for Central.

Have you tested in Central?

I left some existential feedback for xform-engine, but it seems out of scope for this work and can be addressed in a separate ticket, so I'm approving this PR now to unblock.

garethbowen:
> Could you link this to the Central channel? They can benefit from the map teardown logic, since they have their own implementation of maps for Central.

Will do.

> Have you tested in Central?

Tested. Leaked fixed. Functionality works as expected.

PR #594, opened 2025-12-18
feat(#507): add setgeopoint action

### LINKED ISSUE #507: Add setgeopoint action
~Blocked by https://github.com/getodk/web-forms/issues/303~

### Description
The `setgeopoint` action can be triggered by the `odk-instance-first-load`, `odk-new-repeat`, `xforms-value-changed` events. It captures the best accuracy point in a 20s window after the event was fired.

ODK XForms spec: https://getodk.github.io/xforms-spec/#events

User-facing docs:
* [start-geopoint](https://docs.getodk.org/form-question-types/#metadata-start-geopoint) 

---

### Checklist
         Maybe. We could add some messaging around providing permissions. In Collect there's a notice that the location is being captured in the background
  - [ ] No
  - [x] Yes. Specify the endpoints, methods, or data structures needed.
         The client needs to be able to provide an implementation for getting user location and may need to know about the action in order to provide messaging.
  - [ ] No
  - [x] Yes. Outline the state flow requirements.
         The client will need to know whether permissions have been granted
  - [ ] No
  - [x] Yes. Add the test plan as a comment in this task or attach the document.
  - [ ] No
---

### User Stories

- **As a** form designer,
  **I want** to capture form users' location in the background,
  **so that** I can audit their behavior or get locations of points of interest without requiring user intervention.

---

### Design

- **Figma**: [Insert URL]
- **Design Docs**: [Insert URL]

---

### Dependencies

-

---

### Additional Notes

-

---

### Related Issues

-

latin-panda:
Closes #507

ODK XForms spec: https://getodk.github.io/xforms-spec/#events
User-facing docs: [start-geopoint](https://docs.getodk.org/form-question-types/#metadata-start-geopoint) (odk:setgeopoint with odk-instance-first-load)

These questions were already answered by the team; see the summary [here](https://github.com/getodk/web-forms/pull/594#issuecomment-3746449636).

1. Does it need to show an indicator just like Collect does? (I think we decided not add it, but double-checking)
    > Any time a survey with a start-geopoint question is opened in Collect, the enumerator will see a warning that the form tracks device location .... A location icon will be displayed in the Android status bar while the geolocation is being requested by Collect.

2. The `odk-new-repeat` is fired when a new repeat instance is created. At that moment, geolocation is requested once. Subsequent repeat instances reuse this same geolocation reading, so all instances end up with the same location value. That's expected, correct?

3. The `background-geopoint` is an `odk:setgeopoint` with `xforms-value-changed`, that, as per the specs: 
   > `lets you automatically capture a single geolocation in geopoint format when the user changes the value of another question`. 

   3.1  The location is requested once, and subsequent value changes reuse the same geolocation reading (or empty string if no permissions given) for all questions. Is that correct? This is a lighter and less complex option than requesting the location every time a question value changes.

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [ ] Safari (iOS)
- [x] Chrome for Android
- [ ] Not applicable

<details><summary>Displays error message when it cannot access the geolocation</summary>

</details>

<details><summary>Displays multiple messages in the top error banner</summary>

</details>

<details><summary>Tests all events except new repeat instance</summary>

https://github.com/user-attachments/assets/b947e648-0662-424d-b2d9-2ad6bcfb21c9

</details>

<details><summary>Tests new repeat instance event</summary>

https://github.com/user-attachments/assets/062a5a1f-4773-463d-9da0-8a1e81a68778

</details>

Reuses the existing action structure, sets the value without re-evaluating the expression (since it comes from the geolocation provider), and reuses geopoint codecs for validation.

Existing features will continue to work as usual with no changes. 
Forms with setgeopoint will now request geolocation in the background, which may affect low-end phones by increasing resource usage (battery, CPU), as expected. The geolocation request stops after 20 seconds, in accordance with the specs. 

There are 3 forms in the common package to test the feature. I also have a form that includes all events in one form:

<details><summary>Form: setgeopoint with events</summary>

```xml
<?xml version="1.0" encoding="utf-8"?>
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa" xmlns:odk="http://www.opendatakit.org/xforms">
  <h:head>
    <h:title>combined-setgeopoint-test</h:title>
    <model>
      <instance>
        <data id="combined-setgeopoint-test">
          <required_text />
          <location_first_load />
          <location_instance_load />
          <trigger_text />
          <location_value_changed />
          <setvalue_target />
          <person jr:template="">
            <age />
            <location_new_repeat />
          </person>
        </data>
      </instance>
      <bind nodeset="/data/required_text" type="string" required="true()" />
      <bind nodeset="/data/location_first_load" type="string" readonly="true()" />
      <bind nodeset="/data/location_instance_load" type="string" readonly="true()" />
      <bind nodeset="/data/trigger_text" type="string" />
      <bind nodeset="/data/location_value_changed" type="string" readonly="true()" />
      <bind nodeset="/data/setvalue_target" type="string" readonly="true()" />
      <bind nodeset="/data/person/age" type="integer" />
      <bind nodeset="/data/person/location_new_repeat" type="string" readonly="true()" />
      <odk:setgeopoint ref="/data/location_first_load" event="odk-instance-first-load" />
      <odk:setgeopoint ref="/data/location_instance_load" event="odk-instance-load" />
    </model>
  </h:head>
  <h:body>
    <input ref="/data/required_text">
      <label>Required Text</label>
    </input>
    <input ref="/data/location_first_load">
      <label>Location (odk-instance-first-load)</label>
    </input>
    <input ref="/data/location_instance_load">
      <label>Location (odk-instance-load)</label>
    </input>
    <input ref="/data/trigger_text">
      <label>Trigger Text (change to trigger value-changed)</label>
      <odk:setgeopoint ref="/data/location_value_changed" event="xforms-value-changed" />
      <odk:setvalue ref="/data/setvalue_target" event="xforms-value-changed" value="/data/trigger_text" />
    </input>
    <input ref="/data/location_value_changed">
      <label>Location (xforms-value-changed)</label>
    </input>
    <input ref="/data/setvalue_target">
      <label>Setvalue Target</label>
    </input>
    <group ref="/data/person">
      <label>Person Repeat Group</label>
      <repeat nodeset="/data/person">
        <input ref="/data/person/age">
          <label>Age</label>
        </input>
        <odk:setgeopoint event="odk-new-repeat" ref="/data/person/location_new_repeat" />
        <input ref="/data/person/location_new_repeat">
          <label>Location (odk-new-repeat)</label>
        </input>
      </repeat>
    </group>
  </h:body>
</h:html>

```

</details>

### What's changed
- Introduces a geolocation provider interface in the engine and enables support for consuming it in the primary instance, which is later used for dispatching geolocation actions in createInstanceValueState.ts.
- Reuses most of the action structure implemented for setValue
- Adds scenario tests 
- Introduces a singleton geolocation service that operates independently without affecting other question types requesting location permission. It includes a tear-down method to free memory.

### DIFF
--- a/packages/web-forms/src/components/OdkWebForm.vue
+++ b/packages/web-forms/src/components/OdkWebForm.vue
@@ -173,7 +173,7 @@ const getLocation = async (): Promise<string> => {
 	let point = '';
 	try {
 		geolocationErrorMessage.value = '';
-		point = await geolocationService.getBestGeopoint();
+		point = await geolocationService.getCurrentPosition();
 	} catch (error) {
 		// eslint-disable-next-line no-console -- Skip silently to match Collect behaviour.
 		console.warn('Error occurred while retrieving background location.', error);
--- a/packages/web-forms/src/lib/services/geolocationService.ts
+++ b/packages/web-forms/src/lib/services/geolocationService.ts
@@ -16,7 +16,7 @@ class GeolocationService {
 	 * @param timeoutSeconds - The maximum time allowed to obtain a location before a TIMEOUT error.
 	 * @returns A promise resolving to an ODK geopoint string.
 	 */
-	public async getBestGeopoint(timeoutSeconds = 20): Promise<string> {
+	public getCurrentPosition(timeoutSeconds = 20): Promise<string> {
 		if (this.activePromise) {
 			return this.activePromise;
 		}


packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
```diff
@@ -283,7 +283,16 @@ export const createInstanceValueState = (context: ValueContext): InstanceValueSt
 
 		const action = context.definition.model.actions.get(context.contextReference());
 		if (action) {
-			registerAction(context, setValue, action);
+			if (action.element.nodeName === 'odk:setgeopoint') {
+				context.parent.rootDocument.getBackgroundGeopoint()?.then((value: string) => {
```
latin-panda: The primary instance is the parent, not the context itself :(

latin-panda:
@garethbowen This code isn't pretty and still experimental, but I wanted to get a version working first and get your thoughts from the engine side.

In the screenshot, the xform-engine is working; it sets the value when the event runs, and the value comes from a promise. 

Screenshot 2025-12-19 at 2 15 04 AM

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
```diff
@@ -283,7 +283,16 @@ export const createInstanceValueState = (context: ValueContext): InstanceValueSt
...
+					if (value) {
+						action.buildComputation(`"${value}"`);
+						registerAction(context, setValue, action);
```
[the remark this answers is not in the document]
latin-panda: Thanks! I've fixed it

packages/scenario/test/actions-events/set-geopoint-action.test.ts
```diff
@@ -1,150 +1,188 @@
...
-			try {
-				await Scenario.init(r('setgeopoint-action-bad-namespace.xml'));
-			} catch (error) {
-				caught = error;
-			}
-
-			expect(caught).toBeInstanceOf(Error);
-		});
+
+describe('odk:setgeopoint action', () => {
```
garethbowen: Thanks for renaming the tests! If these tests replace the ones ported from openrosa please include a link to the original so we can easily tell we haven't lost coverage.

packages/scenario/test/actions-events/set-geopoint-action.test.ts
```diff
@@ -1,150 +1,188 @@
...
+				),
+				body(input('/data/source', setgeopoint('xforms-value-changed', '/data/destination')))
+			),
+			geolocationProvider
+		);
+
+		scenario.answer('/data/source', 22);
+		expect(scenario.answerOf('/data/destination')).toEqualAnswer(
+			stringAnswer('38.295 21.7567 110 5')
+		);
```
garethbowen: As with the `odk-new-repeat` test, can you modify the provider to return a different value on the second call, and ensure that it changes when the source is updated?

packages/web-forms/src/components/OdkWebForm.vue
```diff
@@ -249,9 +281,18 @@ watchEffect(() => {
...
+				class="form-error-message"
+				:closable="true"
+				@close="floatingErrorActive = false"
+			>
 				<IconSVG name="mdiAlertCircleOutline" variant="error" />
-				<span>{{ validationErrorMessage }}</span>
+				<span class="form-error-text-wrap">
+					<span v-if="validationErrorMessage?.length">{{ validationErrorMessage }}</span>
+					<span v-if="geolocationErrorMessage?.length">{{ geolocationErrorMessage }}</span>
+				</span>
```
garethbowen: Using block level elements would be more semantic. I think an `<ul>` would be ideal, but `<p>` elements would work too.

packages/web-forms/src/lib/services/geolocationService.ts
```diff
@@ -0,0 +1,88 @@
...
+			if (!navigator.geolocation) {
+				this.complete(resolve, reject, 'Geolocation is not supported by this browser.');
+			}
+
+			this.watcherId = navigator.geolocation.watchPosition(
+				(point) => {
+					this.bestPoint =
+						!this.bestPoint || point.coords.accuracy < this.bestPoint.coords.accuracy
+							? point
+							: this.bestPoint;
```
garethbowen: This is a bit much for a ternary IMO. Can we use a basic if?


```suggestion
					if (!this.bestPoint || point.coords.accuracy < this.bestPoint.coords.accuracy) {
						this.bestPoint = point;
					}
```

packages/web-forms/src/lib/services/geolocationService.ts
```diff
@@ -0,0 +1,88 @@
...
+		return `${latitude} ${longitude} ${altitude} ${accuracy}`;
+	}
+
+	private complete(
+		resolveFn: (val: string) => void,
+		rejectFn: (err: Error) => void,
+		errorMessage: string
+	): void {
+		if (this.bestPoint) {
+			resolveFn?.(this.formatGeopoint(this.bestPoint));
```
garethbowen: When could `resolveFn` be null/undefined? As far as I can see it's only called from inside the Promise which holds on to the reference and passes it in.

packages/web-forms/src/lib/services/geolocationService.ts
```diff
@@ -0,0 +1,88 @@
...
+			setTimeout(() => {
+				this.complete(resolve, reject, 'No geolocation readings received within the time window.');
+			}, timeoutSeconds * 1000);
+		});
+		return this.activePromise;
+	}
+
+	private formatGeopoint(position: GeolocationPosition): string {
+		const { latitude, longitude, accuracy } = position.coords;
+		const altitude = position.coords.altitude ?? 0;
```
garethbowen: I think you can do this...


```suggestion
		const { latitude, longitude, accuracy, altitude=0 } = position.coords;
```

packages/web-forms/src/lib/services/geolocationService.ts
```diff
@@ -0,0 +1,88 @@
...
+
+	private complete(
+		resolveFn: (val: string) => void,
+		rejectFn: (err: Error) => void,
+		errorMessage: string
+	): void {
+		if (this.bestPoint) {
+			resolveFn?.(this.formatGeopoint(this.bestPoint));
+		} else {
+			rejectFn?.(new Error(errorMessage));
```
garethbowen: Ditto.

packages/xforms-engine/src/instance/PrimaryInstance.ts
```diff
@@ -289,4 +295,22 @@ export class PrimaryInstance<
...
 		return Promise.resolve(result);
 	}
+
+	async getBackgroundGeopoint() {
+		if (this.backgroundGeopoint != null) {
+			return await this.backgroundGeopoint;
+		}
+
+		try {
+			this.backgroundGeopoint = this.geolocationProvider?.getLocation() ?? Promise.resolve('');
```
garethbowen: I think we can return early in the case where the geolocation provider is not given.


```suggestion
		if (!this.geolocationProvider) {
			return '';
		}

		try {
			this.backgroundGeopoint = this.geolocationProvider.getLocation();
```

garethbowen (changes requested):
Fantastic. Looks like a nice generalisation of the setvalue code!

I haven't tried running it yet (hopefully will do that later today), but thought I'd get this feedback to you ASAP.

packages/scenario/test/actions-events/set-geopoint-action.test.ts
```diff
@@ -1,150 +1,188 @@
...
+		await flushPromises();
+		expect(scenario.answerOf('/data/person[1]/location')).toEqualAnswer(
+			stringAnswer('38.295 21.7567 110 5')
+		);
+
+		scenario.createNewRepeat('/data/person');
+		await flushPromises();
+		expect(scenario.answerOf('/data/person[2]/location')).toEqualAnswer(
+			stringAnswer('38.295 21.7567 110 5')
+		);
```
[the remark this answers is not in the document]
garethbowen: Testing this in Firefox it looks like it only requests the location the first time because the second repeat has the identical value immediately. Is this the desired behaviour? I assumed that the only reason to use the `odk-new-repeat` event would be because you wanted to gather a new location, not just duplicate the one that's already there. Might be worth checking the spec...

packages/web-forms/src/lib/services/geolocationService.ts
```diff
@@ -0,0 +1,88 @@
+/**
+ * Singleton for geolocation. Ensures a single active watchPosition process.
+ */
+class GeolocationService {
+	private static instance: GeolocationService | null = null;
+	private watcherId: number | null = null;
+	private activePromise: Promise<string> | null = null;
+	private bestPoint: GeolocationPosition | undefined = undefined;
+	private options: PositionOptions = { enableHighAccuracy: true, maximumAge: 0 };
```
garethbowen: I can't find anywhere `maximumAge` is used - should this be removed or implemented?

packages/web-forms/src/components/OdkWebForm.vue
latin-panda: Done

packages/web-forms/src/lib/services/geolocationService.ts
latin-panda: It doesn't catch the case when altitude is null ([mdn ref](https://developer.mozilla.org/en-US/docs/Web/API/GeolocationCoordinates/altitude))

latin-panda:
@garethbowen This is ready for review!

packages/scenario/test/actions-events/set-geopoint-action.test.ts
```diff
@@ -1,150 +1,226 @@
...
-
-			try {
-				await Scenario.init(r('setgeopoint-action-bad-namespace.xml'));
-			} catch (error) {
-				caught = error;
-			}
 
-			expect(caught).toBeInstanceOf(Error);
-		});
+// Original JavaRosa Tests: https://github.com/getodk/javarosa/blob/master/src/test/java/org/javarosa/core/model/actions/SetGeopointActionTest.java
```
garethbowen: Consider using a permalink so it doesn't 404 if the file gets moved...


```suggestion
// Original JavaRosa Tests: https://github.com/getodk/javarosa/blob/2e10861b2141f4ce40ac64d78e38e8548a01f7f1/src/test/java/org/javarosa/core/model/actions/SetGeopointActionTest.java
```

packages/web-forms/src/components/OdkWebForm.vue
```diff
@@ -186,6 +219,11 @@ const handleSubmit = (currentState: FormStateSuccessResult) => {
 	const { root } = currentState;
 
 	if (root.validationState.violations.length === 0) {
+		// TODO: translations
+		geolocationService.resolveNow(
+			new Error('No geolocation readings received within the time window.')
```
garethbowen: Passing in the Error feels strange because it looks to the reader as though something has gone wrong, but we don't know if it has yet.

One option would be to change the API so the Error is optional and this message is the default if bestPoint is undefined.

garethbowen (approved):
Great work!

I haven't tested exhaustively and there are plenty of edge cases so should go through QA on real devices before merging.

I've left one nit pick and one API suggestion but nothing blocking.

Regarding Promises vs EventEmitters... Your solution solves my main problem, so I think it's good to go. However, EEs would be a little more powerful because they would set the `value` in the form more than once. A couple of examples I can think of off the top of my head are...

1. Validations that require a certain level of accuracy before allowing the user to submit. With the current implementation they would have to wait 20s for the timeout to set the location to pass the validation.
2. Calculations which could update a value based on the location, for example, if you're in a district, show only options available in that district.

I don't know if these are valid use cases or not, and it might be a question for @lognaturel

lognaturel:
Wonderful to see this getting close!

> would set the value in the form more than once

The main argument I currently have in favor of writing to the model each time a higher accuracy point is received is that it would be ideal to capture a location in case of accidental browser page close / navigating away. This only starts becoming relevant when we save drafts. For example:
* User starts new form
* `setgeopoint` on `odk-instance-first-load` starts capturing location
* User fills out a couple of fields
* User accidentally closes the form at second 11
* User re-opens form and restores from draft
* `setgeopoint` doesn't fire again because the `odk-instance-first-load` event isn't emitted 
* Ideally the highest-accuracy point captured in the initial 11 seconds is in the restored instance

We could group this refinement with draft saving?

> Validations that require a certain level of accuracy before allowing the user to submit. With the current implementation they would have to wait 20s for the timeout to set the location to pass the validation.

This should indeed be possible to express. I'd do it with a note that says "location not high enough accuracy to submit" and a relevant condition that checks the accuracy of the automatically-captured point. But since there's a human in the loop, I'd tend to reach for `geopoint` with no appearance instead so that the user can get feedback and continue trying until they get a high enough accuracy point. I think of this action as a way to capture location more as metadata to do things like check the data collector was roughly in the expected place.

> Calculations which could update a value based on the location, for example, if you're in a district, show only options available in that district.

Similar to above! This is possible. It could be really weird, though, because the list might change dramatically as the device gets higher and higher accuracy points. I'd tend to have the user first capture a geopoint with user intervention and then use it as a filter.

latin-panda:
> The main argument I currently have in favor of writing to the model each time a higher accuracy point is received is that it would be ideal to capture a location in case of accidental browser page close / navigating away. This only starts becoming relevant when we save drafts.

Good case! Thanks for the clear explanation! It could be a good candidate to support writing the model each time there's a better value. It does raise some concerns worth discussing:

- Frequent updates could trigger undesired recalcs, and it feels unexpected on `odk-instance-first-load`. It can be inefficient for forms not designed for that.

- It risks poor-quality data in drafts, like early low accuracy points that users/project managers might miss or can't easily fix.

There are some options to mitigate this 
- Drop the 20s mark to 10-15s for better balance.
- Add a visual indicator ("Getting location. Don't close the tab!") 
- On draft open, if the value's unset, run the `odk-instance-first-load` event again and probably ensuring good data. Likely the user will spend more than 20s filling up the form. 

> We could group this refinement with draft saving?

It makes sense, there are many cases there for draft that we need to design.

lognaturel:
> If using EventEmitters updates the form too frequently we can introduce rate limiting in the service, and other clients can limit to emitting once if they want a promise like operation, without requiring an API change.

This sounds right to me. The spec limits the total watch time to 20s because it's expected to be heavy. An implementation is free to do rate limiting if needed but I would say `watchPosition` already inherently does that. That is, the frequency at which events fire is already determined by how much the device is actually moving.

> I suggest we switch the implementation to use getCurrentPosition rather than watchPosition

I agree `getCurrentPosition` with a `timeout` of 20s and `enableHighAccuracy` (which I think we should definite set either way) could do the trick. The downside is that then we wouldn't have the option of writing interim locations for drafts or other considerations like GPS losing lock (though probably `getCurrentPosition` accounts for that and would still return the best location so far, not sure).

PR #633, opened 2026-01-27
chore: sync changelogs and package versions

latin-panda:

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

### What's changed
- Sync changelog and package.json files 
- Updates yarn lock
- Adds patch release steps


garethbowen (approved):
Thanks. Held off on this till now to get the release out first, which means I've even tested out your release process!

If we go with CD then this will be obsolete but we may as well merge it now because it might be useful before that change lands.

PR #640, opened 2026-01-30
feat(#617): improve map full-screen UX on mobile

### LINKED ISSUE #617: Make full-screen map the default editing mode for mobile users
### Description
Transition the map view to a full-screen layout on mobile devices. This includes fixing existing iPhone-specific rendering issues in full-screen mode and experimenting with JavaScript-based mobile detection (avoiding user agent strings) to enable the experience on browser resize. 

Must resolve the challenge of the full-screen map becoming scrollable (advanced panel is open) to prevent adding points accidentally 

---

### Checklist
  - [x] Yes. Define how overlays and navigation elements sit on top of the full-screen map.
  - [ ] No
  - [x] No
  - [ ] Yes. Outline how the map enters and exits full-screen mode during navigation.
  - [x] No
  - [ ] Yes.
  - [x] No. Reuse existing test plans

---

### User Stories
- **As a** mobile user,  
  **I want** the map to utilize the entire screen,  
  **so that** I have a clearer view of my surroundings and adding points are intentional (no my mistake when scrolling the page)

---

### Design
- **Figma**: [Insert URL]
- **Design Docs**: N/A

---

### Dependencies
- [Must resolve iPhone-specific full-screen bug](https://github.com/getodk/web-forms/issues/589)

---

### Additional Notes
- **Implementation**: We usually prefer mobile detection using responsive CSS because it's lightweight and reliable. For this feature, I'll experiment with JS to find the best option (not using user agent). I already have a few in mind.
- **Priority**: iPhone full-screen fixes must be addressed first.

---

latin-panda:
Closes #617

---- 

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [ ] Safari (iOS)
- [x] Chrome for Android
- [ ] Not applicable

<details><summary>Desktop: full screen maps</summary>

https://github.com/user-attachments/assets/a83e7a7b-addf-4fdf-be49-e20c143a0f23

</details>

<details><summary>Mobile: full screen maps</summary>

https://github.com/user-attachments/assets/fca2eed5-24d2-43b7-9fc2-2531580dc715

</details>

Clicking anywhere on the map, including buttons like zoom and find current location, should trigger full screen. A transparent overlay captures clicks and initiates full-screen mode. This approach avoids complicating current map interactions and keeps the full-screen feature contained in the view layer (Vue component).  

Use forms from demo > geo location 

### What's changed
- Inline-map is reduced in height for mobile
- Hide all the buttons on the map except the full-screen button when it's on mobile
- The full-screen button's icon is replaced by the expand and collapse icons for desktop and mobile
- The Location denied error is shown in full-screen mode and it has a close button. 
- For geopoint with "maps" appearance - the overlay is shown, if point is removed, it display a close "X" button

### DIFF
--- a/packages/web-forms/src/components/common/map/MapBlock.vue
+++ b/packages/web-forms/src/components/common/map/MapBlock.vue
@@ -251,7 +251,7 @@ const toggleFullScreen = async () => {
 					severity="contrast"
 					closable
 					size="small"
-					:class="{ 'map-message': true, 'above-secondary-controls': showSecondaryControls }"
+					:class="{ 'map-message': true, 'above-secondary-controls': isFullScreen && showSecondaryControls }"
 				>
 					<span v-if="pointPlaced">{{ instructionMessage.placed }}</span>
 					<span v-else>{{ instructionMessage.default }}</span>
@@ -264,6 +264,7 @@ const toggleFullScreen = async () => {
 				:can-view-details="mapHandler.canViewProperties()"
 				:single-feature-type="singleFeatureType"
 				:is-capturing="mapHandler.currentState.value === STATES.CAPTURING"
+				:is-full-screen="isFullScreen"
 				:saved-feature-value="savedFeatureValue"
 				:selected-vertex="selectedVertex"
 				class="map-status-bar-component"
--- a/packages/web-forms/src/components/common/map/MapStatusBar.vue
+++ b/packages/web-forms/src/components/common/map/MapStatusBar.vue
@@ -24,6 +24,7 @@ const props = defineProps<{
 	selectedVertex: Coordinate | undefined;
 	singleFeatureType?: SingleFeatureType;
 	isCapturing: boolean;
+	isFullScreen: boolean;
 	canRemove: boolean;
 	canViewDetails: boolean;
 	canOpenAdvancedPanel: boolean;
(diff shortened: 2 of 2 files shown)


latin-panda:
Aly's feedback is resolved too:

> -  I removed the advanced button on the preview map in the designs because the fields are disabled anyway. They can paste the location data, but that’s more of a secondary action.
> - The snackbar is also floating in the preview map, likely because the other actions are on the left once you go into full screen mode. Can we anchor the snackbar to the bottom? Not a big deal if not.

<details><summary>Fixed, video attached</summary>

https://github.com/user-attachments/assets/d3a5f01c-8dcf-4d2a-9865-f96b8449d4ce

</details>

garethbowen:
> > Something weird happens with the get location now... In inline mode when I hit get location it finds where I am, zooms in, and then zooms way back out to the full world view again. It only happens on the first load - if I remove the point and do it again it works as expected.

> I've tried Chrome, Safari, and Firefox in inline and full-screen modes. Do you remember any other previous steps?
If this is still an issue, it seems unrelated to full-screen. I suggest addressing it in another ticket.

Ooooooh! Turns out this is a NZ only issue! It also happens in Chrome which is handy because it's easier to mock the location in Chrome. If you set your location to return Auckland (lat: -36.8509, long: 174.7645) then it happens. If I select Berlin then it doesn't.

I don't know if this is on purpose but you can't scroll past the international date line, so it could be that NZ is too close to the edge that it aborts for some reason.

Anyway, definitely not a blocker!

PR #649, opened 2026-02-12
fix(#648): fixes xforms-value-changed firing when editing submission

### LINKED ISSUE #648: Editing a submission with an `xforms-value-changed` action fires update on load
### Expected Behaviour

The xforms-value-changed event should only fire when the value has been changed after the form has loaded. So when loading a form for editing a submission which already has values set, the action should not fire.

### Observed Behaviour

The action fires on load and updates the value.

### Reproduction Steps

Form (XML or XLSX file):

```xml
<?xml version="1.0"?>
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:odk="http://www.opendatakit.org/xforms" xmlns:orx="http://openrosa.org/xforms">
  <h:head>
    <h:title>Value changed event</h:title>
    <model>
      <instance>
        <data id="xforms-value-changed-event">
          <source/>
          <destination/>
          <meta>
            <instanceID />
          </meta>
        </data>
      </instance>
      <bind nodeset="/data/destination" type="dateTime"/>
      <bind jr:preload="uid" nodeset="/data/meta/instanceID" readonly="true()" type="string"/>
    </model>
  </h:head>
  <h:body>
    <input ref="/data/source">
      <setvalue event="xforms-value-changed" ref="/data/destination" value="now()"/>
    </input>
  </h:body>
</h:html>
```

Steps:
1. Load the form above
2. Update the source field
3. Submit the form above and see that the destination field is filled in with the current time
4. Edit the submission and submit again without changing the source field value
5. Notice that the destination field date is updated

garethbowen:
Closes #648

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

### DIFF
--- a/packages/scenario/src/client/init.ts
+++ b/packages/scenario/src/client/init.ts
@@ -1,6 +1,5 @@
 import type { JRResourceService } from '@getodk/common/jr-resources/JRResourceService.ts';
 import type {
-	EditFormInstanceInput,
 	FormResource,
 	GeolocationProvider,
 	InstanceAttachmentsConfig,
@@ -10,12 +9,17 @@ import type {
 	MissingResourceBehavior,
 	OpaqueReactiveObjectFactory,
 	PreloadProperties,
+	ResolvedFormInstanceInputType,
 	RootNode,
 } from '@getodk/xforms-engine';
 import { createInstance, editInstance } from '@getodk/xforms-engine';
 import type { Owner } from 'solid-js';
 import { createRoot } from 'solid-js';
 import { getAssertedOwner, runInSolidScope } from './solid-helpers.ts';
+
+import { constants, type InstanceData } from '@getodk/xforms-engine';
+
+const { INSTANCE_FILE_NAME, INSTANCE_FILE_TYPE } = constants;
 
 /**
  * @todo Currently we stub resource fetching. We can address this as needed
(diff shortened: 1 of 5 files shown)


packages/scenario/src/client/init.ts
```diff
@@ -73,7 +77,16 @@ export const initializeTestForm = async (
...
+				const instanceFile = new File([options.editInstance], INSTANCE_FILE_NAME, {
+					type: INSTANCE_FILE_TYPE,
+				});
+				const instanceData = new FormData();
+				instanceData.set(INSTANCE_FILE_NAME, instanceFile);
+				const instance = {
+					inputType: 'FORM_INSTANCE_INPUT_RESOLVED' as ResolvedFormInstanceInputType,
+					data: [instanceData as InstanceData] as const,
+				};
+				return editInstance(formResource, instance, initOptions);
```
garethbowen: This was a refactor to make it easier to test edits - now you just pass in the xml string.

latin-panda:
I'll review this and test it first thing tomorrow

latin-panda (approved):
LGTM! Thanks!

PR #651, opened 2026-02-13
fix(#650): allows multiple setvalue actions to reference the same element

### LINKED ISSUE #650: Allow for multiple actions that reference the same element
### Expected Behaviour

Multiple setvalue actions with the same ref should be allowed to be defined in a form, for example, if they're triggered on different events.

### Observed Behaviour

The last action is the only action that is registered.

### Reproduction Steps

Form (XML or XLSX file): 

[multiple-setvalue.xml](https://github.com/user-attachments/files/25276622/multiple-setvalue.xml)

Steps:
1. Load the form above
2. Add a member repeat
3. Submit
4. Check the xml - both members should have a uuid but actually only the second one does.

garethbowen:
Closes #650

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

### DIFF
--- a/packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
+++ b/packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
@@ -347,11 +347,7 @@ export const createInstanceValueState = (context: ValueContext): InstanceValueSt
 		}
 
 		const actions = context.definition.model.actions.get(context.contextReference());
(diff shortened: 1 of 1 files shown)


latin-panda:
I'll review this and test it first thing tomorrow

PR #657, opened 2026-02-17
fix: improves display of map status and fit to feature 

latin-panda:

- [X] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

Manual testing and e2e tests passing

### What's changed
- The status message `Capturing location...` was wrapping instead of using the available space when loading on mobile devices.
- The advanced panel wasn't closing when exiting full-screen mode. This is particularly inconvenient on mobile devices, where we hide the advance button. 
- The "fit view to feature" was using the current zoom. If the zoom level is 1 but the feature is too large, it doesn't fit within the map viewport. It now calculates the feature's zoom level.


latin-panda:
Now that all enhancements have been implemented, I ran regression tests and found 3 small issues, which I've fixed. 
@garethbowen could you please review when you have time?

garethbowen (approved):
One code suggestion inline. Approved to unblock.

PR #661, opened 2026-02-18
fix(#659): Fix micro-panning interference, optimize full-screen UI and location tracking

### LINKED ISSUE #659: Fix micro-panning interference, optimize full-screen UI and location tracking
There are some issues reported in [the forum](https://forum.getodk.org/t/let-us-know-what-you-think-of-maps-in-web-forms/57437/10) regarding map interactions.

On mobile devices, users frequently experience issues where tapping the map fails to add a vertex. This occurs because OpenLayers' default moveTolerance is only 1px. If a user's finger moves even slightly during a tap (which is common on touchscreens), the map registers a "drag/pan" event instead of a "click," causing the add vertex interaction to be ignored.

A race condition occurs (more noticeable in mobile view) when tapping the remove button on a geopoint with "maps", causing the overlay to flash while retrieving the location. 

On mobile, the current location stays active after exiting full-screen mode, which isn't very useful. It should be deactivated.

latin-panda:
Closes #659

- [x] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [x] Chrome for Android
- [ ] Not applicable

### What's changed
- Adjust the zoom controls' height to allow point placement just below the zoom controls in mobile view.

- Prevent micro-panning from canceling click actions to make point creation easier.
- Display the full-screen close button on the desktop overlay when in full-screen mode.

- Clear setTimeout when stopping current location tracking to prevent race conditions when removing a saved geopoint with `maps`.
- Stop watching the location automatically when exiting full-screen mode.

Test server has been updated with the changes

### DIFF
--- a/packages/web-forms/src/components/common/map/MapBlock.vue
+++ b/packages/web-forms/src/components/common/map/MapBlock.vue
@@ -133,6 +133,15 @@ watch(
 	(newValue) => mapHandler.setupMapInteractions(newValue)
 );
 
+watch(isFullScreen, async (newValue) => {
+	if (!newValue) {
+		await nextTick();
+		isAdvancedPanelOpen.value = false;
+		mapHandler.fitToAllFeatures();
+		mapHandler.stopWatchingCurrentLocation();
+	}
+});
+
 const onFeaturePlacement = () => emitSavedFeature();
 
 watch(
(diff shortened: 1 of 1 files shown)


latin-panda:
@garethbowen, Aly and I reviewed the [user feedback from the forum](https://forum.getodk.org/t/let-us-know-what-you-think-of-maps-in-web-forms/57437/10). We’re going to wait for more input on some of the items, but I’ve addressed the others in this PR.

PR #669, opened 2026-02-24
chore: update vitest

garethbowen:
I'm planning to merge this into the other dependency update PR before merging that to main if/when it passes review.


packages/common/types/vitest-config.ts
```diff
@@ -1,3 +0,0 @@
-import type { UserConfig } from 'vitest/config';
```
latin-panda: 💛

latin-panda (approved):
Excelent!

PR #684, opened 2026-02-25
chore: update matrix and add issue links for incomplete items

lognaturel:
I have a few updates to propose based on things I believe are now complete, new specs that have been added, and organization I find is clearer. Explanations inline.

I added links to issues because I keep finding myself wanting to confirm which items we do and don't have issues for when looking at the readme. If we find it helpful I can make sure we have issues for every item in the list and update all the links.


feature-matrix.json
```diff
@@ -93,9 +93,11 @@
     "range start, end, step": "✅",
     "image max-pixels": "",
     "audio quality": "",
-    "Audit: location-priority, location-min-interval, location-max-age, track-changes, track-changes-reasons, identify-user": ""
+    "Audit: location-priority, location-min-interval, location-max-age, track-changes, track-changes-reasons, identify-user": "",
+    "geotrace/shape incremental=true": "",
```
lognaturel: Two new specs

PR #693, opened 2026-03-01
fix(#692): handle last-saved gracefully

### LINKED ISSUE #692: Gracefully handle last-saved functionality so forms still load
### Description

Until the [last-saved virtual secondary instance](https://github.com/getodk/web-forms/issues/306) is supported, WF should gracefully handle loading forms that use it, and behave as if there is no last saved instance on this device. This is acceptable in this case because last-saved is mostly used as a convenience to reduce data re-entry between form submissions.

Currently WF engine fails with an error.

---

### Checklist
  - [x] No
  - [x] No
  - [x] No
  - [x] No
---

### User Stories

- **As a** data entry user,
  **I want** to be able to load forms which use last-saved,
  **so that** I can use WF.

garethbowen:
Closes #692 

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

### DIFF
--- a/packages/common/src/fixtures/test-javarosa/resources/org/javarosa/xform/parse/last-saved-blank.xml
+++ b/packages/common/src/fixtures/test-javarosa/resources/org/javarosa/xform/parse/last-saved-blank.xml
@@ -8,6 +8,9 @@
             <instance>
                 <data id="data">
                     <item/>
+                    <meta>
+                        <instanceID/>
+                    </meta>
                 </data>
             </instance>
             <instance id="last-saved" src="jr://instance/last-saved"/>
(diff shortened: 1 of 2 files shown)


packages/common/src/fixtures/test-javarosa/resources/org/javarosa/xform/parse/last-saved-blank.xml
```diff
@@ -6,7 +6,7 @@
         <h:title>Form with last-saved instance (blank)</h:title>
         <model>
             <instance>
-                <data>
+                <data id="data">
```
garethbowen: Previously this test was failing on a missing ID which is arguably still incorrect but should be outside the scope of this test...

PR #700, opened 2026-03-02
fix(#691): bind invalid values as empty strings

### LINKED ISSUE #691: Don't throw Errors when loading invalid values
### Expected Behaviour

When editing a form which has invalid values these should be gracefully emptied out rather than throwing uncaught Errors.

### Observed Behaviour

Errors are thrown.

### Reproduction Steps

Form (XML or XLSX file):

```xml
<?xml version="1.0"?>
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:h="http://www.w3.org/1999/xhtml"
        xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:xsd="http://www.w3.org/2001/XMLSchema"
        xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms"
        xmlns:odk="http://www.opendatakit.org/xforms">
  <h:head>
    <h:title>Int</h:title>
    <model odk:xforms-version="1.0.0">
      <instance>
        <data id="int" version="2025020401">
          <val/>
          <meta>
            <instanceID/>
          </meta>
        </data>
      </instance>
      <bind nodeset="/data/val" type="int" />
    </model>
  </h:head>
  <h:body>
    <input ref="/data/val">
      <label>Integer:</label>
    </input>
  </h:body>
</h:html>
```

Steps:
1. Load the form in central
2. Using Enketo submit the form with the value "9999771234"
3. Switch to using Web Forms
4. Edit the submission from 2
5. The form fails to load with the fatal error: "(int) Unable to decode int value from 9999771234: expected value to be between -2147483648 and 2147483647 (inclusive)"

### Screenshots or Videos

Image

garethbowen:
Closes #691

- [ ] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

### DIFF
--- a/.changeset/tender-snakes-rescue.md
+++ b/.changeset/tender-snakes-rescue.md
@@ -1,0 +1,7 @@
+---
+'@getodk/xforms-engine': patch
+'@getodk/web-forms': patch
+'@getodk/scenario': patch
+---
+
+Fixes integer and range codecs to not throw errors when given invalid data


packages/scenario/test/range.test.ts
```diff
@@ -0,0 +1,25 @@
...
+	it('should allow range to be set', () => {
+		scenario.answer('/stats/balance', '-1');
+		expect(scenario.answerOf('/stats/balance').getValue()).toBe('-1.0');
+	});
+
+	it('nulls out value if value is out of range', () => {
+		scenario.answer('/stats/balance', '-5');
+		expect(scenario.answerOf('/stats/balance').getValue()).toBe('');
+	});
+});
```
garethbowen: I couldn't find any existing tests for range which was weird, so wrote this very basic one. It didn't fit in bind-types because range is a style not a type.

PR #709, opened 2026-03-04
fix(#708): add div to sanitizer allowlist to preserve inline styles

### LINKED ISSUE #708: HTML <div> and associated styles are stripped by DOMPurify sanitizer
### Description

The div tag is currently missing from the ALLOWED_TAGS list in the DOM_PURIFY_SETTINGS within the Vue layer.

Because of this, any HTML content containing a div, which is frequently used for alignment and structural styling, is stripped by the sanitizer. This results in the complete loss of the style attributes (such as text-align:center) during the final render, even if the engine has otherwise processed the content correctly.

### Reproduction Steps

Form (XML or XLSX file): [markdown.xml](https://github.com/getodk/web-forms/blob/main/packages/common/src/fixtures/notes/3-notes-with-markdown.xml)

Steps:
1. Load form 
2. Find "centered div" text
3. Observe it's left-aligned. It should be centered.

### Environment

- Browser version: Chrome
- OS: MacOS
- Web Forms version: Observed in v0.20.0 

### Additional Context

- Observed while analyzing the diff of the visual e2e tests around markdown.

latin-panda:
Closes #708 

- [x] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

### What's changed

The div tag is currently missing from the ALLOWED_TAGS, causing the sanitizer to replace it with a span with no styles.

From a security standpoint, adding div is functionally the same as adding `p`, `li` or `table`. If we've already decided it's safe to allow those with a style attribute, div doesn't introduce any unique technical vulnerabilities.

### DIFF
--- a/.changeset/gold-dingos-play.md
+++ b/.changeset/gold-dingos-play.md
@@ -1,0 +1,5 @@
+---
+'@getodk/web-forms': patch
+---
+
+Add div to sanitizer allowlist to preserve inline styles


garethbowen (approved):
Thanks! No idea how I missed this...

PR #725, opened 2026-03-10
fix(#722): markdown supports computed link urls

### LINKED ISSUE #722: Phone number link is broken
In my XLSForm, I have `[${phone}](tel:${phone})` where the `phone` field has a value like +255718774260

The link rendered by Web Forms is

 tel:`--ODK-OUTPUT-STRING-9--`

I don't think the `tel` pattern is very common but I think that using form values in links does happen. I'm not entirely sure how this issue manifests so it feels worth investigating to me.

Example form: https://staging.getodk.cloud/projects/199/forms/hh_listing_baseline/

garethbowen:
Closes #722

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable


packages/web-forms/src/lib/format/markdown.ts
```diff
@@ -37,7 +37,11 @@ export const getStylePropertyMap = (node: ParentMarkdownNode): StyleValue | unde
 
 export const getUrl = (node: ParentMarkdownNode): string | undefined => {
 	if (node.elementName === 'a') {
-		return (node as AnchorMarkdownNode).url;
+		const url = (node as AnchorMarkdownNode).url;
+		if (!DOMPurify.isValidAttribute('a', 'href', url)) {
+			return 'about:blank';
+		}
```
garethbowen: This wasn't identified in the issue but now that this is fixed it was possible for user entered strings to be injected into the url and execute JS.

PR #727, opened 2026-03-10
feat(#332): add translation support

### LINKED ISSUE #332: Add translation support for UI components based on browser locale
### Description
Currently, UI components such as buttons, labels, and other text elements (e.g., the PrimeVue `Calendar` component) do not adapt to the user’s browser language. To improve internationalization (i18n), these components should display text in the language corresponding to the browser’s locale (e.g., `navigator.language`). This will enhance the user experience by providing a localized interface without requiring manual language selection.

Calendar

Geopoint

### User Story
As a user,  
I want all UI elements on the screen (e.g., buttons, labels, calendar labels) to automatically display in my native language based on my browser’s locale,  
So that I can interact with the application in a familiar and comfortable way without needing to manually set my language preference.

### Considerations
- How should UI language interact with form language (see [forum thread](https://forum.getodk.org/t/setting-a-default-language-in-enketo/32873) about Enketo behavior)
- [PrimeVue has a configuration](https://v3.primevue.org/configuration/#locale) to define the language as json. [There is a project already](https://github.com/primefaces/primelocale) providing translations to many languages, but it seems heavy (1MB unpacked) 
  - We can take some of these translations to keep the bundle light. 
  - Ideally only the required language is requested
- Not all components are PrimeVue, and they still need translations (Geopoint). Search by `TODO: translations` to find places where translations are needed. 
- The browser’s locale is determined using `navigator.language` (e.g., "en-US" → "en").
- Only static UI text (e.g., button labels, calendar labels) needs translation at this stage; [dynamic content is translated in the form definition](https://docs.getodk.org/guide-form-language/). 
- A fallback language (English) is acceptable when the user’s locale is unsupported.
- Transifex to match other ODK tools
- Bootstrap with Collect/Central/Enketo strings?
  - [Enketo](https://github.com/enketo/enketo/blob/main/packages/enketo-express/locales/src/en/translation.json)
  - [Collect](https://github.com/getodk/collect/blob/master/strings/src/main/res/values/strings.xml)
  - [Central](https://github.com/getodk/central-frontend/blob/master/transifex/strings_en.json)
- Define strings in a single giant file like Collect/Enketo or have option to define by component like with Central?
- RTL considerations such as https://github.com/getodk/web-forms/pull/273#discussion_r1915420541

### Tasks
- [ ] **Design**: Design a uniform solution for all components (PrimeVue and non-PrimeVue components) 
- [ ] **Define Translations**: Define languages and create a translation object/map covering buttons, labels, etc.
- [ ] **Test**: Create a test plan. Verify the UI updates correctly when switching browser languages (e.g., English → Spanish → French). 
    - [ticket: add e2e tests for translations ](https://github.com/getodk/web-forms/issues/326)
- [ ] **Implementation**: Add logic to detect the browser’s language using `navigator.language` and map it to supported locales. Ensure buttons, labels, and other UI elements use the detected locale’s translations.
- [ ] **Documentation**: Update the project’s README or developer guide with instructions on adding new languages.

### Related
- https://github.com/getodk/web-forms/issues/393
- https://github.com/getodk/web-forms/issues/670

latin-panda:
Closes #332 

TODO
- Add project in https://app.transifex.com/getodk/
- Add more test coverage 
- Test more cases
- Finish translations english, spanish, indonesian
- Add changeset
- Document translation guide and best practices 

- [x] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

<details><summary>Browsers with different default languages</summary>

**See message in error banner translated**

</details>

The objective is to implement a scalable, lightweight internationalization that:
- Allows host applications to fetch (Transifex) and provide translation files.
- Supports complex variables and plurals using standard syntax.
- Keeps the Web Forms bundle size minimal for low-end devices.
- Provides contextual information (developer_comment) to translators. Just as Central and Collect do.

**Boundary definition:**
Web Forms will manage its own internal locale state and browser detection, but it will remain agnostic to how translation files are retrieved (Transifex).
- Web forms component 
  -  Responsible for detecting the user's preferred language via standard browser APIs (`navigator.language`) upon initialization
  -  Owns the active locale state and the language-switcher UI dropdown (there's an open question about language-switcher below).
  - Exposes a required prop/callback (e.g., fetchTranslations(localeCode)) that it triggers when it mounts or when the user switches languages.
- The host app (Central)
  -  The data provider. It implements the `fetchTranslations` function passed to Web Forms, handling the actual network request to fetch the appropriate JSON translation file from the backend or Transifex.

**Translation format:**
Web Forms will utilize a nested JSON structure with strings and developer comments to align with Central and Collect, following ICU Message Format syntax for handling variables and plurals. 

It will include TypeScript interfaces to ensure a valid structure and documentation to guide contributors.   

**Library:**
To parse the ICU strings, Web Forms will use `intl-messageformat` (from FormatJS) rather than heavier frameworks. 
- It's lightweight, 9.6kB gzipped.
- Strictly follows ECMA-402. It offloads the heavy lifting of pluralization rules to the browser's native Intl API.
- Transifex natively understands ICU MessageFormat
- It has a strong community that keeps it updated with the latest ECMA-402.

<details><summary>Other libraries considered</summary>

- vue-i18n: Adds ~15-20kb to the bundle size. It uses <i18n> block that embeds strings into the compiled component (like Central does), making components less readable (I'm not a fan)
- typesafe-i18n: Fast and lightweight (~1.2kb), but translations must be in a custom syntax. It doesn't support Transifex natively; implementing that would require complex, custom intermediate parsers.
- Native Intl.PluralRules Wrapper: Building custom code around browser APIs without dependencies places the burden of maintenance on us. 

</details>

**Implementation details:**

Each component that needs translations will have a sibling file: `<component>.i18n.ts`
This separation maintains component readability and focus on code. It is also extensible; if additional languages need to be added in the future, the main component remains unaffected. And good for tree-shaking support (more on that below).

There's a script (package/web-forms/script/extract-translations.js) that dynamically gets all `<component>.i18n.ts` and compiles them into a single `strings_en.json` file for Transifex upload. Maybe the upload can be automated as part of the release step in Github Actions. 

It uses one Vue composable to initialize the internalization and fetch the translation files from the Host app. Another Vue composable keeps the locale state and executes the `intl-messageformat` parser.

Components will access strings via dot-notation: `<span>{{ t('errors.invalidInputs', { count: 3 }) }}</span>`

**Possible future considerations (not part of this delivery)**

- Have a `destructure-translations.js` script that takes the downloaded translated JSON from Transifex and maps the localized strings back to their respective component i18n, allowing the application to easily serve them on request (tree-shaking for all languages, not just the default language).
- Path to PWA (If): No longer rely only on the host app's translations, but dynamically import the generated `*.i18n.[locale].ts` directly. The current structure ensures zero refactoring in the Vue templates/components. 

### Open Question:

If switching language from the dropdown
Screenshot 2026-03-11 at 2 44 51 AM

How to get the locale code if form definitions don't provide the standard:


latin-panda:
@garethbowen There's still plenty to do, but the PR is at a good point to share for feedback. 
What are your thoughts on this approach? I've included an explanation in the description box.

packages/web-forms/src/components/OdkWebForm.i18n.ts
```diff
@@ -0,0 +1,27 @@
...
+			string: '1 question with error.',
+			developer_comment:
+				'Message shown in the error banner when exactly one question has a validation error.',
+		},
+		validationMultiple: {
+			string: '{count} questions with errors.',
+			developer_comment:
+				'Message shown in the error banner when multiple questions have validation errors. {count} is the number of errors.',
+		},
+	},
```
garethbowen: It'd be better to combine these into one message with logic to handle the plurals.

packages/web-forms/src/lib/i18n/useTranslation.ts
```diff
@@ -0,0 +1,65 @@
...
+				cache.set(cacheKey, new IntlMessageFormat(pattern, locale));
+			}
+			return cache.get(cacheKey)!.format(values) as string;
+		} catch {
+			return pattern;
+		}
+	};
+
+	return { t };
+}
```
garethbowen: I'm nervous about providing custom code and caching, and would lean towards a more complete library such as vue-i18n which does more for us. Plus we have the benefit of standardisation with central and benefiting from things they've learned and tools they've built.

PR #736, opened 2026-03-12
chore: update lint, gh actions, turbo

garethbowen:
My main focuses this round were (a) ticking off more dependabot security warnings, and (b) resolving build warnings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,15 +75,6 @@ jobs:
             packages/web-forms/dist-demo
           key: build-${{ matrix.node-version }}-${{ github.sha }}
 
-      # `@getodk/xpath` tests (currently) expect this time zone
-      # GB: If commenting this out doesn't work, then just inline the plugin with this:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -75,13 +75,6 @@ jobs:
             packages/web-forms/dist-demo
           key: build-${{ matrix.node-version }}-${{ github.sha }}
 
-      # `@getodk/xpath` tests (currently) expect this time zone
-      - uses: szenius/set-timezone@v2.0
-        with:
-          timezoneLinux: ${{ env.TZ }}
-          timezoneMacos: ${{ env.TZ }}
-          timezoneWindows: ${{ env.TZ }}
-
```
garethbowen: This action looks to be in maintenance mode. People have forked it but because the build only runs on linux it's overkill. Turns out we don't need to set the tz anyway, but if tests start breaking we can just set the tz ourselves.

PR #766, opened 2026-04-02
feat(#448): support encrypting submission payload

### LINKED ISSUE #448: Add submission data encryption and display encryption status
### Description
Web Forms currently does not encrypt submission data, though the [base64EncryptedKey is already extracted from the XML](https://github.com/getodk/web-forms/blob/8dc572af77d00ee71d351d6c4d980245181bf0fe/packages/xforms-engine/src/parse/model/SubmissionDefinition.ts#L37) and passed to the host app in the payload. 

Consider implementing the encryption of submission data in Web Forms and adding a UI indicator to inform users if their submissions are encrypted. 

Encryption could be implemented in Web Forms or in the host app (e.g. Central Frontend). However, implementing it in Web Forms would benefit other clients by eliminating the need for them to develop their own encryption solutions.

---

### Checklist
  - [ ] No
  - [x] Yes. Specify the endpoints, methods, or data structures needed.
  - [ ] No
  - [x] No
---

### User Stories

- **As a** program manager
  **I want** to know if the form's data is being encrypted 
  **so that** I can add the encryption key in the form 

---

### Design

- **Figma**: [Insert URL]

---

### Dependencies

-

---

### Additional Notes
- [Enketo has an encryptor](https://github.com/enketo/enketo/blob/ba2bec44acf561f278fd022aa361447674c1d9a8/packages/enketo-express/public/js/src/module/encryptor.js#L99) that we can use as reference. 
- Specs: https://getodk.github.io/xforms-spec/#submission-attributes
- Some notes at https://github.com/getodk/web-forms/issues/318

---

### Related Issues

- Central: [Cleartext submissions can be accepted to an encrypted form](https://github.com/getodk/central/issues/782#top)

garethbowen:
Closes #448 

- [x] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

Minimal testing. I'd really like to add an e2e test one day that actually verifies that the submission can be decrypted, but that requires actually starting up backend. If we end up in the same repo as frontend then there already is a framework to do this so postponing it until we figure out where we're going to go!

No specific form but it only works in central (not demo) and only if you set up encryption as laid out in the test plan.


packages/xforms-engine/src/parse/model/SubmissionManifestDefinition.ts
```diff
@@ -0,0 +1,97 @@
...
+
+		// TODO remove this if it's not implemented
+		// if (this.signature) {
+		// 	const el3 = document.createElementNS(
+		// 		ODK_SUBMISSIONS_NAMESPACE_URI,
+		// 		'base64EncryptedElementSignature'
+		// 	);
+		// 	el3.textContent = this.signature;
+		// 	manifest.appendChild(el3);
+		// }
```
garethbowen: This part is still waiting on a decision. I suspect it's going to get removed completely because it's a) optional, and b) not implemented in central.

garethbowen:
@latin-panda This is ready for review. I'm still waiting on a couple of answers and it'll need a fair amount of QA testing, but as I'm AFK for the next week I thought I'd send it over for review in the meantime. Thanks!

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/README.md
```diff
@@ -0,0 +1,15 @@
...
+> Custom encryption is a bad idea. Do not use this unless absolutely necessary.
+
+> [!CAUTION]
+> Modification of this code requires great care as a bug in the encryption algorithm will make submissions unrecoverable.
+
+This code implements the [ODK Spec](https://getodk.github.io/xforms-spec/encryption) which is very particular about how it's done so as to be compatible with other implementations.
+
+## Implementation
+
+The symmetric encryption parts of the spec are implemnted using CryptoJS because the particularly algorithm required by the spec is not supported by Subtle Crypto, and we use CryptoJS elsewhere.
```
latin-panda:
```suggestion
The symmetric encryption parts of the spec are implemented using CryptoJS because the particular algorithm required by the spec is not supported by Subtle Crypto, and we use CryptoJS elsewhere.
```

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
```diff
@@ -0,0 +1,99 @@
...
+	});
+	return wordArrayToArrayBuffer(encrypted.ciphertext);
+};
+
+const encryptAttachment = async (
+	attachment: File,
+	symmetricKey: Uint8Array<ArrayBuffer>,
+	seed: Seed
+): Promise<File> => {
+	const content = await getBlobData(attachment);
```
latin-panda, line 52: I'm curious about the amount of memory WF is allocating to submissions overall, including attachments and raw data, since it loads the submission into memory again. It'd be good to check how large forms perform on low-end devices, as part of the test plan.

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
```diff
@@ -0,0 +1,99 @@
...
+ */
+import { getBlobData } from '@getodk/common/lib/web-compat/blob.ts';
+import * as CryptoJS from 'crypto-js';
+import { ENCRYPTED_SUBMISSION_ATTACHMENT_NAME, ENCRYPTED_SUFFIX } from './encryption';
+
+const arrayBufferToWordArray = (buffer: Uint8Array<ArrayBuffer>) => {
+	const bytes = [];
+	for (let i = 0; i < buffer.length; i += 4) {
+		bytes.push(
+			(buffer[i]! << 24) | (buffer[i + 1]! << 16) | (buffer[i + 2]! << 8) | buffer[i + 3]!
```
latin-panda: This increases by 4, and it will go out of bounds. However, the bitwise operation treats `undefined` as `0`. It would be better to still check that buffer[i], buffer[i + 1], etc., are defined to make it more robust.

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
```diff
@@ -0,0 +1,99 @@
...
+
+class Seed {
+	readonly ivSeedArray;
+	counter = 0;
+
+	public constructor(
+		readonly instanceId: string,
+		readonly symmetricKey: Uint8Array<ArrayBuffer>
+	) {
+		const key = CryptoJS.lib.WordArray.create(symmetricKey);
```
latin-panda: Does it make sense to use this function here too?
```suggestion
		const key = arrayBufferToWordArray(symmetricKey);
```

packages/xforms-engine/src/parse/model/SubmissionManifestDefinition.ts
```diff
@@ -0,0 +1,97 @@
...
+			manifest.setAttribute('version', this.formVersion);
+		}
+
+		const el = document.createElementNS(ODK_SUBMISSIONS_NAMESPACE_URI, 'base64EncryptedKey');
+		el.textContent = this.base64EncryptedKey;
+		manifest.appendChild(el);
+
+		const el2 = document.createElementNS(ODK_SUBMISSIONS_NAMESPACE_URI, 'encryptedXmlFile');
+		el2.textContent = ENCRYPTED_SUBMISSION_ATTACHMENT_NAME;
+		manifest.appendChild(el2);
```
latin-panda:
```suggestion
		const keyEl = document.createElementNS(ODK_SUBMISSIONS_NAMESPACE_URI, 'base64EncryptedKey');
		el.textContent = this.base64EncryptedKey;
		manifest.appendChild(el);

		const xmlFileEl = document.createElementNS(ODK_SUBMISSIONS_NAMESPACE_URI, 'encryptedXmlFile');
		el2.textContent = ENCRYPTED_SUBMISSION_ATTACHMENT_NAME;
		manifest.appendChild(el2);
```

packages/xforms-engine/src/parse/model/EncryptedSubmissionManifestDefinition.ts
```diff
@@ -0,0 +1,97 @@
+import {
+	ODK_SUBMISSIONS_NAMESPACE_URI,
+	OPENROSA_XFORMS_NAMESPACE_URI,
+} from '@getodk/common/constants/xmlns.ts';
```
latin-panda, line 4: How about moving this to `/instance-state`? It seems like this file is handling a lot of DOM element work, and currently `parse/model` isn't importing from `client-reactivity`. What do you think?

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
```diff
@@ -0,0 +1,99 @@
...
+	const bytes = new Uint8Array(wordArray.sigBytes);
+	for (let j = 0; j < wordArray.sigBytes; j++) {
+		bytes[j] = (wordArray.words[j >>> 2]! >>> (24 - 8 * (j % 4))) & 0xff;
+	}
+	return bytes;
+};
+
+class Seed {
+	readonly ivSeedArray;
+	counter = 0;
```
latin-panda: I think it can be private
```suggestion
	private counter = 0;
```

latin-panda (changes requested):
Really interesting! I left some comments below

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
garethbowen: Done.

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
garethbowen: In this case we don't want to convert the chars to 32-bit because that would corrupt the key, right?

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
garethbowen: I've switched to nullish coalescing.

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
garethbowen: Added to the test plan, thanks!

packages/xforms-engine/src/parse/model/SubmissionManifestDefinition.ts
garethbowen: Thanks, fixed!

packages/xforms-engine/src/parse/model/EncryptedSubmissionManifestDefinition.ts
garethbowen: Yes good thinking. I think I've made an improvement - what do you think?

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
latin-panda: The cryptoJS algorithm requires the key to be formatted as an array of 32-bit words, right?
If we pass the 8-bit array directly, cryptoJS misinterprets the length and compensates by adding zeros. The `arrayBufferToWordArray` handles it by grouping the 32 bytes into 8 blocks of 32-bit words without changing the bits. Were you thinking this might cause a mismatch with Central?

latin-panda: What do you think about adding Alex as a second pair of eyes on these PRs? I think he has evaluated security across the Central layers. Maybe he has some feedback to share 🤔

garethbowen: > Were you thinking this might cause a mismatch with Central?

Yes exactly. We send the symmetric key through to central so it can decrypt the message. I made this change without changing the `asymmetric` code and the decryption no longer worked. It's fiddly because the asymmetric side cannot be supported by crypto-js so it's using native crypto instead which has a completely different API, so it's very fragile. Do you think I should endeavour to update the asymmetric code to match, or leave this as is?

> What do you think about adding Alex as a second pair of eyes on these PRs?

Yes, good idea. I was hoping someone who had implemented encryption in Collect or Enketo would be able to help out but that code hasn't changed for ~7 years and I don't recognise any of the GH usernames. Talking with LN this morning she also recommended Alex so I'll reach out to him.

packages/scenario/test/submission-encryption.test.ts
```diff
@@ -0,0 +1,81 @@
...
+			submissionElements: [t(`submission base64RsaPublicKey="${base64RsaPublicKey}"`)],
+		});
+		const submissionResult = await scenario.prepareWebFormsInstancePayload();
+
+		expect(submissionResult.submissionMeta).toMatchObject({
+			encryptionKey: base64RsaPublicKey,
+		});
+
+		const entries = submissionResult.data[0].entries();
+		expect(submissionResult.data.length).to.equal(1);
```
alxndrsn:
```suggestion
		expect(submissionResult.data.length).to.equal(1);
		const entries = submissionResult.data[0].entries();
```

alxndrsn: Does it make more sense to check the length _before_ dereferencing the first item?

packages/scenario/test/submission-encryption.test.ts
alxndrsn, line 1: There doesn't seem to be any test here that the encrypted content can be _de_-crypted.  Is that expected?

packages/scenario/test/submission.test.ts
```diff
@@ -808,27 +806,6 @@ describe('Form submission', () => {
...
 				}
 			);
-
-			it('includes a form-specified `base64RsaPublicKey` as encryptionKey', async () => {
-				const base64RsaPublicKey = btoa(createUniqueId());
-				const scenario = await buildSubmissionPayloadScenario({
-					submissionElements: [
-						// Note: `t()` fails here, presumably because the ported JavaRosa
-						// `parseAttributes` doesn't expect equals signs as produced in
-						// the trailing base64 value.
```
alxndrsn, line 818: How come `t()` no longer fails in the transplanted+updated test?

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/README.md
```diff
@@ -0,0 +1,15 @@
+# Encryption
+
+> [!CAUTION]
+> Custom encryption is a bad idea. Do not use this unless absolutely necessary.
+
+> [!CAUTION]
+> Modification of this code requires great care as a bug in the encryption algorithm will make submissions unrecoverable.
+
+This code implements the [ODK Spec](https://getodk.github.io/xforms-spec/encryption) which is very particular about how it's done so as to be compatible with other implementations.
```
alxndrsn:
```suggestion
The code in this directory implements the [ODK Spec](https://getodk.github.io/xforms-spec/encryption) which is very particular about how it's done so as to be compatible with other implementations.
```

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/README.md
alxndrsn, line 1: This felt familiar as it's a similar approach to https://github.com/getodk/central-backend/tree/142c1c85501d90d225116862ba79dd914c5fcfc1/lib/util/quarantine

[Added for info; no resolution required.]

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/encryption.ts
```diff
@@ -0,0 +1,51 @@
...
+export const encryptSubmission = async (
+	formId: string,
+	formVersion: string | undefined,
+	instanceId: string,
+	instanceXML: string,
+	attachments: readonly File[],
+	encryptionKey: string
+): Promise<Submission> => {
+	const symmetricKey = generateSymmetricKey();
+	const base64EncryptedSymmetricKey = await getEncryptedSymmetricKey(encryptionKey, symmetricKey);
```
alxndrsn: On first reading, this name implies that the key is "base64-encrypted" 🤔

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
```diff
@@ -0,0 +1,102 @@
...
+import { ENCRYPTED_SUBMISSION_ATTACHMENT_NAME, ENCRYPTED_SUFFIX } from './encryption';
+
+const arrayBufferToWordArray = (buffer: Uint8Array<ArrayBuffer>) => {
+	const bytes = [];
+	for (let i = 0; i < buffer.length; i += 4) {
+		bytes.push(
+			(buffer[i] ?? 0 << 24) |
+				(buffer[i + 1] ?? 0 << 16) |
+				(buffer[i + 2] ?? 0 << 8) |
+				(buffer[i + 3] ?? 0)
```
alxndrsn: 🤔  does this work?

```js
const b = [1, 2, 4, 8];
const currentAlgo     = (b[0] ?? 0 << 24) | (b[1] ?? 0 << 16) | (b[2] ?? 0 << 8) | (b[3] ?? 0);
const withoutCoalesce = (b[0]      << 24) | (b[1]      << 16) | (b[2]      << 8) |  b[3];
console.log({ currentAlgo, withoutCoalesce });
```

prints

```js
{ currentAlgo: 15, withoutCoalesce: 16909320 }
```

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
```diff
@@ -0,0 +1,102 @@
...
+ * WARNING: DO NOT USE
+ *
+ * More info: README.md
+ */
+import { getBlobData } from '@getodk/common/lib/web-compat/blob.ts';
+import * as CryptoJS from 'crypto-js';
+import { ENCRYPTED_SUBMISSION_ATTACHMENT_NAME, ENCRYPTED_SUFFIX } from './encryption';
+
+const arrayBufferToWordArray = (buffer: Uint8Array<ArrayBuffer>) => {
+	const bytes = [];
```
alxndrsn: Are these `bytes` "bytes", or 32-bit words?

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
```diff
@@ -0,0 +1,102 @@
...
+			(buffer[i] ?? 0 << 24) |
+				(buffer[i + 1] ?? 0 << 16) |
+				(buffer[i + 2] ?? 0 << 8) |
+				(buffer[i + 3] ?? 0)
+		);
+	}
+	return CryptoJS.lib.WordArray.create(bytes, buffer.byteLength);
+};
+
+const wordArrayToArrayBuffer = (wordArray: CryptoJS.lib.WordArray): Uint8Array<ArrayBuffer> => {
```
alxndrsn: Are there tests for `arrayBufferToWordArray()` and `wordArrayToArrayBuffer()`?  It would give me a bit more confidence they're working as intended.

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/asymmetric.ts
```diff
@@ -0,0 +1,44 @@
...
+const KEY_FORMAT = 'spki';
+
+// Equivalent to "RSA/NONE/OAEPWithSHA256AndMGF1Padding"
+const rsaEncrypt = async (symmetricKey: Uint8Array<ArrayBuffer>, publicKey: CryptoKey) => {
+	const encrypted = await crypto.subtle.encrypt(
+		{ name: ASYMMETRIC_ALGORITHM },
+		publicKey,
+		symmetricKey
+	);
+	return btoa(String.fromCharCode(...new Uint8Array(encrypted)));
```
alxndrsn, line 17: What browsers is WebForms aiming to support?

It would be nice to be able to use `new Uint8Array(encrypted).toBase64()`.

Currently at 85%: https://caniuse.com/mdn-javascript_builtins_uint8array_tobase64

alxndrsn: (same for `atob()`)

packages/scenario/test/submission-encryption.test.ts
garethbowen: Done.

packages/scenario/test/submission-encryption.test.ts
garethbowen: Yes that is the glaring hole in this PR... The reason is that WF doesn't currently have the testing infrastructure to start a Central instance to test against. I think we should build that out for this and many other tests that would benefit from full e2e testing. However if the code ends up in the central-frontend repo a lot of this work is already done, so I'm delaying it till then. Issue raised to capture this: https://github.com/getodk/web-forms/issues/775

packages/scenario/test/submission.test.ts
garethbowen: It might still fail... The base64 value in the transplanted test happens to not have any trailing `=` sign so maybe I got lucky?

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/README.md
garethbowen: Done.

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/encryption.ts
garethbowen: True! Even as the author I was thrown for a second. Renamed to `encryptedSymmetricKey`

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
garethbowen: Confirmed. The issue is the operator precedence with the bitwise operator running before the nullish coalesce. I'm 99% sure I tested this e2e into central but maybe this one change got missed somehow.

I'll test again and make sure it fails before adding parens to enforce the order.

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
garethbowen: Good point - renamed to words.

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
garethbowen: Unit tests added for these functions specifically as well as the asymmetric unit.

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/asymmetric.ts
garethbowen: I've deferred this because it'd require changing the tsconfig, which we should probably do but I don't want to expand this PR scope...

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
latin-panda: > Do you think I should endeavour to update the asymmetric code to match, or leave this as is?

Ah, I see. Updating the asymmetric code with extreme care feels like the right thing to do, otherwise it's a big piece of code debt that might sit there for another couple of years. But on the other side, we would need to add backward compatibility for Central (versioning the encryption), right? :( 

If you think the risk and time aren't worth it for this PR, maybe we can leave it for later? We should probably add a comment to document it so the context isn't lost

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
alxndrsn: Resolved?

packages/scenario/test/submission-encryption.test.ts
alxndrsn: What about copy/pasting the central code here?  Or cloning the central repo, and running the relevant code directly with node?

packages/scenario/test/submission.test.ts
alxndrsn: No idea.  I was puzzled by the original comment and I thought you might have solved a mystery :shrug:

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/asymmetric.ts
alxndrsn: :+1: worth commenting similar alongside `atob()` and `btoa()` invocations?

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/wordArrayUtils.ts
alxndrsn, line 1: Any chance of tests for this?  Something like:

```js
describe('wordArrayUtils', () => {
  [
    []
    [ 0 ],
    [ 0, 1 ],
    [ 0, 1, 2 ],
    [ 0, 1, 2, 3 ],
    [ 0x80 ],
    [ 0xFF, 0xFF 0xFF, 0xFF ],
    ...any other edge cases
  ].forEach((input, idx) => {
    it(`should not mangle example #${idx}`, () => {
      assert.deepEqual(
        wordArrayToBuffer(arrayBufferToWordArray(new Uint8Array(input).buffer)),
        input,
      );
    });
  });
});
```

I think this would have helped with the operator precedence bug, and catch any future modifications to these functions.

packages/xforms-engine/src/parse/model/EncryptedSubmissionManifestDefinition.ts
alxndrsn, line 1: Is there a "submission manifest" for non-encrypted submissions?

packages/xforms-engine/test/lib/client-reactivity/instance-state/quarantine/asymmetric.test.ts
```diff
@@ -0,0 +1,53 @@
...
+				modulusLength: 2048,
+				publicExponent: new Uint8Array([1, 0, 1]),
+				hash: 'SHA-256',
+			},
+			true,
+			['encrypt', 'decrypt']
+		);
+
+		const exported = await crypto.subtle.exportKey('spki', keyPair.publicKey);
+		publicKeyBase64 = btoa(String.fromCharCode(...new Uint8Array(exported)));
```
alxndrsn, line 23: Can tests avoid `btoa()`?

packages/xforms-engine/test/lib/client-reactivity/instance-state/quarantine/asymmetric.test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+	it('should return a valid base64 encrypted string', async () => {
+		const actual = await getEncryptedSymmetricKey(publicKeyBase64, symmetricKey);
+
+		expect(typeof actual).toBe('string');
+		expect(actual.length).toBeGreaterThan(0);
+		expect(() => atob(actual)).not.toThrow();
+	});
+
+	it('should produce different ciphertexts for the same input', async () => {
```
alxndrsn, line 39: > should produce different ciphertexts for the same input

Should it be obvious _why_?

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/wordArrayUtils.ts
alxndrsn: Tests seen below in `symmetric.test.ts`.

alxndrsn: I'd still be in favour of adding a data-driven test similar to the one above so that adding new cases is trivial if/when any edge cases are found.

packages/xforms-engine/test/lib/client-reactivity/instance-state/quarantine/symmetric.test.ts
```diff
@@ -0,0 +1,81 @@
+import * as CryptoJS from 'crypto-js';
+import { describe, expect, it } from 'vitest';
+import {
+	arrayBufferToWordArray,
+	wordArrayToArrayBuffer,
+} from '../../../../../src/lib/client-reactivity/instance-state/quarantine/wordArrayUtils';
+
+describe('symmetric encryption', () => {
```
alxndrsn, line 8: Is it worth adding something like:

```js
    describe('encryptAttachment()', () => {
        // Currently not tested because X...
    });
```

packages/xforms-engine/src/parse/model/EncryptedSubmissionManifestDefinition.ts
garethbowen: No there's no submission manifest.

Non-encrypted submissions send the instance xml as the payload. Encrypted submissions encrypt the instance xml and attach it as the first attachment, and then use the submission manifest as the payload.

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
garethbowen: Ok, I tried again a bit more extensively, and got it to work against central. Thanks!

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/asymmetric.ts
garethbowen: They're used throughout the code. I've raised an issue to replace everywhere: https://github.com/getodk/web-forms/issues/776

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/wordArrayUtils.ts
garethbowen: Done.

packages/xforms-engine/test/lib/client-reactivity/instance-state/quarantine/asymmetric.test.ts
garethbowen: Same issue as src code. Will be solved with the raised issue.

packages/xforms-engine/test/lib/client-reactivity/instance-state/quarantine/asymmetric.test.ts
garethbowen: Comment added.

packages/xforms-engine/test/lib/client-reactivity/instance-state/quarantine/symmetric.test.ts
garethbowen: Done.

packages/xforms-engine/test/lib/client-reactivity/instance-state/quarantine/asymmetric.test.ts
alxndrsn: #776

packages/xforms-engine/src/parse/model/EncryptedSubmissionManifestDefinition.ts
alxndrsn: Got it, thanks 👍 

I wonder if that'll be obvious to future readers, or if it would be helpful to mark this file more clearly, e.g. rename as `EncryptedSubmissionManifestDefinition`, or put in an encryption-specific subdirectory.

packages/xforms-engine/test/lib/client-reactivity/instance-state/quarantine/asymmetric.test.ts
alxndrsn: 👍  I'll suggest a more specific version

packages/xforms-engine/test/lib/client-reactivity/instance-state/quarantine/asymmetric.test.ts
```diff
@@ -0,0 +1,54 @@
...
+		// ensures attacker cannot figure out what the plaintext is by looking up known encrypted submissions
+		const result1 = await getEncryptedSymmetricKey(publicKeyBase64, symmetricKey);
```
alxndrsn, line 40:
```suggestion

    // See:
    // * https://en.wikipedia.org/wiki/Optimal_asymmetric_encryption_padding
    // * https://en.wikipedia.org/wiki/Probabilistic_encryption
	it('should produce different ciphertexts for the same input', async () => {
		const result1 = await getEncryptedSymmetricKey(publicKeyBase64, symmetricKey);
```

alxndrsn: Not sure if this is over the top - perhaps this is all obvious to anyone who should be touching this code?

packages/xforms-engine/test/lib/client-reactivity/instance-state/quarantine/symmetric.test.ts
```diff
@@ -0,0 +1,96 @@
...
+				const wordArray = CryptoJS.lib.WordArray.create([], 0);
+				const result = wordArrayToArrayBuffer(wordArray);
+
+				expect(result.length).toBe(0);
+				expect(result).toBeInstanceOf(Uint8Array);
+			});
+		});
+
+		describe('functions are symmetrical', () => {
+			[[][0], [0, 1], [0, 1, 2], [0, 1, 2, 3], [0x80], [0xff, 0xff, 0xff, 0xff]].forEach(
```
alxndrsn:
```suggestion
			[[], [0], [0, 1], [0, 1, 2], [0, 1, 2, 3], [0x80], [0xff, 0xff, 0xff, 0xff]].forEach(
```

packages/xforms-engine/src/parse/model/EncryptedSubmissionManifestDefinition.ts
garethbowen: Done.

packages/xforms-engine/test/lib/client-reactivity/instance-state/quarantine/asymmetric.test.ts
garethbowen: Done.

packages/xforms-engine/test/lib/client-reactivity/instance-state/quarantine/symmetric.test.ts
garethbowen: Done.

alxndrsn (changes requested):
As discussed, this code is critical - user data could be lost if there are bugs now or in future.  Any bugfixes or other changes are at high risk of introducing regressions.  Best to wait until tests can be included in this PR.

packages/common/src/constants/xmlns.ts
```diff
@@ -24,6 +24,9 @@ export type JAVAROSA_NAMESPACE_URI = typeof JAVAROSA_NAMESPACE_URI;
 export const ODK_NAMESPACE_URI = 'http://www.opendatakit.org/xforms';
 export type ODK_NAMESPACE_URI = typeof ODK_NAMESPACE_URI;
 
+export const ODK_SUBMISSIONS_NAMESPACE_URI = 'http://opendatakit.org/submissions';
```
lognaturel: Collect and the original spec use `http://www.opendatakit.org/xforms/encrypted` as the namespace and we should match that. I don't know how Enketo ended up using `http://opendatakit.org/submissions` as the namespace but that's also when the ODK XForms spec was backfilled so that what has ended up documented. We should match Collect.

Additional context: https://github.com/getodk/xforms-spec/issues/344

garethbowen: D'oh! Good timing. Fixed now.

garethbowen:
Migrated to here: https://github.com/getodk/central-frontend/pull/1562

PR #779, opened 2026-04-20
feat(#774): lazy load submission attachments

### LINKED ISSUE #774: Lazy load submission attachments
### Description

When editing forms with large submission attachments form rendering is blocked until the attachments are downloaded. It would be better to lazily load the attachments so users can begin filling in the form while the attachments load.

The code has some todos which calls out the need for this addition

https://github.com/getodk/web-forms/blob/e18c44ba60346819a1889d93976ecf241ec00a26/packages/xforms-engine/src/instance/input/InstanceAttachmentMap.ts#L47-L60

Prior art: Form attachments already lazily load and have an animation style that should be reused for submission attachments.

garethbowen:
Closes: #774

- [x] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [x] Chrome for Android
- [ ] Not applicable

Any file upload form, but you'll need to edit, so it needs to be run with central.
To see the loading animation I slowed down the server by 5 seconds by adding a `new Promise(resolve => setTimeout(resolve, 5000)).then...` on this line: https://github.com/getodk/central-frontend/blob/bad2df032ecaa742c7eacf180100ca512cc9273d/src/components/web-form-renderer.vue#L197

### What's changed

1. The loading has been changed so it doesn't wait for the async attachments to complete.
2. The attachment state was expanded to capture the loading and loadingError values so the UI knows what's going on
3. The UI was updated to have a loading animation and error message

Screenshots in [the issue](#774)

### DIFF
--- a/packages/scenario/test/instance-attachments/binary-output.test.ts
+++ b/packages/scenario/test/instance-attachments/binary-output.test.ts
@@ -800,7 +800,9 @@ describe('Instance attachments: binary output', () => {
 									t('file-upload-2'),
 									t('meta',
 										t('instanceID', FAKE_INSTANCE_ID)))),
-							bind('/data/file-upload').type('binary'))),
+							bind('/data/file-upload-1').type('binary'),
+							bind('/data/file-upload-2').type('binary'),
+						)),
 					body(
 						upload('/data/file-upload-1'),
 						upload('/data/file-upload-2')
--- a/packages/scenario/test/instance-attachments/binary-output.test.ts
+++ b/packages/scenario/test/instance-attachments/binary-output.test.ts
@@ -784,4 +784,63 @@ describe('Instance attachments: binary output', () => {
 			expect(attachments.size).toBe(0);
 		});
 	});
+
+	describe('unmodified attachments', () => {
+		it('omits unmodified attachments in edits', async () => {
+			const scenario = await Scenario.init(
+				'Basic upload control',
+				// prettier-ignore
+				html(
+					head(
+						title('Basic upload control'),
+						model(
+							mainInstance(
+								t('data id="basic-upload-control"',
+									t('file-upload-1'),
+									t('file-upload-2'),
+									t('meta',
+										t('instanceID', FAKE_INSTANCE_ID)))),
+							bind('/data/file-upload').type('binary'))),
+					body(
+						upload('/data/file-upload-1'),
+						upload('/data/file-upload-2')
+					))
+			);
+
+			const uploadValue1 = new File(['first'], 'upload-1.txt', {
+				type: 'text/plain',
+			});
+			const uploadValue2 = new File(['second'], 'upload-2.txt', {
+				type: 'text/plain',
+			});
+
+			const uploadName = 'upload-2-edited.txt';
+			const uploadData = 'second edited';
+			const uploadValue2Edited = new File([uploadData], uploadName, {
+				type: 'text/plain',
+			});
+
+			scenario.answer('/data/file-upload-1', uploadValue1);
+			scenario.answer('/data/file-upload-2', uploadValue2);
+
+			const edited = await scenario.editCurrentInstance();
+			edited.answer('/data/file-upload-2', uploadValue2Edited);
+
+			const { data } = await edited.prepareWebFormsInstancePayload();
+			const attachments = new Map(
+				Array.from(data[0]).filter(([key]) => key !== ENGINE_CONSTANTS.INSTANCE_FILE_NAME)
+			);
+
+			expect(attachments.size).toBe(1);
+
+			const attachment = attachments.get(uploadName);
+
+			assert(attachment != null);
+			expect(attachment.name).toBe(uploadName);
+
+			const actualData = await getBlobText(attachment);
+
+			expect(actualData).toBe(uploadData);
+		});
+	});
 });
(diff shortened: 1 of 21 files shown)


packages/web-forms/locales/strings_en.json
```diff
@@ -411,6 +411,10 @@
     "string": "Are you sure you want to delete this file?",
     "developer_comment": "Body text of the confirmation dialog shown before deleting an uploaded file."
   },
+  "upload_control.downloading.error": {
```
latin-panda, line 414: If you haven't already, when merging, please upload this translation file to Transifex and just let me know. I'll take care of translating it into Spanish, Indonesian and French.

packages/scenario/test/instance-attachments/binary-output.test.ts
```diff
@@ -784,4 +784,63 @@ describe('Instance attachments: binary output', () => {
...
+					head(
+						title('Basic upload control'),
+						model(
+							mainInstance(
+								t('data id="basic-upload-control"',
+									t('file-upload-1'),
+									t('file-upload-2'),
+									t('meta',
+										t('instanceID', FAKE_INSTANCE_ID)))),
+							bind('/data/file-upload').type('binary'))),
```
latin-panda: Each upload question below needs the bind, right?
```suggestion
							bind('/data/file-upload-1').type('binary'))),
							bind('/data/file-upload-2').type('binary'))),
```

packages/xforms-engine/src/lib/reactivity/createInstanceAttachment.ts
```diff
@@ -137,15 +151,68 @@ export const createInstanceAttachment = (
 	return context.scope.runTask(() => {
```
latin-panda, line 178: The callback for this runTask is getting quite large now. I see 2 groups of things happening: 1. async loading lifecycle and 2. value/name management. What do you think about extracting one of these groups? Perhaps the loading lifecycle.

packages/xforms-engine/src/lib/reactivity/createInstanceAttachment.ts
```diff
@@ -137,15 +151,68 @@ export const createInstanceAttachment = (
...
+				})
+				.catch((_) => {
+					setState(createErrorState());
+				});
+		};
+
+		const retry = () => {
+			setState(createInitialState());
+			attachments.retryFileValue(context.instanceNode);
+			const filePromise = attachments.getInitialFileValue(context.instanceNode);
```
latin-panda: If the client calls retry multiple times while the promise is still pending, it shouldn't create a new one, correct? And avoids any possible race condition. That's likely a job for retryFileValue

packages/xforms-engine/src/instance/UploadControl.ts
```diff
@@ -198,4 +204,8 @@ export class UploadControl
 
 		return this.root;
 	}
+
+	retryDownload() {
```
latin-panda: The UI may include a media download button in the future (I remember seeing that request on forum discussions/tickets). I suggest reserving the term 'download' specifically for actions that save files directly to the user's local file system

```suggestion
	retryFetch() {
```

latin-panda (changes requested):
Nice job! I left some suggestions below

packages/web-forms/locales/strings_en.json
garethbowen: Done.

The idea is for that to be automatic eventually, right?

packages/scenario/test/instance-attachments/binary-output.test.ts
garethbowen: Good catch. Fixed and test still passes.

packages/xforms-engine/src/instance/UploadControl.ts
garethbowen: Done.

packages/xforms-engine/src/lib/reactivity/createInstanceAttachment.ts
garethbowen: Just checking if you mean from a readability perspective or a performance one?

Assuming you meant readability, I've refactored it so more functions are defined outside of the runTask function.

packages/xforms-engine/src/lib/reactivity/createInstanceAttachment.ts
garethbowen: I've added a debounce. I put it in the `createInstanceAttachment` instead of `retryFileValue` because the former has the state information to tell if it's current loading or not.

packages/web-forms/locales/strings_en.json
latin-panda: yes, we need a CI step for that. Better after merging repos with Central

packages/xforms-engine/src/lib/reactivity/createInstanceAttachment.ts
latin-panda: Yes! It's from the readability perspective, thanks!

latin-panda (approved):
I did a smoke test, and the PR looks good to go!

packages/web-forms/locales/strings_en.json
latin-panda: I've provided the translations in Transifex for the 3 languages.

PR #783, opened 2026-04-22
feat(#697): adds support to datetime question type

### LINKED ISSUE #697: Datetime question type
### Description

Implement the `datetime` question type ([specs](https://docs.getodk.org/form-question-types/#datetime-widget)) to allow for the simultaneous capture of both date and time within a single widget. 

---

### User Stories

* As a data collector, I want to record a specific date and time in one screen so that I can capture event timestamps more efficiently.
* As a researcher, I want date and time data to be stored in a standardized format to ensure temporal accuracy across different time zones.
* As a user, I want a clear interface that shows both the selected date and time to confirm the accuracy of the data before moving to the next question.

---

### Additional Notes

- The "no-calendar" support will be covered here: https://github.com/getodk/web-forms/issues/781

---

### Related Issues

- https://github.com/getodk/web-forms/issues/304

latin-panda:
Closes #697

- [x] Chrome
- [x] Firefox
- [x] Safari (macOS)
- [ ] Safari (iOS)
- [x] Chrome for Android
- [ ] Not applicable

### What's changed
- Adds codec, client component and scenario tests
- Extracts common code for reusability (CSS and composable)
- Remove unused code in XPath and clean up datetime constants in the common package.

### DIFF
--- a/.changeset/nice-spiders-write.md
+++ b/.changeset/nice-spiders-write.md
@@ -3,6 +3,7 @@
 '@getodk/web-forms': minor
 '@getodk/scenario': patch
 '@getodk/common': patch
+'@getodk/xpath': patch
 ---
 
 Add support to datetime question type
--- a/.changeset/nice-spiders-write.md
+++ b/.changeset/nice-spiders-write.md
@@ -1,0 +1,8 @@
+---
+'@getodk/xforms-engine': minor
+'@getodk/web-forms': minor
+'@getodk/scenario': patch
+'@getodk/common': patch
+---
+
+Add support to datetime question type
--- a/.changeset/nice-spiders-write.md
+++ b/.changeset/nice-spiders-write.md
@@ -6,4 +6,4 @@
 '@getodk/xpath': patch
 ---
 
-Add support to datetime question type
+Add support to datetime question type.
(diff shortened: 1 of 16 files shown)


latin-panda:
@garethbowen Could you please take a look when you have some time? :)

garethbowen (approved):
Great work. Tested briefly in Firefox and Chrome and looks great.

I hate to leave no comments so I spent a few minutes coming up with something to say... The only thing I can think of is there's quite a lot of duplication between InputTime, InputDate, and InputDateTime. If there was a bug/improvement made in one it's quite likely I would forget to update the other two to match so they may drift. What do you think about merging them into one file with conditions where needed?

latin-panda:
> What do you think about merging them into one file with conditions where needed?

I considered merging the components, but decided to extract the core logic into `useDateTimeInput.ts` instead, we could probably extract a bit more, like that null check. The three components are actually small (< 50 lines each), and merging them now would likely lead to extra union types and extra conditionals throughout the scripts and templates. To me it feels cleaner to keep three focused files rather than one complex one. We can definitely look into moving `isDisabled` from all question types into a common composable. 

I still have two more features to build, so I’ll keep an eye on things. If the complexity starts to climb, we can look at extracting more shared logic from these inputs then.

latin-panda:
The GitHub Actions seemed to be out of service or extremely slow to start (+15min). The second commit is just testing whether the CI is working.

PR #784, opened 2026-04-23
chore(#780): move tree-sitter-xpath into separate repo

### LINKED ISSUE #780: Move tree-sitter-xpath package into its own repo
### Description

Create a new repo and extract the tree-sitter-xpath package into it. Clean up web-forms to remove unnecessary dev dependencies.

garethbowen:
Closes #780 

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable


.github/workflows/ci.yml
latin-panda: Is this comment still needed?

packages/xpath/package.json
```diff
@@ -58,7 +58,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.28.5",
-    "@getodk/tree-sitter-xpath": "0.2.2",
+    "@getodk/tree-sitter-xpath": "^0.2.2",
```
latin-panda, line 61: We need the same in [xforms-engine](https://github.com/getodk/web-forms/blob/86a54ab2dbe28b3cf546a245724b36fb6dc9e696/packages/xforms-engine/package.json#L65), right?

package.json
latin-panda: Is there any chance these 2 can be removed or moved somehow to the new tree-sitter repo? That way, we only include ODK's tree-sitter.

latin-panda (approved):
So happy to see this extracted from the repo ✨

Can this be removed too: [tsconfig.tools.json](https://github.com/getodk/web-forms/blob/86a54ab2dbe28b3cf546a245724b36fb6dc9e696/tsconfig.tools.json#L12)?

Approving now to unblock, since the suggestions below are minor.

.github/workflows/ci.yml
garethbowen: I think it's not needed as it's caching the build output which is now just part of node_modules. Removed.

packages/xpath/package.json
garethbowen: Good call. Done.

package.json
garethbowen: Yes I think so. I'm cautious because of the web-tree-sitter peerdependency. Because WF has to list it as a dependency I don't think the resolutions from tree-sitter-xpath will be respected. I've removed them and nothing has changed in the lock file so I think we're good to go for now at least - if it breaks in the future we can add them back.

PR #789, opened 2026-04-24
fix(#788): number() on the dateTime string to return day count.

### LINKED ISSUE #788: The number() on dateTime string returns datetime string instead of day count
### Expected Behavior

When a `dateTime` field (e.g. `start_date_time = "2026-08-27T08:27:00.000+07:00"`) is used in a `calculate` expression with `number()`, the stored result should be the integer day count since epoch (e.g. `20692`), matching ODK Collect / JavaRosa behavior.

### Reproduction Steps

Form (XML or XLSX file): 

[date_time_comparison 02.xml](https://github.com/user-attachments/files/27059729/date_time_comparison.02.xml)

Steps:
1. Open the form in web-forms
2. Enter `start_date_time = 2026/08/27 08:27:00` (UTC+7)
3. Enter `end_date_time = 2026/10/28 19:23:00` (UTC+7)
4. Observe `Calc03` and `Calc04` (both use `number()`) displayed the datetime string

### Environment

- Browser version: Chrome
- OS: MacOS
- Web Forms version: latest

latin-panda:
Closes #788

- [x] Chrome
- [ ] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

[date_time_comparison 02.xml](https://github.com/user-attachments/files/27059879/date_time_comparison.02.xml)

### DIFF
--- a/.changeset/modern-adults-read.md
+++ b/.changeset/modern-adults-read.md
@@ -1,0 +1,5 @@
+---
+'@getodk/xpath': patch
+---
+
+Fix the number() on the dateTime string to return day count.
--- a/packages/xpath/src/functions/xforms/number.ts
+++ b/packages/xpath/src/functions/xforms/number.ts
@@ -60,13 +60,9 @@ export const number = new FunctionImplementation(
 			const dateTime = dateTimeFromString(context.timeZone, stringValue);
 
 			if (dateTime != null) {
-				const dayCount = stringValue.includes('T')
-					? String(Math.floor(dateTime.epochMilliseconds / DAY_MILLISECONDS))
-					: stringValue;
-
 				return new DateTimeLikeEvaluation(context, dateTime, {
 					booleanValue: true,
-					stringValue: dayCount,
+					stringValue: String(Math.floor(dateTime.epochMilliseconds / DAY_MILLISECONDS)),
 				});
 			}
 		}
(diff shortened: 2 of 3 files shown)


latin-panda:
@garethbowen this is ready for review :)

packages/xpath/src/functions/xforms/number.ts
```diff
@@ -59,9 +60,13 @@ export const number = new FunctionImplementation(
 			const dateTime = dateTimeFromString(context.timeZone, stringValue);
 
 			if (dateTime != null) {
+				const dayCount = stringValue.includes('T')
+					? String(Math.floor(dateTime.epochMilliseconds / DAY_MILLISECONDS))
+					: stringValue;
```
garethbowen: From [the spec](https://getodk.github.io/xforms-spec/#fn:number) it also suggests that `date` (no time) should convert to number of days since epoch. This fix only works on dateTime - when I test locally I get: `number(${start_date_time}) = 2026-04-07`.

I think it would make sense to also fix `date` while you're here because it probably simplifies the implementation anyway as there would be no need to check for "T".

Either that, or raise an issue for the equivalent `date` bug.

garethbowen (changes requested):
Nice find!

Let's expand the issue, PR, testing, etc to also fix `date` (comment inline).

.changeset/modern-adults-read.md
latin-panda, line 1: 🤣 I just noticed this filename (autogenerated)

packages/xpath/src/functions/xforms/number.ts
latin-panda: I don't know how I missed it. Fixed and added unit tests.

latin-panda:
@garethbowen this is ready for review

garethbowen (approved):
Nice!

PR #792, opened 2026-04-28
fix(#791): execute setvalue when first relevant

### LINKED ISSUE #791: Value changed events don't fire when made relevant for the first time
Reported at: https://forum.getodk.org/t/webforms-not-loading-data-from-entities/58012/3

### Expected Behaviour

A field with a `xforms-value-changed` event which is not relevant on load should be updated when made relevant.

### Observed Behaviour

The value is not set when the field is made relevant, but if the source is changed again then the value is set as expected.

### Reproduction Steps

Form (XML or XLSX file):  [default_relevance.xml](https://github.com/user-attachments/files/27146817/default_relevance.xml)

Steps:
1. Load form
2. Select option A
3. The Count field is displayed with no value set
4. Select option B
5. Now the value is set as expected

garethbowen:
Closes #791 

- [x] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

### DIFF
--- a/packages/scenario/test/actions-events/set-value-action.test.ts
+++ b/packages/scenario/test/actions-events/set-value-action.test.ts
@@ -762,6 +762,80 @@ describe('setvalue action', () => {
 
 			expect(scenario.answerOf('/data/destination')).toEqualAnswer(stringAnswer('myvalue'));
 		});
+
+		it('is triggered only once target is relevant', async () => {
+			const scenario = await Scenario.init(
+				'xforms-value-changed-event',
+				html(
+					head(
+						title('Value changed event'),
+						model(
+							mainInstance(
+								t(
+									'data id="xforms-value-changed-event"',
+									t('source', 'foo'),
+									t('trigger'),
+									t('grp', t('destination'))
+								)
+							),
+							bind('/data/source').type('string'),
+							bind('/data/trigger').type('string'),
+							bind('/data/grp/destination').type('string').relevant("/data/trigger = 'yes'")
+						)
+					),
+					body(
+						input(
+							'/data/source',
+							setvalue('xforms-value-changed', '/data/grp/destination', '/data/source')
+						),
+						input('/data/trigger'),
+						group('/data/grp', input('/data/grp/destination'))
+					)
+				)
+			);
+
+			expect(scenario.answerOf('/data/grp/destination')).toEqualAnswer(stringAnswer(''));
+			scenario.answer('/data/trigger', 'yes'); // make destination relevant
+			expect(scenario.answerOf('/data/grp/destination')).toEqualAnswer(stringAnswer(''));
+		});
+
+		it('sets value even if not relevant (yet)', async () => {
+			const scenario = await Scenario.init(
+				'xforms-value-changed-event',
+				html(
+					head(
+						title('Value changed event'),
+						model(
+							mainInstance(
+								t(
+									'data id="xforms-value-changed-event"',
+									t('source', 'foo'),
+									t('trigger'),
+									t('grp', t('destination'))
+								)
+							),
+							bind('/data/source').type('string'),
+							bind('/data/trigger').type('string'),
+							bind('/data/grp/destination').type('string').relevant("/data/trigger = 'yes'")
+						)
+					),
+					body(
+						input(
+							'/data/source',
+							setvalue('xforms-value-changed', '/data/grp/destination', '/data/source')
+						),
+						input('/data/trigger'),
+						group('/data/grp', input('/data/grp/destination'))
+					)
+				)
+			);
+
+			expect(scenario.answerOf('/data/grp/destination')).toEqualAnswer(stringAnswer(''));
+			scenario.answer('/data/trigger', 'no');
+			scenario.answer('/data/source', 'bar');
+			scenario.answer('/data/trigger', 'yes');
+			expect(scenario.answerOf('/data/grp/destination')).toEqualAnswer(stringAnswer('bar'));
+		});
 	});
 
 	describe('`setvalue`', () => {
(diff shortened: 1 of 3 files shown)


packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
```diff
@@ -240,17 +240,17 @@ const createValueChangedCalculation = (
 		// No element to listen to
 		return;
 	}
-	let previous: string;
 	const sourceElementExpression = new ActionComputationExpression('string', source);
 	const calculateValueSource = createComputedExpression(context, sourceElementExpression); // Registers listener
+	let previous = calculateValueSource();
 	createComputed(() => {
+		if (isEditInitialLoad(context)) {
```
latin-panda: In edit mode, does `calculateValueSource()` always return the actual instance value before `isAttached()` becomes true, or can both happen in the same reactive batch? 🤔

If they occur in the same batch, the `isEditInitialLoad` reset in this line is skipped, and the setvalue may fire incorrectly when loading an edit instance (L255-L261), right?

Is it possible to add a test that loads an edit instance where the question already has a value, and assert that the value is unchanged after initialization?

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
```diff
@@ -240,17 +240,17 @@ const createValueChangedCalculation = (
...
+			// reset the initial value now that the edit instance is bound
+			previous = calculateValueSource();
```
latin-panda: Also, `isEditInitialLoad` is only true when `!isAttached()`, so this `calculateValueSource()` call takes the try/catch path in `createComputedExpression` rather than the [if (context.isAttached())](https://github.com/getodk/web-forms/blob/a64d092caa3fe56aa810781dd3a22ce4ea3eccf3/packages/xforms-engine/src/lib/reactivity/createComputedExpression.ts#L128) branch, which could be a valid path eventually 🤔
Does it run for edit subscriptions, initial load, but attached?

latin-panda (commented):
I have a couple of questions below:

lognaturel:
@latin-panda brings up really good questions. Any chance an approach where setvalue affects non-relevant nodes would be feasible?

Running actions after relevance has changed feels possible but tricky to get right because there could be multiple levels of relevance and events that may or may not have fired before a relevance change. If this seems like the way to go then another test I’d like to see is:
- fieldA outside groupA, initial value of « foo »
- fieldB outside groupA that when set to yes makes the contents of groupA relevant
- fieldC inside groupA with a setvalue to the value of fieldA when fieldA changes

If fieldB is set to yes without changing fieldA, fieldC should be blank. 

If fieldB is then set to not-yes, fieldA is changed to « bar » and then fieldB is set to yes, fieldC should be « bar ».

Does setgeopoint share some of this infrastructure or would it need to be verified for potential relevance issues separately?

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
garethbowen: I believe that is what [this test](https://github.com/getodk/web-forms/blob/aed13d1b914358114af5ac6d580662c79f3fdb0a/packages/scenario/test/actions-events/set-value-action.test.ts#L709-L710) is doing...

1. It loads an instance for editing with the value set to a date and a setvalue that updates the date on `xforms-value-changed`
2. It verifies the date is unchanged
3. Then it triggers the change and verifies the date has been updated

Is that what you were thinking?

garethbowen:
> Running actions after relevance has changed feels possible but tricky to get right because there could be multiple levels of relevance and events that may or may not have fired before a relevance change

I think the idea was that reactivity would take care of all the different ways relevance can be triggered. The only benefit I can think of would be performance where there's no point running calculations for non-relevant fields. But having said that, the case(s) you specified weren't behaving as expected. I've added both of those as scenario tests and got them passing.

> Does setgeopoint share some of this infrastructure or would it need to be verified for potential relevance issues separately?

setgeopoint is almost identical. The only difference is because "calculation" of acquiring the location is async so there's a branch to handle that differently so I think the relevance will have the same behaviour as setvalue.

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
garethbowen: Based on @lognaturel comment below I removed the `isRelevant` check which made this branch no longer required and all tests still pass.

latin-panda (approved):
Beautiful test coverage! 

Are we planning a point release with this fix?
Or perhaps a minor release with everything we have done so far?

A point release seems nice for those wanting to use entities in WF, but I am open to a minor release as well.

garethbowen:
> Are we planning a point release with this fix?

Discussion raised in Slack.

packages/scenario/test/actions-events/set-value-action.test.ts
```diff
@@ -762,6 +762,80 @@ describe('setvalue action', () => {
 
 			expect(scenario.answerOf('/data/destination')).toEqualAnswer(stringAnswer('myvalue'));
 		});
+
+		it('is triggered only once target is relevant', async () => {
```
lognaturel, line 766: Is this the right name for this test? I'm having trouble connecting it to what it checks. It looks to me like it ensures making `destination` relevant doesn't change its value if `source` hasn't changed?

garethbowen: True - I think it was a remanent from a previous iteration. Fixed...

PR #793, opened 2026-04-28
fix(#790): input numeric should update reactively when value changes

### LINKED ISSUE #790: Reactively updating numeric fields does not update the UI
### Expected Behaviour

If a user-visible field has a setvalue targeting it, the value is populated and visible when the corresponding event is fired.

### Observed Behaviour

The UI is not updated, however if you submit the form the correct value is submitted.

### Reproduction Steps

Steps:
1. Use [defaults_relevance.xlsx](https://github.com/user-attachments/files/27139069/defaults_relevance.xlsx)
2. Notice that selecting a value in the first question does not set a default for the subsequent questions

### Environment

- Browser version: Firefox 149.0.2
- OS: macOS
- Web Forms version: v0.23.1

garethbowen:
Closes #790 

- [x] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

### DIFF
--- a/packages/web-forms/src/components/form-elements/input/InputNumeric.vue
+++ b/packages/web-forms/src/components/form-elements/input/InputNumeric.vue
@@ -84,7 +84,6 @@ const modelValue = computed<number | null>({
 				renderKey.value++;
 			}
 		} catch {
-			props.setNumericValue(currentValue);
 			// Re-render to restore previous value if something fails
 			renderKey.value++;
 		}
--- a/packages/web-forms/src/components/form-elements/input/InputNumeric.vue
+++ b/packages/web-forms/src/components/form-elements/input/InputNumeric.vue
@@ -1,8 +1,7 @@
 <script setup lang="ts">
 import type { InputNumberInputEvent } from 'primevue/inputnumber';
 import InputNumber from 'primevue/inputnumber';
-import { type ComponentPublicInstance, nextTick, watch } from 'vue';
-import { customRef, ref } from 'vue';
+import { type ComponentPublicInstance, computed, nextTick, ref, watch } from 'vue';
 
 interface NumericNodeState {
 	get required(): boolean;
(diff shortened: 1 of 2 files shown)


packages/web-forms/src/components/form-elements/input/InputNumeric.vue
```diff
@@ -63,37 +62,33 @@ if (props.isDecimal) {
...
-			const stringValue = assignedValue != null ? assignedValue.toString() : '';
-			let newValue = stringValue.length <= props.maxCharacters ? assignedValue : currentValue;
-
-			if (newValue != null) {
-				const { min = newValue, max = newValue } = props;
-
-				if (min !== newValue || max !== newValue) {
-					newValue = Math.max(min, Math.min(newValue, max));
-				}
+const modelValue = computed<number | null>({
```
garethbowen, line 65: The actual change was switching from using customRef to computed and not storing the internal value but referring back to the props.

lognaturel: Oh! Good one. Does this also address the form from the forum? Those weren’t all numeric values.

garethbowen: @lognaturel The forum form needs #791 which is a completely unrelated issue!

packages/web-forms/src/components/form-elements/input/InputNumeric.vue
```diff
@@ -63,37 +62,33 @@ if (props.isDecimal) {
...
+			if (newValue !== assignedValue) {
+				// Re-render if value is clamped
 				renderKey.value++;
 			}
-		},
-	};
+		} catch {
+			props.setNumericValue(currentValue);
+			// Re-render to restore previous value if something fails
+			renderKey.value++;
```
latin-panda: In the rare event that `props.setNumericValue` throws an error, calling it again within the catch block can trigger a secondary uncaught exception. This leaves the UI in an inconsistent state: the input displays the unsaved value (e.g., 3), while the form has the previously saved value (e.g., 2), and like that the form can be submitted. 

I suggest removing the call in the catch block. The `renderKey` remounts the component, and the `props.numericValue` getter runs to sync the input with the saved value, so the UI matches the form's values and no errors appear in the browser's console. 

```suggestion
			// Re-render to restore previous value if something fails
			renderKey.value++;
```

latin-panda (approved):
Love this improvement! Thanks. One minor comment below, approving now to unblock.

PR #796, opened 2026-04-29
chore(#787): migrate from yarn to npm

### LINKED ISSUE #787: Convert to using npm instead of yarn
To facilitate merging web-forms into central-frontend we need to standardise on one package manager.

garethbowen:
Closes #787

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

This is quite risky because unfortunately there was no way to convert back to npm while preserving the lock file so it required a clean `npm install`. I manually reverted some known libraries that a pegged to a version but I may have missed some. So we should regression test this fully before release, BUT moving WF to CF will also be risky and require regression testing so I suggest we defer it until the next minor Central release.


README.md
```diff
@@ -422,32 +422,32 @@ For developers, see [TRANSLATIONS.md](./packages/web-forms/TRANSLATIONS.md) for
...
 ### Running locally
 
 To run ODK Web Forms from this repository:
 
 ```sh
-yarn
-yarn build
-yarn workspace @getodk/web-forms dev
+npm ci
+npm build
```
latin-panda:
```suggestion
npm run build
```

scripts/feature-matrix/render.js
```diff
@@ -100,4 +100,4 @@ await fs.writeFile(new URL('./README.md', rootUrl), updatedReadme, { encoding: '
 
 const rootDir = fileURLToPath(rootUrl);
 
-spawnSync('yarn', ['format:readme-only'], { cwd: rootDir });
+spawnSync('npm', ['run format:readme-only'], { cwd: rootDir });
```
latin-panda: The formatter isn't running properly (silently)

```suggestion
spawnSync('npm', ['run', 'format:readme-only'], { cwd: rootDir });
```

The right side shows the readme with wrong format.

Image

latin-panda (approved):
It looks good! I haven't done a deep test yet, but I did leave a couple of minor fixes below.

garethbowen:
@latin-panda I'd like to get this merged soon because if anything changes in the yarn.lock it's hard to port it to the package-lock.json. However it's fairly risky because of unknown changes in dependencies. However because we've just released now is the perfect time to merge and find any issues. So I'm going to go for it. Sorry in advance if I've missed anything!

Issue #777, opened 2026-04-17
Increase the hit target for jumping to a point on a range


lognaturel:
### Description
Currently, jumping to a specific value on a range requires tapping/clicking precisely on the track. The track is not very tall (for horizontal, or wide for vertical) so this is hard to do precisely. It would be helpful to increase the touch target size.

This is low priority!

---

### Checklist
- **Does it need UI/UX design?**
  - [x] Yes, just the number of screen pixels tolerance to add
  - [ ] No
- **Does it need API design?**
  - [ ] Yes. Specify the endpoints, methods, or data structures needed.
  - [x] No
- **Does it need design around state flow?**
  - [ ] Yes. Outline the state flow requirements.
  - [x] No
- **Does it need a test plan?**
  - [x] Yes. Add the test plan as a comment in this task or attach the document.
  - [ ] No
---

### User Stories

- **As a** [user type],
  **I want** [functionality],
  **so that** [benefit].

---

### Design

- **Figma**: [Insert URL]
- **Design Docs**: [Insert URL]

---

### Dependencies

-

---

### Additional Notes

-

---

### Related Issues

-

garethbowen:
This was a quick fix in the end, and it was annoying me greatly, so I fixed it.

The behaviour still isn't perfect - I think the primevue widget has a bug in the JS implementation of the click handler where it selects the tick above where you clicked (ie: ceil) but I think you want the tick closest to where you clicked. This makes it particularly hard to click on the smallest number - you actually have to hit off the bottom/left of the slider.

But that is a separate bug with a separate fix. This is a step in the right direction!

lognaturel:
> it was annoying me greatly

Glad to know I was not the only one :D 

> it selects the tick above where you clicked (ie: ceil) but I think you want the tick closest to where you clicked

Yes, that's not great either. I've added a comment to a somewhat-related thread for PrimeVue: https://github.com/primefaces/primevue/issues/7609#issuecomment-4374838034

PR #800, opened 2026-05-05
feat(#714): show selected value on range questions

### LINKED ISSUE #714: Show the selected value for range question
### Description
The selected value should be shown above/next to the slider track.
---

### Checklist
  - [ ] No
  - [ ] No
  - [ ] No
  - [ ] No
---

### User Stories

- **As a** [user type],
  **I want** [functionality],
  **so that** [benefit].

---

### Design

- **Figma**: [Insert URL]
- **Design Docs**: [Insert URL]

---

### Dependencies

-

---

### Additional Notes

-

---

### Related Issues

-

garethbowen:
Closes #714 

- [x] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

### DIFF
--- a/packages/web-forms/src/components/form-elements/range/RangeControl.vue
+++ b/packages/web-forms/src/components/form-elements/range/RangeControl.vue
@@ -56,11 +56,11 @@ const orientation = props.node.appearances.vertical ? 'vertical' : 'horizontal';
 
 	<template v-else>
 		<div :class="['range-control-container', orientation]">
+			<div class="range-value">
+				<span>{{ numberValue }}</span>
+			</div>
 			<div class="range-bound range-min">
 				{{ start }}
-			</div>
-			<div class="range-value">
-				{{ numberValue }}
 			</div>
 			<RangeSlider
 				:id="node.nodeId"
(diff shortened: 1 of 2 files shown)


garethbowen:
@latin-panda This is ready for review. Screenshots and some explanation on the issue. Thanks!

packages/web-forms/src/components/form-elements/range/RangeControl.vue
```diff
@@ -118,41 +117,63 @@ const orientation = props.node.appearances.vertical ? 'vertical' : 'horizontal';
 		line-height: 1;
 	}
 
+	:deep(.p-slider-handle) {
```
latin-panda: Can we make the selected value 'pop' more? When it's at the very start or end of the slider, it’s harder to distinguish the selection from the labels at first glance. We can center the selected value, and shrinking the min/max numbers, that'd match Collect, for example:

packages/web-forms/src/components/form-elements/range/RangeControl.vue
```diff
@@ -59,6 +59,9 @@ const orientation = props.node.appearances.vertical ? 'vertical' : 'horizontal';
 			<div class="range-bound range-min">
 				{{ start }}
 			</div>
+			<div class="range-value">
+				{{ numberValue }}
```
latin-panda: Now that the selected value is showing, I've noticed a bit of a glitch:
Sometimes I need to tap the circle several times to select it. This happens in both Chrome and Firefox. 

https://github.com/user-attachments/assets/714abbc1-ef48-4317-9fd8-4afef6cb40d8

Is it a quick win we can include in this PR?

latin-panda (commented):
This is a much-needed improvement! 🤩

packages/web-forms/src/components/form-elements/range/RangeControl.vue
lognaturel: I also find the stationary selected value in the center of the range a little easier to use than when it's following the thumb.

packages/web-forms/src/components/form-elements/range/RangeControl.vue
lognaturel: I bet you're experiencing the issue @gareth pointed out at https://github.com/getodk/web-forms/issues/777#issuecomment-4374870790 Seeing this video makes me think it's more serious than I originally thought.

garethbowen: I can't think of a way to fix that without changing primevue. The logic to determine the new value is clearly broken which makes it very difficult to select the lowest value. The workaround is to drag the slider which works reliably for me.

I've raised a new issue to track this separately as it's not a quick fix as far as I can see: #801

packages/web-forms/src/components/form-elements/range/RangeControl.vue
garethbowen: I'm not a fan of the stationary selected value because it's not clear to me that it's the selected value and not just the value at the centre of the line. Once you start dragging it's obvious, of course, but if a default value is set then it's important that it's obvious what the value is set to.

> than when it's following the thumb

@lognaturel Do you mean because the thumb obscures the value so it's difficult to set accurately on mobile?

lognaturel: Interesting point! Nicole might have some knowledge of best practices to reference if you want to pull her in. I'm also ok with saying we follow e.g. Material 3 which does pin the value to the thumb: https://m3.material.io/components/sliders/overview

> Do you mean because the thumb obscures the value so it's difficult to set accurately on mobile?

To be explicit, I meant the range thumb as in the circle that you move around to select a value, not a human thumb! I don't think that thumb can ever cover the value. For me it's more about having a single place where I always see the current value if I'm looking for a numeric representation rather than looking at where the thumb is. But then again, maybe the more important thing for range questions is the relative position.

I agree with @latin-panda's general observation that styling the value a bit differently from the tick labels would make it clearer that they represent something different. Maybe making it slightly bigger would be enough. I see Material 3 styles it like a callout.

garethbowen: Awkward timing but I've just changed the slider to centre the value to make it consistent with Collect. It's easy enough to revert the last commit, so have a look below and between the three of us we can just go with the majority?

image

lognaturel: You're lightning fast! I'm very happy with that.

latin-panda (approved):
Looks good! Thanks!

PR #802, opened 2026-05-06
fix(#749): handle negative step values

### LINKED ISSUE #749: Use absolute value of range step attribute value
As discussed at https://github.com/XLSForm/pyxform/issues/817 to match Collect

When a negative step is passed in, the PrimeVue component currently doesn't work well.

This is low-priority but also likely very small!

garethbowen:
Closes #749

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

Test written

This would've been much easier to do in the vue component but because it's a spec feature I think it's better to have it in the engine for all clients to benefit from.

```xml
<h:html xmlns="http://www.w3.org/2002/xforms"
  xmlns:h="http://www.w3.org/1999/xhtml">

  <h:head>
    <h:title>Range controls</h:title>
    <model>
      <instance>
        <root id="range-controls">
          <range-decimal>1.0</range-decimal>
          <range-int>2</range-int>

          <vertical>
            <v-decimal>1.25</v-decimal>
            <v-int>1</v-int>
          </vertical>
        </root>
      </instance>
      <bind nodeset="/root/range-decimal" type="xsd:decimal" />
      <bind nodeset="/root/range-int" type="int" />
      <bind nodeset="/root/vertical/v-decimal" type="decimal" />
      <bind nodeset="/root/vertical/v-int" type="xsd:int" />
    </model>
  </h:head>

  <h:body>
    <range ref="/root/range-decimal" start="-2.0" end="2.0" step="-0.5">
      <label>Range control (decimal)</label>
    </range>
    <range ref="/root/range-int" start="-6" end="4" step="-2">
      <label>Range control (int)</label>
    </range>

    <group ref="/root/vertical">
      <range ref="/root/vertical/v-decimal" start="-5.0" end="7.0" step="-0.5"
        appearance="vertical">
        <label>Range control (decimal)</label>
      </range>
      <range ref="/root/vertical/v-int" start="-3" end="4" step="-3" appearance="vertical no-ticks">
        <label>Range control (int)</label>
      </range>
    </group>
  </h:body>

</h:html>

```

### DIFF
--- a/.changeset/proud-years-chew.md
+++ b/.changeset/proud-years-chew.md
@@ -1,0 +1,6 @@
+---
+'@getodk/xforms-engine': patch
+'@getodk/common': patch
+---
+
+Fixed handling of ranges with negative step values
--- a/packages/common/test-utils/xform-dsl/index.ts
+++ b/packages/common/test-utils/xform-dsl/index.ts
@@ -127,9 +127,9 @@ export const select1 = (ref: string, ...children: XFormsElement[]): XFormsElemen
 };
 
 interface RangeAttributes {
-	start: string;
-	end: string;
-	step: string;
+	start: number;
+	end: number;
+	step: number;
 }
 export const range = (
 	ref: string,
--- a/packages/common/test-utils/xform-dsl/index.ts
+++ b/packages/common/test-utils/xform-dsl/index.ts
@@ -126,6 +126,22 @@ export const select1 = (ref: string, ...children: XFormsElement[]): XFormsElemen
 	return t(`select1 ref="${ref}"`, ...children);
 };
 
+interface RangeAttributes {
+	start: string;
+	end: string;
+	step: string;
+}
+export const range = (
+	ref: string,
+	attributes: RangeAttributes,
+	...children: XFormsElement[]
+): XFormsElement => {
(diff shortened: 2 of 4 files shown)


latin-panda (approved):
One small comment below, but approving now to unblock

latin-panda:
_I couldn't comment inline 🤷🏽‍♀️ pasting it here instead:_

[Line 87:](https://github.com/getodk/web-forms/blob/a73b84aa5bec5898ef30b67351c438a61aea584a/packages/xforms-engine/src/parse/body/control/RangeControlDefinition.ts#L87) The flag looks odd to me. I'd prefer parsing here. 
```js
		const step = parseNumericStringAttribute(element, 'step');
		const absStep = (step.startsWith('-') ? step.substring(1) : step) as NumericString; 

return new this(start, end, absStep);
```

garethbowen:
> The flag looks odd to me. I'd prefer parsing here.

Yeah me too... but I also didn't like the cast. I've created a second function, which calls the assertion after the `-` has been split off which feels safer to me.

PR #804, opened 2026-05-06
fix(#799): flips min and max labels to match the selected value

### LINKED ISSUE #799: Range widget in vertical aligment sets the wrong value
### Expected Behaviour

When you pick a value from a vertical range widget you expect that value to be set and submitted.

### Observed Behaviour

You actually get a different value. The actual set value is the inverse of what you wanted, for example, if you select the smallest possible value the widget will set the largest possible value.

### Reproduction Steps

Form (XML or XLSX file): [range-controls.xml](https://github.com/getodk/web-forms/blob/c5764e8ed7780a4959f6f70b885eca4946b2834f/packages/common/src/fixtures/controls/range-controls.xml)

Steps:
1. Set the slider for the "Range control (decimal)" to the top, which is labelled -5
2. Submit the form
3. The submitted value is 7.0

### Environment

- Browser version:
- OS:
- Web Forms version:

garethbowen:
Closes #799

- [x] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

The other option would be to keep the labels where they were and invert the value out of the primevue widget. However I think the min should be at the bottom, and this is consistent with other implementations, so should be more intuitive.

### What's changed

I put the min label at the bottom of the slider, and the max label at the top.

### DIFF
--- a/packages/web-forms/src/components/form-elements/range/RangeControl.vue
+++ b/packages/web-forms/src/components/form-elements/range/RangeControl.vue
@@ -59,9 +59,6 @@ const orientation = props.node.appearances.vertical ? 'vertical' : 'horizontal';
 			<div class="range-value">
 				<span>{{ numberValue }}</span>
 			</div>
-			<div class="range-bound range-min">
-				{{ start }}
-			</div>
 			<RangeSlider
 				:id="node.nodeId"
 				:disabled="node.currentState.readonly"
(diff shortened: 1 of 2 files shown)


packages/web-forms/src/components/form-elements/range/RangeControl.vue
```diff
@@ -72,6 +69,9 @@ const orientation = props.node.appearances.vertical ? 'vertical' : 'horizontal';
 				:model-value="numberValue"
 				@update:model-value="setValue"
 			/>
+			<div class="range-bound range-min">
+				{{ start }}
+			</div>
```
garethbowen, line 74: This move wasn't strictly necessary but it feels more logical for both vertical and horizontal to have the min and max labels beside each other.

latin-panda (approved):
You've killed so many fleas in this feature 😅

garethbowen:
> You've killed so many fleas in this feature

I got sick of being bitten! :face_with_head_bandage:

PR #806, opened 2026-05-07
fix(#805): allow labels with computed media urls

### LINKED ISSUE #805: Outputs in text range media chunks do not work
### Expected Behaviour

It should be possible to set the value of a media chunk based on a calculation. For example in the xlsx setting the `image` column to `${myimg}`.

### Observed Behaviour

Instead the form fails to load with an error, similar to "URL constructor:   is not a valid URL."

### Reproduction Steps

Form (XML or XLSX file):

```xml
<?xml version="1.0"?>
<h:html xmlns="http://www.w3.org/2002/xforms"
  xmlns:h="http://www.w3.org/1999/xhtml"
  xmlns:ev="http://www.w3.org/2001/xml-events"
  xmlns:xsd="http://www.w3.org/2001/XMLSchema"
  xmlns:jr="http://openrosa.org/javarosa"
  xmlns:orx="http://openrosa.org/xforms"
  xmlns:odk="http://www.opendatakit.org/xforms">
  <h:head>
    <h:title>images-choice</h:title>
    <model odk:xforms-version="1.0.0">
      <itext>
        <translation lang="default" default="true()">
          <text id="/data/note:label">
            <value>Reference image</value>
            <value form="image"> jr://images/<output value=" /data/animal "/> </value>
          </text>
        </translation>
      </itext>
      <instance>
        <data id="images-choice" version="20240611120218">
          <animal/>
          <note/>
          <meta>
            <instanceID/>
          </meta>
        </data>
      </instance>
      <instance id="animals">
        <root>
          <item>
            <itextId>animals-0</itextId>
            <name>tiger</name>
            <img>tiger.jpg</img>
          </item>
          <item>
            <itextId>animals-1</itextId>
            <name>camel</name>
            <img>camel.jpg</img>
          </item>
        </root>
      </instance>
      <bind nodeset="/data/animal" type="string"/>
      <bind nodeset="/data/meta/instanceID" type="string" readonly="true()" jr:preload="uid"/>
    </model>
  </h:head>
  <h:body>
    <select1 ref="/data/animal">
      <label>Animal</label>
      <itemset nodeset="instance('animals')/root/item">
        <value ref="img"/>
        <label ref="img"/>
      </itemset>
    </select1>
    <input ref="/data/note">
      <label ref="jr:itext('/data/note:label')"/>
    </input>
  </h:body>
</h:html>
```

Steps:
1. Load the form.
2. See error

NB: This is the XML produced from xls2xform: `<value form="image"> <output value=" /data/animal "/> </value>`. If you manually remove the whitespace around the output then you don't get the error but the image doesn't load either.

### Environment

- Browser version:
- OS:
- Web Forms version:

### Additional Context

Reported on the forum: https://forum.getodk.org/t/limiting-entity-access-for-app-users-and-data-collectors/54386/23

garethbowen:
Closes #805

- [ ] Chrome
- [x] Firefox
- [ ] Safari (macOS)
- [ ] Safari (iOS)
- [ ] Chrome for Android
- [ ] Not applicable

7-images-choice-computed.xml in this PR

### What's changed

Media chunks are combined into a single string which is treated as the URL so if there are multiple elements in the chunk it still works.

### DIFF
--- a/packages/web-forms/src/components/common/media/MediaBlockBase.vue
+++ b/packages/web-forms/src/components/common/media/MediaBlockBase.vue
@@ -63,6 +63,7 @@ const loadMedia = async (src?: JRResourceURL): Promise<void> => {
 const setMedia = (value: string) => {
 	mediaUrl.value = value;
 	loading.value = false;
+	errorMessage.value = '';
 };
 
 const handleError = (error: string) => {
--- a/packages/xforms-engine/src/lib/reactivity/text/createTextRange.ts
+++ b/packages/xforms-engine/src/lib/reactivity/text/createTextRange.ts
@@ -20,29 +20,47 @@ interface ChunksAndMedia {
 	mediaSources: MediaSources;
 }
 
+const generateResourceChunk = (context: EvaluationContext, child: Element, type: ResourceType) => {
+	const parts = [];
+	for (const grandchild of child.childNodes) {
+		if (isElementNode(grandchild)) {
+			const expression = TextChunkExpression.fromOutput(grandchild);
+			if (expression) {
+				parts.push(createComputedExpression(context, expression)());
+			}
+		} else if (isTextNode(grandchild)) {
+			parts.push(grandchild.data);
+		}
+	}
+	const url = parts.join('') as JRResourceURLString;
+	return TextChunkExpression.fromResource(url, type);
+};
+
 const generateChunk = (node: Node): TextChunkExpression<'string'> | null => {
 	if (isElementNode(node)) {
 		return TextChunkExpression.fromOutput(node);
 	}
 	if (isTextNode(node)) {
-		const formAttribute = node.parentElement!.getAttribute('form') as ResourceType;
-		if (isResourceType(formAttribute)) {
-			return TextChunkExpression.fromResource(node.data as JRResourceURLString, formAttribute);
-		}
 		return TextChunkExpression.fromLiteral(node.data);
 	}
 	return null;
 };
 
 const generateChunksForTranslation = (
+	context: EvaluationContext,
 	textElement: Element
 ): Array<TextChunkExpression<'string'>> => {
 	const chunks = [];
-	for (const child of textElement.childNodes) {
-		for (const grandchild of child.childNodes) {
-			const chunk = generateChunk(grandchild);
-			if (chunk) {
-				chunks.push(chunk);
+	for (const child of textElement.children) {
+		const formAttribute = child.getAttribute('form') as ResourceType;
+		if (isResourceType(formAttribute)) {
+			chunks.push(generateResourceChunk(context, child, formAttribute));
+		} else {
+			for (const grandchild of child.childNodes) {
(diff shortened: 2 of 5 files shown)


packages/web-forms/src/components/common/media/MediaBlockBase.vue
```diff
@@ -63,6 +63,7 @@ const loadMedia = async (src?: JRResourceURL): Promise<void> => {
 const setMedia = (value: string) => {
 	mediaUrl.value = value;
 	loading.value = false;
+	errorMessage.value = '';
```
garethbowen, line 66: Without this the first time through media was successfully loaded but because it's async it never showed because the message took precedence.

packages/xforms-engine/src/lib/reactivity/text/createTextRange.ts
```diff
@@ -20,29 +20,47 @@ interface ChunksAndMedia {
 	mediaSources: MediaSources;
 }
 
+const generateResourceChunk = (context: EvaluationContext, child: Element, type: ResourceType) => {
+	const parts = [];
+	for (const grandchild of child.childNodes) {
+		if (isElementNode(grandchild)) {
+			const expression = TextChunkExpression.fromOutput(grandchild);
+			if (expression) {
+				parts.push(createComputedExpression(context, expression)());
```
garethbowen, line 29: This is the crux of the change - by evaluating the output element the reactive code updates when the value changes.

packages/xforms-engine/src/lib/reactivity/text/createTextRange.ts
```diff
@@ -83,9 +101,10 @@ const createTextChunks = <Role extends TextRole>(
 	const chunkExpressions = getChunkExpressions(context, definition);
 	chunkExpressions.forEach((chunkExpression) => {
 		if (chunkExpression.resourceType) {
-			mediaSources[chunkExpression.resourceType] = JRResourceURL.from(
-				chunkExpression.stringValue as JRResourceURLString
-			);
+			const url = chunkExpression.stringValue?.trim();
+			if (JRResourceURL.isJRResourceReference(url)) {
```
garethbowen, line 105: This validation step is important... if you call `JRResourceURL.from` with an invalid URL it throws an Error which can be fatal. If this is based on user input, or at load time the input is blank, then we should just handle it gracefully.

garethbowen:
@latin-panda This isn't urgent IMO but was an interesting problem to solve while I was waiting for the monorepo code to land. But I'm happy enough with the solution to send it for review.

latin-panda (approved):
Nice ✨ I didn't spot any regression. Could you please add a scenario test for this case?

garethbowen:
> Could you please add a scenario test for this case?

Added testing for image, audio, and video, both hardcoded and calculated.

latin-panda:
Thank you! It's ready to ship!

garethbowen:
Ported to: https://github.com/garethbowen/central-frontend/tree/805-outputs-in-media

